PR #726, opened 2017-05-03
Defer updates based on filesystem events when Atom is in the background

### LINKED ISSUE #348: Pulling changes with the git panel open can throw a fatal index.lock error
### Steps to Reproduce

1. Have the git panel open (make sure you don't have any staged or unstaged changes though)
1. `git pull` from the remote repo associated with your open Atom project on the command line
1. Get this error on the command line:

```
fatal: Unable to create '/Users/machiste/code/my-sweet-app/.git/index.lock': File exists.

Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.
```

There might be more precise steps to repro than this – I'm not sure yet 🤔

### LINKED ISSUE #569: Don't run `status` while in the background
We want to avoid `index.lock` contention when on the command line, so we should avoid running `git status` variants when the app is in the background.

Wondering if this means we can't respond to *any* changes in the background??

kuychaco:
Fixes #569.
Fixes #348.

We were seeing issues with lock contention when interacting with git on the command line for a repo that was open in Atom. This is because when Atom detects changes to the repo, it updates all of its git data by shelling out and running `git status`. `status` locks the index while performing caching, thus preventing command line git commands from running successfully and resulting in errors such as `Another git process seems to be running in this repository`.

This PR implements DeferredCallbackQueue which tracks whether the window is focused or blurred and pauses updates accordingly. When blurred, update events are stored off in a queue and processed after a specified quiet time (currently set to 3 seconds) to reduce contention. The idea is that Atom will wait until the users activity has died down before it tries to perform its update. If the application is focused before the wait period ends, the update is triggered immediately upon focus.

### DIFF
--- a/lib/deferred-callback-queue.js
+++ b/lib/deferred-callback-queue.js
@@ -4,7 +4,6 @@ import {autobind} from 'core-decorators';
 
 export default class DeferredCallbackQueue {
   constructor(wait, callback, options = {}) {
-    console.log('create DeferredCallbackQueue');
     this.wait = wait;
     const currentWindow = remote.getCurrentWindow();
 
@@ -32,13 +31,13 @@ export default class DeferredCallbackQueue {
   @autobind
   pause() {
     if (this.paused) { return; }
-    console.log('paused');
     this.paused = true;
   }
 
   @autobind
   resume() {
     if (!this.paused) { return; }
+    console.log('resume');
     this.paused = false;
     if (this.timer) {
       clearTimeout(this.timer);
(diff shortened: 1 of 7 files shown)


smashwilson:
> And I do like your idea of moving refresh/blur logic from WorkspaceChangeObserver to DeferredCallbackQueue. Will give it a stab and see what it looks like

Cool, cool. How would you feel about merging this as-is and doing the consolidation work in another PR? We're ✅ here and I don't want to delay the fix with a bunch of refactoring suggestions :grin:

Issue #740, opened 2017-05-05
Flake: "Git commands for CompositeGitStrategy ... https authentication fails the command on dialog cancel"


smashwilson:
Test failures seen on Travis on a Mac instance.

smashwilson:
Resolved in #748. :sparkles:

PR #742, opened 2017-05-06
Allow previous commit message to be erased when amending

### LINKED ISSUE #463: Amended commit message does not disappear when using backspace
Repro steps:

1. Click the amend check box
2. Hit backspace until entire message is deleted

Expected: commit message box is empty
Actual: amended message comes back

Note: Highlighting entire amended commit message and deleting results in an empty commit message box. Highlighting message that is NOT the amended commit message and deleting results momentarily results in an empty message box, but then the amended commit message comes back.

### Versions

Atom Version (`atom --version`):

GitHub Package Version (`git --git-dir ~/.atom/packages/github/.git rev-parse head`): 65a240383f854b41fc037018c9215dca2e116824

### LINKED ISSUE #573: Amend commit message keeps getting replaced when emptied
Every time I backspace to the beginning, my message is replaced. This was proceeded by seeing a stale message, which I deleted, only to have it replaced with the correct message repeatedly as shown. Strangely, when I hit `ctrl-a` to go to the beginning of the line and then killed it with `ctrl-k` it didn't replace my message. But then I typed a few characters and backspaced and it happened again.

amend problem

### Versions

Atom Version (`atom --version`):
Atom    : 1.16.0-dev-91181a331
Electron: 1.4.15
Chrome  : 53.0.2785.143
Node    : 6.5.0

GitHub Package Version (`git --git-dir ~/.atom/packages/github/.git rev-parse head`):
9ffc68101b64712fbf441e6398750631288c09f4

BinaryMuse:
Fixes #573 
Fixes #463

Previously, when preparing a commit with amending enabled, we were trying to determine at render-time when to prefill the commit message editor with the message from the previous commit (i.e. the commit we were amending). We did so any time there was no text in the text editor and amending was enabled. This changes the behavior so that we only do so when (1) the amend checkbox is first checked and (2) there is no saved in-progress amend-mode message.

Presently, this means that the only way to *re*-prefill the editor with the previous commit's message is to completely erase any in-progress message and then uncheck and re-check the amending checkbox.

github-package-amended-amending

/cc @kuychaco @nathansobo


smashwilson:
The test failure here is the flake #740. Hopefully #748 will take care of that one, once Travis is back on its feet.

smashwilson:
✅ :sparkles:

PR #748, opened 2017-05-07
Make failing test fail

BinaryMuse:
@smashwilson This test was throwing asynchronously and causing the test to time out instead of failing. Not 100% sure what the proper solution to the actual failure is (seems to be a messaging thing?)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -28,6 +28,7 @@
     "xdescribe": true,
     "it": true,
     "xit": true,
+    "stress": true,
     "waitsForPromise": true,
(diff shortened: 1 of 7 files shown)


smashwilson:
Let's see if this helps 🤞

smashwilson:
🙌  This should be more reliable. I copied that pattern to the rest of the test cases, and added stronger sequencing around when the git process and the helper process are killed. It takes another few IPC messages but it'll prevent cancelling a dialog at the wrong time from falling back to a different dialog by mistake.

PR #758, opened 2017-05-08
Create absentWorkdirContext lazily

BinaryMuse:
Previously, the `absentWorkdirContext` singleton was being created eagerly during require time. This breaks bundling because that code path creates a `CompositeGitStrategy` which needs to require `os` (but `require` isn't available during the snapshotting phase).

@smashwilson I'm going to merge this so I can bump and continue the snapshotting work but feel free to suggest a different mechanism for handling this.


smashwilson:
> [..] feel free to suggest a different mechanism for handling this.

Nope, this is :100: :+1:

I need to internalize the things that will cause issues during bundling 🤔

Issue #534, opened 2017-02-17
Commit panel should be automatically focused when toggled via a key binding


nathansobo:
I frequently show the commit panel with `ctrl-9`. In a situation where I am showing the panel for the first time, I think it makes sense to transfer focus to it so it can be used, even though there's a separate binding for toggling focus independently of visibility.

PR #777, opened 2017-05-10
Revisit keybindings and focus management

smashwilson:
Our keybindings and focus restoration logic has been inconsistent and buggy for a while, so I'm doing a pass to correct this. While I'm at it, I'll implement @nathansobo's suggested changes to the keybinding strategy:

> Can we subtly massage the tree view bindings around ctrl-0 to make them blend in with the bindings for the commit panel. I'd like a pattern to emerge around using ctrl-# and ctrl-shift-# to toggle various panels in an intuitive way.
>
> Maybe ctrl-# toggles focus, showing the panel if necessary (because you can't focus it if it isn't visible) and ctrl-shift-# toggles visibility (including for the tree view).

I'll open a separate PR for discussion about adding keymap entries for commands that currently have none, like the workspace-global `github:commit`, `github:clone`, and others.

#### Remaining

- [x] Update keymap
- [x] Use Symbols instead of DOM references to remember the last focus within the git panel, because DOM nodes may be created and destroyed
- [x] Fix focus changes on navigation between the StagingView and CommitView
- [x] Fix focus changes on navigation between the Git tab and the FilePatchView
- [x] Stub out focus handling for the GitHub tab
- [x] Eliminate the extra click necessary to bring focus to the Git and FilePatchView tabs after activation
- [x] Actually pay attention to the tests
- [x] Make ctrl-shift-# actually do the thing I said I was going to, unless it turns out not to be possible at all.

/cc @jasonrudolph because he expressed interest :grin:
/cc #534


nathansobo:
Very glad you're doing a pass on this!

lib/controllers/git-tab-controller.js
```diff
@@ -327,16 +333,25 @@ export default class GitTabController {
...
   }
 
-  focusAndSelectStagingItem(filePath, stagingStatus) {
-    return this.refs.gitTab.focusAndSelectStagingItem(filePath, stagingStatus);
+  restoreFocus() {
+    this.refs.gitTab.setFocus(this.lastFocus);
+  }
+
+  hasFocus() {
+    return this.element.contains(document.activeElement);
```
BinaryMuse: Will `this.element === document.activeElement` ever be true?

lib/controllers/root-controller.js
```diff
@@ -81,7 +81,7 @@ export default class RootController extends React.Component {
       ...nullFilePatchState,
       amending: false,
       gitTabActive: props.firstRun || props.savedState.gitTabActive,
-      githubPanelActive: props.firstRun || props.savedState.githubPanelActive,
+      githubTabActive: props.firstRun || props.savedState.githubTabActive || props.savedState.githubPanelActive,
```
BinaryMuse: 👌

lib/controllers/root-controller.js
```diff
@@ -800,3 +742,104 @@ export default class RootController extends React.Component {
...
+    this.controllerKey = `${name}TabController`;
+  }
+
+  getState() {
+    return this.component.state[this.stateKey];
+  }
+
+  setStateKey(value) {
+    return new Promise(resolve => {
+      this.component.setState({[this.stateKey]: value}, resolve);
```
BinaryMuse: How do you feel about injecting a `setState` function rather than reaching into the component?

BinaryMuse: On second hand, we're reaching in a few other places too. As a general pattern that makes me nervous but here I think it'll do fine. 👍

lib/controllers/git-tab-controller.js
smashwilson: It turns out that `.contains()` is [inclusive](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains):

```javascript
> document.body.contains(document.body)
true
```

So, we can save a little typing. Although maybe it's worth it to save an unnecessary DOM operation... ?

lib/controllers/root-controller.js
smashwilson: I like it :+1: Will refactor when ✅

lib/controllers/git-tab-controller.js
BinaryMuse: Ah, no I was thinking of a method that bit me once because it *wasn't* inclusive... wonder what I'm thinking of?

lib/controllers/root-controller.js
smashwilson: Oh, I see...

... So, it turns out that JavaScript _does_ let you nest classes:

screen shot 2017-05-10 at 10 38 31 pm

This might be a good candidate for that?

smashwilson: Or would that just muddle things a lot more 🤔

smashwilson:
Oh, now the `ctrl-shift-#` bindings are both showing and focusing the Git and GitHub panels; not quite what we want. I'll have to see if it's possible to make them toggle the containing dock, but not trigger focus.

keymaps/git.cson
```diff
@@ -1,7 +1,8 @@
 '.workspace':
-  'ctrl-9': 'github:toggle-git-tab'
-  'ctrl-8': 'github:toggle-github-tab'
-  'ctrl-(': 'github:toggle-git-tab-focus' # ctrl-shift-9
+  'ctrl-8': 'github:toggle-github-tab-focus'
+  'ctrl-*': 'github:toggle-github-tab' # ctrl-shift-8
```
50Wliu: I feel like just making the keybinding `'ctrl-shift-8'` would make it immediately clear what the intention was, and allow the comment to be removed?

smashwilson: Huh, for some reason I thought that wouldn't work, but it absolutely does. Thanks :+1:

lib/controllers/root-controller.js
smashwilson: It turns out that I could refactor TabTracker to have injected functions instead of reaching into Component directly without too much trouble. Thanks 🙇

maxbrunsfeld:
👌  Thanks @smashwilson!

PR #779, opened 2017-05-10
Ensure that activating a DockItem shows the corresponding Dock

BinaryMuse:
#673 made it show the Dock items are shown on first launch; however, the right Dock defaults to closed. This changes it so when we `activate` the item, it also shows the Dock.

### DIFF
--- a/lib/views/dock-item.js
+++ b/lib/views/dock-item.js
@@ -109,6 +109,11 @@ export default class DockItem extends React.Component {
       const pane = this.props.workspace.paneForItem(this.dockItem);
       if (pane) {
         pane.activateItem(this.dockItem);
+        const dock = this.props.workspace.getPaneContainers()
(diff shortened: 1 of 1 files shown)


BinaryMuse:
I feel like there must be a better way to do this. :expressionless:

lib/views/dock-item.js
```diff
@@ -109,6 +109,11 @@ export default class DockItem extends React.Component {
       const pane = this.props.workspace.paneForItem(this.dockItem);
       if (pane) {
         pane.activateItem(this.dockItem);
+        const dock = this.props.workspace.getPaneContainers()
+          .find(container => container.getPanes().find(p => p.getItems().includes(this.dockItem)));
+        if (dock && dock.constructor.name === 'Dock') {
```
smashwilson: It might be better to probe for the method directly rather than using the constructor name:

```javascript
dock && dock.show && dock.show();
```

PR #788, opened 2017-05-12
Don't attempt to show diff for added binary files

BinaryMuse:
Our `diff` command returns an empty diff for binary files that exist in the repo; for files that don't exist in the repo, we read the contents of the file ourselves and manually construct a diff. This means that if a binary file was *added* in repository, clicking on it would show binary garbage in a diff view (and likely hang the editor attempting to render it).

This reads the first 50 bytes of the file and attempts to guess if it's binary or not, and if it is, treats the content as being empty.

In a future PR, I'd like to have more intelligent UI if a file is binary.


smashwilson:
Is there any way we could take advantage of `.gitattributes` to do this? That might help us stay consistent with git in weird cases.

I'm assuming there's some kind of plumbing commands for querying attribute data?

BinaryMuse:
Hm, `.gitattributes` would only affect files that were specifically listed as binary in there, right? And not regular binary files (that would normally be guessed properly)?

smashwilson:
Good point. I thought there were some kind of default behavior for attributes, but [looking around on StackOverflow](http://stackoverflow.com/questions/6119956/how-to-determine-if-git-handles-a-file-as-binary-or-as-text), git itself does basically this.

the-j0k3r:
@BinaryMuse this should account for images too, why doesnt it?

PR #789, opened 2017-05-12
Don't update Push button to say Force Push when cmd/ctrl held

### LINKED ISSUE #766: Force Push state changes rapidly when holding down ctrl
The GIF does a horrible job of displaying the issue due to the 8fps default that I forgot to change.  But basically, to reproduce, just open the push/pull panel and hold down cmd/ctrl.
force-push-flicker

Fairly certain I'm on 4f41080a3ee42ce523402231ca67fc1923b39112 (I've pulled since then, but haven't restarted Atom).

BinaryMuse:
It seems that, on certain systems, this doesn't work as expected. Sometimes it flashes very quickly between the two states, and using `ctrl` as a modifier for mouse clicks can also put it into a weird state. I've seen both of these states on my own computer, but have been unable to reproduce consistently.

The last thing we want to do is accidentally cause a user to force push. So, instead, we use the `metaKey` and `ctrlKey` properties of the click event to determine if we should force push. Since the feature is, unfortunately, less discoverable, I added a confirmation that makes sure the user meant to force push.

Fixes #766

### DIFF
--- a/lib/controllers/status-bar-tile-controller.js
+++ b/lib/controllers/status-bar-tile-controller.js
@@ -234,8 +234,8 @@ export default class StatusBarTileController extends React.Component {
   async doPush(options) {
     if (options.force) {
       const choice = this.props.confirm({
-        message: 'Are you sure you wish to force push?',
-        detailedMessage: 'This operation could result it losing data on the remote.',
+        message: 'Are you sure you want to force push?',
+        detailedMessage: 'This operation could result in losing data on the remote.',
         buttons: ['Force Push', 'Cancel Push'],
       });
       if (choice !== 0) { return; }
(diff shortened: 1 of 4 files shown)


lib/controllers/status-bar-tile-controller.js
```diff
@@ -231,6 +232,15 @@ export default class StatusBarTileController extends React.Component {
   }
 
   async doPush(options) {
+    if (options.force) {
+      const choice = this.props.confirm({
+        message: 'Are you sure you wish to force push?',
+        detailedMessage: 'This operation could result it losing data on the remote.',
```
smashwilson: > [...] could result *it* losing [...]

Should be:

> [...] could result *in* losing [...]

PR #790, opened 2017-05-12
Hey look some keybindings

smashwilson:
These are some proposed default keybindings for various `github:` actions for the more keyboard-focused among us ⌨️ . The logic here is that `alt-g` is already used as a chord prefix for [the open-on-github package actions](https://github.com/atom/open-on-github/blob/master/keymaps/to-the-hubs.cson), so we already have "githubby" associations there to take advantage of.

* `alt-g enter` is "commit" for symmetry with the `cmd-enter` binding within the commit message editor (and you should be able to use it within there too).
* `alt-g p` is push and `alt-g P` is force-push (for "extra emphasis push").
* `alt-g f` is fetch and `alt-g F` is pull (like a "fetch and then some").
* `alt-g =` is clone because hitting the `+` key feels like "add a repository".

The conflict resolution bindings are lifted [right from merge-conflicts](https://github.com/smashwilson/merge-conflicts/blob/master/keymaps/merge-conflicts.cson).

I'm not totally married to any of these, I just kind of picked things that weren't taken. If any of you have other bindings that you'd prefer, speak now or hold your peace until, uh, next month I suppose.

bikeshed

/cc @kuychaco @BinaryMuse @nathansobo @as-cii @simurai for thoughts or preferences.


nathansobo:
I like it, except for one concern: @Ben3eeE Could the `alt-g` problematic for international keyboard users?

Ben3eeE:
`alt-[alphanumeric]` should always be safe. AltGr (Or `ctrl-alt`) is the problematic one.

I also had worry about the other keybindings in https://github.com/atom/github/pull/777 because of the menu duplication bug... But I believe @50Wliu addressed that with his review comment.

My worry in #777 was:

Binding to `ctrl-*` would (possibly) get the [duplication bug](https://github.com/electron/electron/issues/8741) where one command is from the menu and one command is from the keybindings system on international layouts. Also `*` is nowhere near `8` so it would be weird. This is something that always felt strange about the way Atom handles keybindings and it is amplified by the menu bug.

@nathansobo How do you feel about not resolving symbols by default? What I mean is pressing `ctrl-shift-8` would always resolve to `ctrl-shift-8` and not resolve to `ctrl-*` on US layout and `ctrl-(` on Swedish layout? Maybe this would break existing keybindings and packages (Like toggle line comments because that uses `ctrl-/`)?.... What is the sane what to handle this without breaking everything?
Having `ctrl-/` for toggle line comments is very common and for me it is very familiar that I access it using `ctrl-shift-7` (In several programs) but if we don't resolve symbols that wouldn't work any more.. Just 💭 out loud about something that has been on my mind for a while.

smashwilson:
> `alt-[alphanumeric]` should always be safe. AltGr (Or `ctrl-alt`) is the problematic one.

Good to know 🌈 

> Binding to ctrl-* would (possibly) get the duplication bug where one command is from the menu and one command is from the keybindings system on international layouts. Also * is nowhere near 8 (Depending on layout) so it would be weird for non US layouts.

As it happens I'd not been aware that you could specify bindings as an unexpanded `ctrl-shift` at all. `ctrl-shift-8` and `ctrl-shift-9` should be okay and portable, then?

Ben3eeE:
Yeah unexpanded should be fine :shipit:

BinaryMuse:
I was initially nervous about a keybinding for force push, but now that we confirm I'd say this is okay. 👍

Ben3eeE:
Now that I think about this again I realize that `alt-` is used to access the menu on Windows I was just considering AltGraph variants when replying...

@damieng Should review this. We already have things on `alt-g` from `open-on-github` and `git-diff` if we can't bind to `alt-` because of the menu maybe those should be changed?

We should really make a list of what is safe or not there are too many things to consider for this to track of 😀

smashwilson:
> We should really make a list of what is safe or not there are too many things to consider for this to track of 😀

:100:

Or, better yet: a linter?

damieng:
alt-g is fine on Windows - none of the top-level menu items map to it (although testing this did make me realize we seem to have lost the menu accelerator underlines when tapping alt - possibly a regression on Electron I should check out)

nathansobo:
> How do you feel about not resolving symbols by default? What I mean is pressing ctrl-shift-8 would always resolve to ctrl-shift-8 and not resolve to ctrl-* on US layout and ctrl-( on Swedish layout?

The way the keybindings system works right now is when the user presses a key, the keyboard event is translated to a string representing the event. The goal of normalization in the keymap definitions is to make the string in the keymap map this string that we map the event to. We compare them with basic string comparison.

To do what you're talking about, I think we'd need a more sophisticated approach to matching keystrokes than simple strings, right? We'd need to look at the event and see if the string from the map matches the event for the current keyboard layout.

keymaps/git.cson
```diff
@@ -3,6 +3,19 @@
   'ctrl-shift-8': 'github:toggle-github-tab'
   'ctrl-9': 'github:toggle-git-tab-focus'
   'ctrl-shift-9': 'github:toggle-git-tab'
+  'alt-g enter': 'github:commit'
+  'alt-g f': 'github:fetch'
+  'alt-g F': 'github:pull'
```
lee-dohm: We've moved away from using the capital letter form because it was confusing for some people that keymaps are case-sensitive. This (and the others like it) should be `alt-g shift-f`.

damieng: I'd love it if we could do alt-g up-arrow and alt-g down-arrow but I'm pretty sure we don't support that kind of keybinding today.

smashwilson: @damieng Sadly those ones are taken by [git-diff](https://github.com/atom/git-diff/blob/master/keymaps/git-diff.cson). I kind of wanted those too, to be honest :smile:

nathansobo: @lee-dohm Have we changed the normalization scheme in the `atom-keymap` source code? If not, there's still places where we'll display upcased shift-modified keycaps in the editor, as in `alt-shift-S` instead of `alt-shift-s`.

lee-dohm: @nathansobo It looks like it was updated so that we should always see the `alt-shift-S` form post-normalization:

https://github.com/atom/atom-keymap/blob/master/spec/helpers-spec.coffee#L8

From the specs it appears that any sub-sequence with a `shift` and an alphabetic character will capitalize the alphabetic character in the normalized form. Personally, I don't think that's a big deal ... but it would probably be less confusing if we normalized to all lower case.

smashwilson: @lee-dohm Done. Thanks for the catch 🙇

smashwilson:
Okay, the build failures here are related to the latest Atom beta, and this diff is pretty small, so...

:sparkles: _YOLO MERGE TIME_ :sparkles:

PR #796, opened 2017-05-13
Include a payload in WorkspaceChangeObserver.didUpdate() from focus events

### LINKED ISSUE #794: Uncaught TypeError: Cannot read property 'map' of undefined
[Enter steps to reproduce:]

1. Open Atom
2. Open some project or folder
3. Make a change to code (JS or HTML)

**Atom**: 1.18.0-dev-1992c17 x64
**Electron**: 1.3.15
**OS**: Ubuntu 16.04.2
**Thrown From**: [github](https://github.com/atom/github) package 0.0.6

### Stack Trace

Uncaught TypeError: Cannot read property 'map' of undefined

```
At /usr/local/share/atom/resources/app/node_modules/github/lib/models/workdir-context.js:148

TypeError: Cannot read property 'map' of undefined
    at WorkdirContext.subs.add.observer.onDidChange.events (/usr/local/share/atom/resources/app/node_modules/github/lib/models/workdir-context.js:148:1)
    at Function.module.exports.Emitter.simpleDispatch (/usr/local/share/atom/resources/app/node_modules/event-kit/lib/emitter.js:25:1)
    at Emitter.module.exports.Emitter.emit (/usr/local/share/atom/resources/app/node_modules/event-kit/lib/emitter.js:141:1)
    at WorkspaceChangeObserver.didChange (/usr/local/share/atom/resources/app/node_modules/github/lib/models/workspace-change-observer.js:138:1)
    at focusHandler (/usr/local/share/atom/resources/app/node_modules/github/lib/models/workspace-change-observer.js:108:1)
```

### Commands

```
     -0:49.3.0 application:open-folder (input.hidden-input)
  2x -0:33.5.0 deprecation-cop:view (atom-workspace.workspace.scrollbars-visible-always.theme-one-dark-syntax.theme-one-dark-ui)
     -0:06.8.0 tree-view:remove (span.name.icon.icon-file-text)
```

### Non-Core Packages

```
atom-clock 0.1.7 
```

smashwilson:
Introduce a `{special: FOCUS}` event payload when a workspace change observer triggers a focus event (which doesn't have a path). The `special` key is propagated through the WorkdirContext event handler and handled by the `Present` state by invalidating status and all file patches.

I've also added an `ATOM_GITHUB_WORKSPACE_OBSERVER` env var to let us override the FilesystemChangeObserver on Mac or Windows to let us test the Linux code paths a little more easily.

Fixes #794.

### DIFF
--- a/lib/models/repository-states/present.js
+++ b/lib/models/repository-states/present.js
@@ -3,6 +3,7 @@ import path from 'path';
 import State from './state';
 
 import {deleteFileOrFolder} from '../../helpers';
+import {FOCUS} from '../workspace-change-observer';
 import FilePatch from '../file-patch';
 import Hunk from '../hunk';
 import HunkLine from '../hunk-line';
@@ -73,6 +74,14 @@ export default class Present extends State {
     const keys = new Set();
     for (let i = 0; i < paths.length; i++) {
       const fullPath = paths[i];
+
+      if (fullPath === FOCUS) {
+        keys.add(Keys.changedFiles);
+        for (const k of Keys.filePatch.eachWithOpts({staged: false})) {
+          keys.add(k);
+        }
+        continue;
+      }
 
       const endsWith = (...segments) => fullPath.endsWith(path.join(...segments));
       const includes = (...segments) => fullPath.includes(path.join(...segments));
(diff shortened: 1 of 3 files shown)


smashwilson:
@BinaryMuse How's this for a stopgap?

I think the Real Solution :tm: is to improve our filesystem watching situation on Linux so we can scale a real watcher to cover the working directory like we do on other platforms, but this'll get us running and behaving reasonably on Linux systems.

BinaryMuse:
Thanks for tackling this.

Issue #786, opened 2017-05-12
Defer displaying large diffs


BinaryMuse:
Clicking on a large diff can lock up the editor while all the lines are displayed in the `FilePatchView`. A quick solution would be to display a notice that the diff is large and a button to load it, similar to GitHub.com.

PR #799, opened 2017-05-14
Don't display large diffs by default

BinaryMuse:
Large diffs can slow down rendering and make Atom feel slow. This is a stopgap solution that presents a GitHub.com-style warning about large diffs that the user can confirm to see the diff; I've set the default threshold to 1000 lines.

Once a user confirms a large diff, its path gets added to a Set that ensures any diff with the same path will always be shown. I debated on this functionality but in the end decided it would make the most sense. 💭 s?

Fixes https://github.com/atom/github/issues/786

### DIFF
--- a/lib/controllers/file-patch-controller.js
+++ b/lib/controllers/file-patch-controller.js
@@ -1,3 +1,5 @@
+import path from 'path';
+
 import React from 'react';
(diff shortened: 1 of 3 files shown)


BinaryMuse:
github-package-large-diff

PR #820, opened 2017-05-16
Fix not matching branches with '/' in name

colepatrickturner:
### Requirements

- [X] Template filled out
- [X] I don't believe this change requires a test

### Description of the Change

This commit fixes issues where branches will not display if they contain a '/' character.

For example: 'hello/world'

### Alternate Designs

This came about because I am required to name certain branches with my initials such as:
ct/hello-world
ct/fix-bug-x

### Benefits

Branches that otherwise will not display in the dropdown menu will be included.

### Possible Drawbacks

Without this pull request, I cannot change to these kinds of branches from the dropdown menu.

### Applicable Issues

N/A

Thanks for your time!

### DIFF
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -591,7 +591,7 @@ export default class GitShellOutStrategy {
   }
 
   async getBranches() {
-    const output = await this.exec(['for-each-ref', '--format=%(refname:short)', 'refs/heads/*']);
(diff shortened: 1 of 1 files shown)


BinaryMuse:
Thank you so much for the fix! It's been merged into master.

We actually do prefer tests for any regressions or uncaught errors; we added one in https://github.com/atom/github/commit/71bd716ab5403f36c0cb800effb0dfb54c4585d8.

PR #831, opened 2017-05-17
wrap file names in list

will-stone:
### Description of the Change

Currently long filenames (and / or when the Git dock is narrow) the file names fall off the side. This will break the word onto multiple lines.

### Alternate Designs

Ideally, I believe the folder name should be shown above the filename for greater customisation but this is a quick fix.

### Benefits

Prevents filenames being hidden in the list, particularly useful for when there are multiple similar filenames: `/blah/blah/blah/blahblah.js` and `/blah/blah/blah/blahblah.scss`

### Possible Drawbacks

The list will be taller.

### DIFF
--- a/styles/file-patch-list-view.less
+++ b/styles/file-patch-list-view.less
@@ -30,7 +30,7 @@
 
   &-path {
     flex: 1;
-    word-break: break-all;
+    word-break: break-word;
   }
 
   &-resolutionItem {
--- a/styles/file-patch-list-view.less
+++ b/styles/file-patch-list-view.less
@@ -30,6 +30,7 @@
 
   &-path {
     flex: 1;
+    word-break: break-all;
   }
 
   &-resolutionItem {


50Wliu:
Could you please include a before/after screenshot?

will-stone:
Before:

image


After:

image

BinaryMuse:
@simurai Would you like to take an 👀 at this?

styles/file-patch-list-view.less
```diff
@@ -30,6 +30,7 @@
 
   &-path {
     flex: 1;
+    word-break: break-all;
```
simurai: Great suggestion. 👍  Although I wonder if we should use `word-break: break-word;` instead?

It might be easier to read whole words? And a file extension like `.scss` wouldn't break into `.s` and `css`. If a word is too long, it will still start breaking every character like for folder names.

`break-all` | `break-word`
--- | ---
break-all | break-word

BinaryMuse: Agreed, I think that looks better. @will-stone if you'd like to make that change, I'll get this merge — or I can make it, either way. Thanks! ✨

will-stone:
Changed to break-word as suggested; much nicer to read.

BinaryMuse:
✨ ✨ ✨  Thanks for your contribution!

Issue #638, opened 2017-04-03
Handle per-package transpilation dependencies better


BinaryMuse:
As we move to bundle this package in Atom, we're running into the situation where we need to pre-transpile the package during the build step. This requires that all the dependencies related to transpilation be present during the installation, which requires they be `dependencies`. However, this adds ~18 megabytes to the bundle size. We should consider how we might fix this given the current constraints that apm imposes (e.g. that we can't pre-transpile the files before publishing).

/cc @nathansobo as we've discussed this in the past.

PR #840, opened 2017-05-17
Make babel dependencies dev dependencies

kuychaco:
This uses the changes in https://github.com/atom/atom/pull/14445 to reduce the dependencies that we bundle in our build artifacts to reduce build packaging time (which is painfully slow on Windows due to nuget and is causing timeouts). Removing babel dependencies should drop the build size by 17MB. 

Ref https://github.com/atom/atom/pull/14445
Ref atom/github#638


Arcanemagus:
@BinaryMuse could you explain why this was reverted? The commit doing has no reasoning behind it.

BinaryMuse:
@Arcanemagus I don't remember the full reasoning, but all this is only relevant for packages bundled with Atom (https://github.com/atom/atom/pull/14445 affects the Atom build script, not installation of third-party packages).

Arcanemagus:
Ah ha! Thanks for the clarification.

PR #851, opened 2017-05-19
Implement status bundle to reduce shell outs

kuychaco:
Use `git status --porcelain=v2 --branch` to fetch changed files, current branch, and ahead/behind count


lib/git-shell-out-strategy.js
```diff
@@ -5,6 +5,7 @@ import {remote} from 'electron';
 import {CompositeDisposable} from 'event-kit';
 import {GitProcess} from 'dugite';
 import {parse as parseDiff} from 'what-the-diff';
+import {parse as parseStatus} from 'what-the-status';
```
smashwilson: This package name made me laugh a bunch 😆

lib/git-shell-out-strategy.js
```diff
@@ -316,61 +317,9 @@ export default class GitShellOutStrategy {
...
-          // skip the next line, which is the copied file
-          i++;
-        }
-      }
-    }
-
-    return {stagedFiles, unstagedFiles, mergeConflictFiles};
+  async getStatusBundle() {
+    const output = await this.exec(['status', '--porcelain=v2', '--branch', '-z']);
+    return parseStatus(output);
```
smashwilson: Something to be careful about here: for this to work on Windows, you'll need to translate any paths returned as part of this bundle through the `toNativePathSep()` function. I think this is the source of at least one of the AppVeyor failures.

I'd been preferring to do the transformation as close to git as possible, here in the GitShellOutStrategy layer, so that as much of the package as possible can work entirely with native path separators while git sees only `/` separators.

lib/git-shell-out-strategy.js
```diff
@@ -316,61 +317,9 @@ export default class GitShellOutStrategy {
...
-    const mergeConflictFiles = {};
-    let statusToHead;
-
-    if (output) {
-      const lines = output.split('\0');
-      for (let i = 0; i < lines.length; i++) {
-        const line = lines[i];
-        if (line === '') { continue; }
-        const [_, x, y, rawFilePath] = line.match(/(.)(.)\s(.*)/); // eslint-disable-line no-unused-vars
-        const filePath = toNativePathSep(rawFilePath);
```
smashwilson: :point_up: Note the translating call here.

lib/git-shell-out-strategy.js
```diff
@@ -595,6 +505,10 @@ export default class GitShellOutStrategy {
     return output.trim().split(LINE_ENDING_REGEX);
   }
 
+  async describeHead() {
+    return (await this.exec(['describe', '--contains', '--all', '--always', 'HEAD'])).trim();
+  }
```
smashwilson: :+1: Cool, I was wondering if you'd preserved the `describe`-for-detached-HEADs bit.

lib/models/repository-states/present.js
```diff
@@ -440,11 +504,14 @@ export default class Present extends State {
...
   }
 
-  getCurrentBranch() {
-    return this.cache.getOrSet(Keys.currentBranch, async () => {
-      const {name, isDetached} = await this.git().getCurrentBranch();
-      return isDetached ? Branch.createDetached(name) : new Branch(name);
-    });
+  async getCurrentBranch() {
+    const {branch} = await this.getStatusBundle();
+    if (branch.head === '(detached)') {
```
smashwilson: Fun question: what happens if someone has a branch named `(detached)`? :grin:

If it's possible to distinguish those cases in the actual status output, maybe we could use a `DETACHED` Symbol exported from `what-the-status`?

lib/models/repository-states/present.js
```diff
@@ -466,16 +533,14 @@ export default class Present extends State {
     });
   }
 
-  getAheadCount(branchName) {
-    return this.cache.getOrSet(Keys.aheadCount.oneWith(branchName), () => {
-      return this.git().getAheadCount(branchName);
-    });
+  async getAheadCount(branchName) {
+    const bundle = await this.getStatusBundle();
+    return bundle.branch.aheadBehind ? bundle.branch.aheadBehind.ahead : null;
```
smashwilson: This might be cleaner if `bundle.branch.aheadBehind` was always present and just defaulted to `{ahead: null, behind: null}`. Unless there's a case where `assert.isNull(aheadBehind)` and `assert.deepEqual(aheadBehind, {ahead: null, behind: null}` are both valid and mean different things 🤔

lib/models/repository-states/present.js
```diff
@@ -681,7 +746,7 @@ class GroupKey {
 }
 
 const Keys = {
-  changedFiles: new CacheKey('changed-files'),
+  statusBundle: new CacheKey('status-bundle'),
```
smashwilson: :+1: on the cache key structure btw.

test/git-strategies.test.js
```diff
@@ -39,130 +39,6 @@ import {fsStat, normalizeGitHelperPath, writeFile} from '../lib/helpers';
...
-          'removed-on-master.txt': {
-            ours: 'deleted',
-            theirs: 'modified',
-            file: 'added',
-          },
-        });
-      });
-
-      if (process.platform === 'win32') {
-        it('normalizes the path separator on Windows', async function() {
```
smashwilson: :point_up: This one is probably worth translating to test `getStatusBundle()`.

test/git-strategies.test.js
```diff
@@ -499,84 +375,20 @@ import {fsStat, normalizeGitHelperPath, writeFile} from '../lib/helpers';
...
-      });
-    });
-
-    describe('getCurrentBranch() and checkout(branchName, {createNew})', function() {
+    describe('checkout(branchName, {createNew})', function() {
       it('returns the current branch name', async function() {
         const workingDirPath = await cloneRepository('merge-conflict');
         const git = createTestStrategy(workingDirPath);
-        assert.deepEqual(await git.getCurrentBranch(), {name: 'master', isDetached: false});
+        assert.deepEqual((await git.exec(['symbolic-ref', '--short', 'HEAD'])).trim(), 'master');
```
smashwilson: A bit unfortunate that we lost this, but eh.

lib/models/repository-states/present.js
BinaryMuse: Oh that's unfortunate.

```
$ git co head^
Note: checking out 'head^'.

You are in 'detached HEAD' state. ...

$ git status --porcelain=v2 --branch
# branch.oid 095c52b044ae5e1b5317a2c626d8c50078a22130
# branch.head (detached)

$ git co -b "(detached)"
Switched to a new branch '(detached)'

$ git status --porcelain=v2 --branch
# branch.oid 095c52b044ae5e1b5317a2c626d8c50078a22130
# branch.head (detached)
```

test/git-strategies.test.js
kuychaco: oh yeah totally, good call 👍

lib/models/repository-states/present.js
BinaryMuse: I believe we've decided to forego handling this extremely edge case. :)

lib/git-shell-out-strategy.js
```diff
@@ -316,61 +317,31 @@ export default class GitShellOutStrategy {
...
-        }
-        if (y !== ' ') {
-          unstagedFiles[filePath] = statusMap[y];
-        }
-        if (x !== ' ' && x !== '?') {
-          stagedFiles[filePath] = statusMap[x];
+  async getStatusBundle() {
+    const output = await this.exec(['status', '--porcelain=v2', '--branch', '-z']);
+    const results = parseStatus(output);
+    const mapToNativePathSep = entries => {
```
BinaryMuse: Not a *huge* deal, but can we extract this function so a new copy doesn't get created for every invocation of `getStatusBundle`?

lib/git-shell-out-strategy.js
```diff
@@ -316,61 +317,31 @@ export default class GitShellOutStrategy {
...
+        if (entry.origFilePath) {
+          entry.origFilePath = toNativePathSep(entry.origFilePath);
         }
+        return entry;
+      });
+    };
+
+    for (const entryType in results) {
+      if (Array.isArray(results[entryType])) {
+        results[entryType] = mapToNativePathSep(results[entryType]);
```
BinaryMuse: Since this is a mutating function, can we just use a `forEach` instead of `map` and remove the reassignment here? Makes it clearer that we're mutating IMHO.

lib/models/repository-states/present.js
```diff
@@ -440,11 +507,14 @@ export default class Present extends State {
...
 
-  getCurrentBranch() {
-    return this.cache.getOrSet(Keys.currentBranch, async () => {
-      const {name, isDetached} = await this.git().getCurrentBranch();
-      return isDetached ? Branch.createDetached(name) : new Branch(name);
-    });
+  async getCurrentBranch() {
+    const {branch} = await this.getStatusBundle();
+    if (branch.head === '(detached)') {
+      const description = await this.git().describeHead();
```
BinaryMuse: Should we cache this in its own key?

lib/models/repository-states/state.js
```diff
@@ -296,6 +296,20 @@ export default class State {
...
+  getStatusBundle() {
+    return Promise.resolve({
+      stagedFiles: [],
+      unstagedFiles: [],
+      mergeConflictFiles: [],
+      headOid: null,
+      aheadCount: null,
+      behindCount: null,
+      branchName: null,
+      remoteName: null, // ??? AND anything else
```
BinaryMuse: Reminder for ☝️

lib/models/repository-states/present.js
kuychaco: Yeah it is super edge-casey. If we really wanted to address it we could first check to see if `(detached)` is listed in the results of `getBranches()`. But it seems like it might not be worth it to do this check every time we suspect we're in detached head state since it is such an edge case

lib/models/repository-states/present.js
```diff
@@ -87,18 +87,16 @@ export default class Present extends State {
...
+        keys.add(Keys.statusBundle);
         continue;
       }
 
       if (endsWith('.git', 'HEAD')) {
         keys.add(Keys.lastCommit);
-        keys.add(Keys.currentBranch);
-        keys.add(Keys.aheadCount.all);
-        keys.add(Keys.behindCount.all);
+        keys.add(Keys.statusBundle);
```
smashwilson: Do we need to invalidate `Keys.headDescription` here, too?

lib/models/repository-states/present.js
```diff
@@ -258,8 +257,8 @@ export default class Present extends State {
   // Remote interactions
 
   @invalidate(branchName => [
-    Keys.aheadCount.oneWith(branchName),
-    Keys.behindCount.oneWith(branchName),
+    Keys.statusBundle,
+    Keys.headDescription,
```
smashwilson: When will fetch or pull invalidate the `describe` output? Does that happen if you've checked out a remote tracking branch directly... ?

smashwilson: Yup, definitely needed. Don't mind me :grin:

kuychaco: If the nearest tag reachable from a commit while in detached head state is `origin/branch-name`, then a fetch or pull that updates a remote tracking branch could affect the commit description

Issue #822, opened 2017-05-16
Uncaught TypeError: Cannot read property 'parentNode' of null


ktornwall:
1. View a git diff for a file with at least two changes
2. Left click on one change
3. Right click on another change
4. Click away from the menu
--> Crash

![kapture 2017-05-16 at 19 10 37](https://cloud.githubusercontent.com/assets/1699388/26132018/68b815b8-3a6b-11e7-8ddf-6967485ff0e4.gif)

**Atom**: 1.18.0-beta0 x64
**Electron**: 1.3.15
**OS**: Mac OS X 10.11.6
**Thrown From**: [github](https://github.com/atom/github) package 0.0.9

### Stack Trace

Uncaught TypeError: Cannot read property 'parentNode' of null

```
At /Applications/Atom Beta.app/Contents/Resources/app/node_modules/github/lib/views/file-patch-view.js:206

TypeError: Cannot read property 'parentNode' of null
    at requestAnimationFrame (/Applications/Atom Beta.app/Contents/Resources/app/node_modules/github/lib/views/file-patch-view.js:206:1)
```

### Commands

```
```

### Non-Core Packages

```
autocomplete-elixir 1.5.14 
busy-signal 1.4.1 
editorconfig 2.2.2 
intentions 1.1.2 
language-elixir 0.17.0 
linter 2.1.4 
linter-eslint 8.1.6 
linter-flow 5.6.0 
linter-rubocop 2.1.1 
linter-ui-default 1.2.3 
vim-mode-plus-ex-mode 0.9.0 
```

smashwilson:
Good catch 🙇  This is fixed in #853, along with some other weirdness with the `contextmenu` event handling, and will be included in the next hotpatch release.

PR #853, opened 2017-05-19
File patch selection and context menu interaction

smashwilson:
Strictly speaking, the "actual fix" for #822 is a missing `event.persist()` call. In `FilePatchView.contextMenuOnItem()`, the MouseEvent is used in the `resend` callback, which is passed to `setState()` and may or may not be invoked asynchronously depending on React's scheduling mechanism. If you reproduce in dev mode you'll see a bunch of warnings to that effect in the console.

However... while I was reproducing this, I discovered that the right-click behavior no longer matched what the code was intended to do. We were _intending_ to add the right-clicked line or hunk to the selection if it was not already selected, but what was actually happening was that the context menu was appearing before the selection was modified.

This likely regressed in #617, because React's synthetic event system relies on event handlers added to the `document`, but Atom implements context menus internally by [adding an event handler to `document` _first_](https://github.com/atom/atom/blob/master/src/window-event-handler.coffee#L36). React's synthetic event callback fires too late to cancel the event.

To fix this, I've introduced a ContextMenuInterceptor component that installs a single native handler on the document with `{capture: true}` so that it'll trigger before Atom's handler.

I also had to change the event re-fire from a `requestAnimationFrame` to a `setImmediate` to actually allow the repaint with the updated selection to occur before the context menu is shown. I'm not entirely sure why that happens, but I suspect it has to do with [this bit from the documentation](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame):

> The window.requestAnimationFrame() method tells the browser that you wish to perform an animation and requests that the browser call a specified function to update an animation **before the next repaint.** The method takes as an argument a callback to be invoked before the repaint.

If I'm reading that right I bet two nested requestAnimationFrame() calls would do the trick as well... but setImmediate() works reliably for me and is somewhat cleaner.

### DIFF
--- a/lib/github-package.js
+++ b/lib/github-package.js
@@ -18,6 +18,7 @@ import StubItem from './atom-items/stub-item';
 import Switchboard from './switchboard';
 import yardstick from './yardstick';
 import GitTimingsView from './views/git-timings-view';
+import ContextMenuInterceptor from './context-menu-interceptor';
 import AsyncQueue from './async-queue';
 import WorkerManager from './worker-manager';
 
@@ -71,6 +72,7 @@ export default class GithubPackage {
           this.setActiveContext(WorkdirContext.absent());
         }
       }),
+      ContextMenuInterceptor,
     );
 
     this.setupYardstick();
--- a/lib/views/hunk-view.js
+++ b/lib/views/hunk-view.js
@@ -109,7 +109,6 @@ class LineView extends React.Component {
           className={`github-HunkView-line ${lineSelectedClass} is-${line.getStatus()}`}
           onMouseDown={event => this.props.mousedown(event, line)}
           onMouseMove={event => this.props.mousemove(event, line)}
-          onContextMenu={event => this.props.contextMenuOnItem(event, line)}
           ref={e => this.props.registerLineElement(line, e)}>
           <div className="github-HunkView-lineNumber is-old">{oldLineNumber}</div>
           <div className="github-HunkView-lineNumber is-new">{newLineNumber}</div>
--- a/lib/views/hunk-view.js
+++ b/lib/views/hunk-view.js
@@ -2,6 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {autobind} from 'core-decorators';
 
+import ContextMenuInterceptor from '../context-menu-interceptor';
+
 export default class HunkView extends React.Component {
   static propTypes = {
     hunk: PropTypes.object.isRequired,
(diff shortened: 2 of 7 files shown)


lib/github-package.js
```diff
@@ -71,6 +72,7 @@ export default class GithubPackage {
           this.setActiveContext(WorkdirContext.absent());
         }
       }),
+      ContextMenuInterceptor,
```
BinaryMuse: I'm not 100% sure I understand why this is necessary, if registrations only happen on mount and are removed on unmount.

lib/views/hunk-view.js
```diff
@@ -102,19 +104,21 @@ class LineView extends React.Component {
...
-        <div className="github-HunkView-lineNumber is-new">{newLineNumber}</div>
-        <div className="github-HunkView-lineContent">
-          <span className="github-HunkView-plusMinus">{line.getOrigin()}</span>
-          <span>{line.getText()}</span>
+      <ContextMenuInterceptor onWillShowContextMenu={event => this.props.contextMenuOnItem(event, line)}>
+        <div
+          className={`github-HunkView-line ${lineSelectedClass} is-${line.getStatus()}`}
+          onMouseDown={event => this.props.mousedown(event, line)}
+          onMouseMove={event => this.props.mousemove(event, line)}
+          onContextMenu={event => this.props.contextMenuOnItem(event, line)}
```
BinaryMuse: I'm not sure so thought I'd ask — can this be removed now that `ContextMenuInterceptor` wraps it?

smashwilson: :+1: :fire: Totally.

lib/github-package.js
smashwilson: The `ContextMenuInterceptor` uses a single native event listener attached to the `document` that's shared among all instances. Each instance registers and deregisters an element in a Map, but the event listener itself isn't removed until the static `ContextMenuInterceptor.dispose()` method is called. This bit :point_up: ensures that we don't leak listeners on package deactivation.

... Although, this might be overkill. I _could_ just register a new document listener for each instance and check `element.includes(event.target)`? I was intending to minimize the number of listeners we register but maybe a single listener that does an O(n) operation on a Map isn't any better.

smashwilson: I gave the simplification here a shot and our test suite leaked a _ton_ of event listeners, because Enzyme doesn't trigger `componentWillUnmount`. I think I'm just going to leave it as -is with the static Map.

PR #859, opened 2017-05-23
Fix regexp for GSOS#getRemotes

### LINKED ISSUE #858: Having a pushurl config in git make the package crash
[Theses lines](https://github.com/atom/github/blob/master/lib/git-shell-out-strategy.js#L634-L637) try to parse remotes and fails if there is a [pushurl config](https://git-scm.com/docs/git-push#_remotes_a_id_remotes_a).
The package fails with a `git-shell-out-strategy.js:871 Uncaught (in promise) TypeError: Cannot read property '1' of null(…)`

After a bit of investigation it seems like the [regex here](https://github.com/atom/github/blob/master/lib/git-shell-out-strategy.js#L629) is not correctly escaped and therefore match pushurl too.

```
$ git config --local --get-regexp '^remote..*.url$'
remote.origin.url git@github.com:paradoxxxzero/whatever.git
remote.origin.pushurl git@github.com:paradoxxxzero/whatever.git
```

```
$ git config --local --get-regexp '^remote\..*\.url$'
remote.origin.url git@github.com:paradoxxxzero/whatever.git
```

kuychaco:
Fixes #858

### DIFF
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -562,8 +562,7 @@ export default class GitShellOutStrategy {
   }
 
   async getRemotes() {
-    // eslint-disable-next-line no-useless-escape
-    let output = await this.getConfig(['--get-regexp', '^remote\..*\.url$'], {local: true});
+    let output = await this.getConfig(['--get-regexp', '^remote\\..*\\.url$'], {local: true});
     if (output) {
       output = output.trim();
       if (!output.length) { return []; }
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -626,7 +626,8 @@ export default class GitShellOutStrategy {
   }
 
   async getRemotes() {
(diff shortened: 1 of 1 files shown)


lib/git-shell-out-strategy.js
```diff
@@ -626,7 +626,8 @@ export default class GitShellOutStrategy {
   }
 
   async getRemotes() {
-    let output = await this.getConfig(['--get-regexp', '^remote..*.url$'], {local: true});
+    // eslint-disable-next-line no-useless-escape
+    let output = await this.getConfig(['--get-regexp', '^remote\..*\.url$'], {local: true});
```
smashwilson: I think you need to double-escape the `\` (hence the linter complaining about it):

```
> console.log('foo\.bar')
foo.bar
undefined
> console.log('foo\\.bar')
foo\.bar
undefined
```

kuychaco: oh good call. heh infinite escaping

Issue #309, opened 2016-11-08
Add support for submodules


kuychaco:
This involves adding submodule support to [`what-the-diff`](https://github.com/kuychaco/what-the-diff/issues/5)

kevinansfield:
I have the following git structure:

```
ghost (primary git repo)
|-- core/client (submodule)
|-- content/themes/casper (submodule)
```

I mostly work in the client by running `cd core/client` then `atom .`. All the `git` operations performed via terminal in that directory treat the submodule as if it is a separate repository which is what I want as it _is_ a completely separate project to the main repo.

However I've just tried the Atom Beta by running `atom-beta .` in the `core/client` directory and it's only picking up the top-level `Ghost` repository which isn't of any use because all the commits, PRs etc are on the submodule's repository.

Is this the same problem that this issue is referring to?

BinaryMuse:
> it's only picking up the top-level `Ghost` repository

Okay, @kuychaco and I just tracked this down (and learned something new about Git in the process. happens a lot on this project 😅)

So submodule working directories *do* contain a `.git` entry, but it's a file, not a folder, and the file contains some text that points to the *actual* repository directory, which lives in the containing repository's `.git/modules` directory. This is problematic for two reasons:

1. `WorkdirCache#walkToRoot` walks up the folder hierarchy trying to find a `.git` folder, and it fails if it finds one but it's not a file. This needs to be adjusted to account for the pseudo-symlink nature of the `.git` file in submodule working dirs.
2. We make assumptions throughout the codebase (probably mostly in `Repository` and friends and `GitShellOutStrategy`) that the `.git` folder is a subdirectory of the working directory (to read things like `MERGE_HEAD`, etc.) We should implement a smarter `getGitDirectoryPath` and use it everywhere instead.

---

I can't say for sure how long it'll be before we have full submodule support in Atom, but in the meantime we are trying to take care of some of the more egregious errors and problems (e.g. #828). I'll open some issues to track specific tasks. Thanks for your help tracking this all down! ✨

bhack:
Can we change or unset the milestone?

smashwilson:
Necroposting with an investigation node: submodule mount points appear in git diffs with a special filemode of `160000` (which just happens to be IF_DIR & IF_LNK). So we can detect 'em at least.

TheCherry:
Any update here? We use in our project multiple Submodules and the Atom Github package isn't still useful for us. Our structure looks like this:
- root_project/
  - .git/
[...]

PR #860, opened 2017-05-23
Support gitdir files

### LINKED ISSUE #839: Unable to treat submodule workdir as repository root
See https://github.com/atom/github/issues/309#issuecomment-302126199

Even when making a submodule working directory one of Atom's project roots, we still don't correctly find the submodule's `.git` folder (there's a `.git` file but it points to a folder in the parent's `.git/modules` folder), so we walk up the tree and find the parent repo instead. We need to:

1. Adjust `WorkdirCache` to account for this.
2. Fix `Repository#getGitDirectoryPath` and use it everywhere we're hardcoding a path to a `.git` folder.

BinaryMuse:
From the [Git docs](https://git-scm.com/docs/gitrepository-layout#_description):

> **Note:** Also you can have a plain text file `.git` at the root of your working tree, containing `gitdir: <path>` to point at the real directory that has the repository.

This PR adds support for this feature; most notably, this adds support for using submodules as root project folders in Atom and having everything work correctly (#839).

Fixes #839 
Ref #309

### DIFF
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -278,9 +278,9 @@ export default class GitShellOutStrategy {
       const output = await this.exec(['rev-parse', '--resolve-git-dir', path.join(this.workingDir, '.git')]);
       const dotGitDir = output.trim();
       if (path.isAbsolute(dotGitDir)) {
-        return dotGitDir;
+        return toNativePathSep(dotGitDir);
       } else {
-        return path.resolve(path.join(this.workingDir, dotGitDir));
+        return toNativePathSep(path.resolve(path.join(this.workingDir, dotGitDir)));
       }
     } catch (e) {
       return null;
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -272,13 +272,18 @@ export default class GitShellOutStrategy {
     });
   }
 
-  async isGitRepository() {
+  async resolveDotGitDir() {
     try {
       await fsStat(this.workingDir); // fails if folder doesn't exist
-      await this.exec(['rev-parse', '--resolve-git-dir', path.join(this.workingDir, '.git')]);
-      return true;
+      const output = await this.exec(['rev-parse', '--resolve-git-dir', path.join(this.workingDir, '.git')]);
+      const dotGitDir = output.trim();
+      if (path.isAbsolute(dotGitDir)) {
+        return dotGitDir;
+      } else {
+        return path.resolve(path.join(this.workingDir, dotGitDir));
+      }
     } catch (e) {
-      return false;
+      return null;
     }
   }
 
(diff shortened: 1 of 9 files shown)


lib/models/workdir-cache.js
```diff
@@ -66,8 +68,12 @@ export default class WorkdirCache {
           }
 
           if (!stat.isDirectory()) {
-            // A file called ".git". Probably not a good idea, but still not a git working directory.
```
smashwilson: Haha, little did I know.

lib/models/workdir-cache.js
```diff
@@ -66,8 +68,12 @@ export default class WorkdirCache {
...
-            return walk();
+            const contents = await readFile(dotGit, 'utf8');
+            if (contents.startsWith('gitdir: ')) {
+              return resolve(currentDir);
+            } else {
+              return walk();
+            }
```
smashwilson: :+1:

Although, I am wondering if it's still worth it to keep this as native filesystem operations as opposed to shelling out to `git rev-parse --show-toplevel`, since it does more than test for `.git` directories.

For one thing, we'll be slightly inconsistent with the command-line in that we won't handle [`GIT_CEILING_DIRECTORIES`](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_repository_locations), for example.

test/git-strategies.test.js
```diff
@@ -28,14 +28,25 @@ import {fsStat, normalizeGitHelperPath, writeFile} from '../lib/helpers';
...
 
         fs.removeSync(path.join(workingDirPath, '.git'));
-        assert.isFalse(await git.isGitRepository(workingDirPath));
+        assert.isNull(await git.resolveDotGitDir(workingDirPath));
+      });
+
+      it('supports gitdir files', async function() {
+        const workingDirPath = await cloneRepository('three-files');
+        const workingDirPathWithDotGitFile = await getTempDir();
+        await writeFile(path.join(workingDirPathWithDotGitFile, '.git'), `gitdir: ${path.join(workingDirPath, '.git')}`);
```
smashwilson: :+1: :cool:

lib/git-shell-out-strategy.js
```diff
@@ -445,9 +450,9 @@ export default class GitShellOutStrategy {
     return this.gpgExec(['merge', branchName], {writeOperation: true});
   }
 
-  async isMerging() {
+  async isMerging(dotGitDir) {
```
BinaryMuse: Interesting, I felt the opposite actually. Wanted to ensure it got kept in a single place, and the state-machine nature of the `Repository` made it easy to ensure that the value was set before we start calling other methods.

lib/models/repository.js
```diff
@@ -140,8 +140,16 @@ export default class Repository {
...
+  setGitDirectoryPath(gitDirectoryPath) {
+    this._gitDirectoryPath = gitDirectoryPath;
+  }
+
   getGitDirectoryPath() {
-    return path.join(this.getWorkingDirectoryPath(), '.git');
+    if (this._gitDirectoryPath) {
+      return this._gitDirectoryPath;
+    } else {
+      return path.join(this.getWorkingDirectoryPath(), '.git');
```
BinaryMuse: I left this mostly for tests in case they didn't go through the normal setup process — but perhaps they all do? I didn't try it.

lib/models/workdir-cache.js
BinaryMuse: Yeah, I'm down for changing this later to better match Git's behavior.

lib/git-shell-out-strategy.js
smashwilson: Maybe we should inject _both_, and keep the strategy itself stateless?

Issue #816, opened 2017-05-16
Split diffs


smashwilson:
From smashwilson/merge-conflicts#247:

> `git mergetool` creates files for each conflicting side and merge base.
>
> Example:
> - app/views/foobar.html_BASE_17711.slim
> - app/views/foobar.html_LOCAL_17711.slim
> - app/views/foobar.html_REMOTE_17711.slim
>
> These files are useful to determine what was actually changed (by comparing base to local change or remote change), to resolve conflict.
>
> There is [split-diff](https://atom.io/packages/split-diff) plugin, so we only need to create above files (like mergetool) and open split diff.

Split diffs or alternate diff views is something that I'd always kind of wanted to investigate, but never had the time. I'm guessing this will be a common request 😆

simurai:
> I'm guessing this will be a common request

💯  Not just handy for merge conflicts, but also for a mini time-machine while coding. Like "how was it before I started to mess everything up". The [split-diff](https://atom.io/packages/split-diff) package is really great for that.

fuzzyweapon:
Yeah I always had troubles with using/reading merge conflict tools that have all changes inline even after extended use (it's one of the reasons I use gui apps to resolve conflicts rather than just sticking to a command-line editor, among other niceties that good split diff style tools provide).

I come from an Intellij background, and aside from them having some usability issues on where they could unify some of their code time machine like features with their main side-by-side differ editors, I feel like they do some things right that I never had to "get used to."

Anyways, even basic solid side-by-side for merge conflicts would be fantastic and make it easier to teach newbies to coding. :)

PR #866, opened 2017-05-24
Implement split diffs

BinaryMuse:
This PR implements split diffs.

unstaged_changes__spec_text-editor-component-spec_js_ ___atom

github-package-unified-split

This is a pretty basic implementation so far and needs some more work:

- [x] Need to adjust colors to work in all themes
- [ ] Splitting lines into chunks could probably be more performant — this was a rough first pass
- [ ] The split mode option should be persistent (and come from `GitController` via props)
- [ ] There are still some odd visual artifacts because I'm bad at CSS
  git_ ___atom_and_unstaged_changes__styles_hunk-view_less_ ___github_github
- [ ] Need to ensure "no newline at end of file" works correctly in split diff mode
- [ ] Tests

Refs (but doesn't totally resolve, I don't think) #816


smashwilson:
whoa

simurai:
> There are still some odd visual artifacts

Next to the gap when the left line starts wrapping, there are also these 1px gaps:

screen shot 2017-05-26 at 1 38 21 pm

They are only visible sometimes. I think it's because table layout rounds to whole pixels and flexbox keeps sub-pixels. :weary:

Another problem is that if the line numbers are bigger than the minimum width, they don't line up anymore. That's because each line number is nested inside multiple elements and isn't a direct child of each row. So the table layout breaks.

Something we could try. Instead of:

```html
<div class="github-HunkView-split-line">
  <div class="left-line">
    <div>
      <div class="github-HunkView-line is-deleted">
        <div class="github-HunkView-lineNumber is-old">12</div>
        <div class="github-HunkView-lineContent"></div>
      </div>
    </div>
  </div>
  <div class="right-line">
    <div>
      <div class="github-HunkView-line is-selected is-added">
        <div class="github-HunkView-lineNumber is-new">12</div>
        <div class="github-HunkView-lineContent"></div>
      </div>
    </div>
  </div>
</div>
<div class="github-HunkView-split-line">
  <div class="left-line">
    <div>
      <div class="github-HunkView-line is-deleted">
        <div class="github-HunkView-lineNumber is-old">12</div>
        <div class="github-HunkView-lineContent"></div>
      </div>
    </div>
  </div>
  <div class="right-line">
    <div>
      <div class="github-HunkView-line is-selected is-added">
        <div class="github-HunkView-lineNumber is-new">12</div>
        <div class="github-HunkView-lineContent"></div>
      </div>
    </div>
  </div>
</div>
```

Move all the cells into the **same** row (`.github-HunkView-line`):

```html
<div class="github-HunkView-line">
  <div class="github-HunkView-lineNumber is-deleted is-old">12</div>
  <div class="github-HunkView-lineContent is-deleted"></div>
  <div class="github-HunkView-lineNumber is-added is-new">12</div>
  <div class="github-HunkView-lineContent is-added"></div>
</div>
<div class="github-HunkView-line">
  <div class="github-HunkView-lineNumber is-deleted is-old">12</div>
  <div class="github-HunkView-lineContent is-deleted"></div>
  <div class="github-HunkView-lineNumber is-added is-new">12</div>
  <div class="github-HunkView-lineContent is-added"></div>
</div>
```

Unified view, would still be very similar with only 1 `github-HunkView-lineContent` per line.

Maybe try on a separate branch, this is only a theory and might have its own problems.

matthewdias:
rip

Issue #864, opened 2017-05-24
Improve changed files list and diff view UX


kuychaco:
In an effort to address #606, #645, #854 and related UX issues around staging/unstaging/discarding, I'd like to start a quick UX discussion first to guide development.

As part of this, the plan is to keep the file diff view open rather than closing it abruptly as is currently the case. Here are the UX interactions I’ve thought through and I’d be curious to hear what others think 💭, especially @simurai. It’s quite likely that I’ve missed something that’s worth considering

**Changed list view UX**
* after hitting enter to stage file, select next file
  * if diff view is open, display next selected file 
  * if diff view is closed, display no file diff
* when there is no other file in list, go to next list and display diff
  * Note: this automatic shift from one type of changes (staged vs unstaged) to another might be unexpected to users, or confusing if they don’t notice that there has been a switch of change type 
    * might be able to address this by styling the diff view headers to make the change type more obvious

**Diff view UX**
* after staging all hunks
  * keep empty diff view open, offer buttons and keyboard shortcuts to 
    * go to the next diff in the list of changed files
    * go to staged/unstaged version 

**Questions**
* _when git panel is closed, should we close file diff?_ I think not. because it could be that the user just wants more screen space to view the diff so they close the panel

simurai:
This sounds like a great plan. Then as soon as you start using the arrow keys, the diff view stays until closed.

> might be able to address this by styling the diff view headers to make the change type more obvious

Yeah, that's a good idea. Currently the diff view takes some effort to know in which list you're in. When reading the title or button labels, I always have to think for a bit what it means. A simple ✅  checkmark in front of "Staged Changes" might already help orient better? I'll try out some options.

> _when git panel is closed, should we close file diff?_ I think not. because it could be that the user just wants more screen space to view the diff so they close the panel

Hmmm... first thought was yes, but you found the only edge case. 😜 Actually, I can think of one more. You could keep the diff view open in a split pane and watch the diff change while coding. To have a handy reference what changed. Kinda like the [split-diff](https://atom.io/packages/split-diff) package.

Ohh.. and you could move the diff-view into the same dock as the Git panel, split to the left. Then you can show/hide them together:

![diff-view](https://cloud.githubusercontent.com/assets/378023/26406119/4e515654-40d2-11e7-844b-f580e97dc12c.gif)

Maybe we can think of something useful to show when the diff view is empty. Some Git tips. Or keybinding cheat-sheet of this package.

kuychaco:
[...]

PR #867, opened 2017-05-24
Add button for discarding hunk/selection in HunkView

### LINKED ISSUE #604: Add "Discard Hunk/Selection" button
The discard feature is awesome, but I worry many users might not discover it via the context menu. I think it warrants its own button next to the "Stage Hunk/Selection" button.

/cc @kuychaco

kuychaco:
Fixes #604 

/cc @simurai for styling 💭 

I'm a bit concerned that the button is so close to the "Stage/Unstage" button and therefore might accidentally get pressed. Maybe we can move it elsewhere in the hunk header? But even if it does accidentally get pressed then the user can always undo the discard.

discard-hunk

### DIFF
--- a/lib/views/file-patch-view.js
+++ b/lib/views/file-patch-view.js
@@ -137,6 +137,7 @@ export default class FilePatchView extends React.Component {
             return (
               <HunkView
                 key={hunk.getHeader()}
+                tooltips={this.props.tooltips}
                 hunk={hunk}
                 isSelected={selectedHunks.has(hunk)}
                 hunkSelectionMode={hunkSelectionMode}
--- a/lib/views/file-patch-view.js
+++ b/lib/views/file-patch-view.js
@@ -130,6 +130,7 @@ export default class FilePatchView extends React.Component {
               stageButtonSuffix += 's';
             }
             const stageButtonLabel = stageButtonLabelPrefix + stageButtonSuffix;
+            const discardButtonLabel = 'Discard' + stageButtonSuffix;
 
             return (
               <HunkView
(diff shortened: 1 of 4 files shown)


simurai:
> I'm a bit concerned that the button is so close to the "Stage/Unstage" button and therefore might accidentally get pressed. Maybe we can move it elsewhere in the hunk header?

### Option A

Could be smaller, maybe just an icon:

screen shot 2017-05-25 at 5 11 56 pm


### Option B

Have the stage/unstage button on the left. Also, "Stage All". Then the good things are on the left and the evil ones on the right:

screen shot 2017-05-25 at 5 22 38 pm

--- 

✋  I tend towards Option B. It would increase the distance to travel from the lists to the stage button. Maybe ok. You can move the Git panel to the left dock, if necessary. I might do that myself, just have to override the arrow keybindings first.

kuychaco:
> Then the good things are on the left and the evil ones on the right:

Haha I like that separation.

> It would increase the distance to travel from the lists to the stage button. Maybe ok.

How about swapping the positions so that the good things are on the right, and the evil ones are further away on the left and harder to get to.

simurai:
> How about swapping the positions so that the good things are on the right, and the evil ones are further away on the left and harder to get to.

Hmm.. that feels a bit weird because the left side is the "main area" with the title and the blue focus. Seeing discard right above might look like it's the "main" action.

screen shot 2017-05-25 at 7 50 03 pm

We can also merge this PR as is, then do a bigger juggling things around in another one. Together with the "it's too hard to know in which list you are" #864. And maybe more.

kuychaco:
> Hmm.. that feels a bit weird because the left side is the "main area" with the title and the blue focus. Seeing discard right above might look like it's the "main" action.

Yeah, good point. Moving the stage buttons all the way to the left feels a bit extreme to me. Don't know how much of that is my being used to having them on the right and how much of it is that it just makes more sense intuitively that they be there. It feels like the right side is more accessible and natural, there's less friction to take the action.

For now I've implemented Option A which I think is a really nice compromise. 

Thanks for your thoughts and mockups @simurai :bowing_woman:

PR #871, opened 2017-05-25
Style diff selections

### LINKED ISSUE #649: Selected hunk line is not visible after clicking Stage/Unstage button
1. Make edits to a given file.  
2. Open the staged changes view
3. select a few things and click stage selection

Note: notice after this the stage selection is still there, when it should change to stage hunk?
If you click on it it will stage the last spot  you had selected, but doesn't show selected on ui.  

**Edit:** after looking closely at this it seems the current behavior is as expected, but our current styling does not show the next selected line since the `FilePatchView` loses its `:focus` when the button is clicked

image

### Versions

Atom Version (`atom --version`):

GitHub Package Version (`git --git-dir ~/.atom/packages/github/.git rev-parse head`):

simurai:
### Description of the Change

Adds a background color to selections in the diff-view.

selection

### Benefits

Let's you see what is selected even when the stage button eats the focus. Makes it consistent with the list selection/focus.

### Applicable Issues

Closes #649

### DIFF
--- a/styles/hunk-view.less
+++ b/styles/hunk-view.less
@@ -77,6 +77,16 @@
 // States
 // -------------------------------
 
+.github-HunkView.is-selected.is-hunkMode .github-HunkView-header {
+  background-color: @background-color-selected;
+  .github-HunkView-title {
+    color: @text-color;
+  }
+  .github-HunkView-stageButton {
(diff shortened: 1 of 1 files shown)


kuychaco:
@simurai :bowing_woman: ✨

Issue #872, opened 2017-05-25
Start-up Behavior


smashwilson:
## The problem

There are three competing factors we need to balance in our start-up behavior:

1. **Discoverability.** Users who launch the next stable version of Atom should be presented with a clear reminder that this package exists. They shouldn't need to hunt around online to discover how to access it.
2. **Screen real estate.** Especially for users who are downloading Atom for the first time and being presented with the Welcome panes, screen real estate is at a premium. Launching with the right dock expanded _and_ the tree-view open by default squashes the Welcome panes and is visually overwhelming: <img width="1420" alt="screen shot 2017-05-25 at 9 21 49 am" src="https://cloud.githubusercontent.com/assets/17565/26453311/4f966bc6-4131-11e7-8519-c24e10cf3d15.png">
3. **Likelihood of use.** When first launching Atom, users are more likely to need to interact with the tree-view right away to explore the current project than they are to perform a git operation. In the general case, they're likely to close the panel right away most of the time.

Our current behavior, introduced in #685, is to activate the git and GitHub tabs on the first run *of any project*. That achieves (1) but I think we all agree is a bit aggressive for (2) and (3).

## Use cases

I believe we have two categories of users to consider:

* Existing Atom users who are upgraded to 1.18.0.
* Users downloading Atom 1.18.0 as their first Atom version.

We can probe `atom.config.get('welcome.showOnStartup')` as a proxy for detecting this.

## Proposed solution

Here's what I'd like to implement:

_If the Welcome panes are active_ default the git and GitHub tabs to _off_ to account for (2). Record that the first launch has occurred in the package config. Rely on the Welcome guide section proposed below to account for (1).

_If the Welcome panes are not active_ and _auto-launch hasn't happened yet in this Atom install_, default the tabs to "on" to aid (1), then record that the first launch has occurred.

_If the Welcome panes are not active_ and _auto-launch has already happened in this Atom install_, default the tabs to "off" to account for (3).

:spiral_notepad: Note that this behavior only applies if deserialized project state is not present. If there is project state, the git and GitHub tabs will continue to respect their last serialized state.

:spiral_notepad: Detecting whether or not _auto-launch has already happened_ requires a place to store project-independent, global state that's "owned" by the package. ~I've opened atom/settings-view#954 to address this.~ To accomplish this, I'm using a flag file in the Atom config directory.

## Alternatives

[...]

PR #875, opened 2017-05-25
Start-up behavior

smashwilson:
This implements the proposed startup behavior described in #872.

* If this is the first Atom launch with the github package installed, but the Welcome pane will be shown, start _closed_. atom/welcome#62 will make it redundant in any case.
* If this is the first Atom launch and the Welcome pane is not shown, start _open_.
* If this is not the first Atom launch, start _closed_.

Uses an (empty except for a comment, right now) flag file called `~/.atom/github.cson` to detect prior runs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,6 +145,12 @@
       "type": "boolean",
       "default": false,
       "description": "Capture CPU profiles"
+    },
+    "firstRun": {
+      "type": "boolean",
+      "default": true,
+      "description": "Display the Git and GitHub tabs on Atom launch. Reverts to false on each launch.",
+      "hidden": true
     }
   },
   "deserializers": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -145,6 +145,12 @@
       "type": "boolean",
       "default": false,
       "description": "Capture CPU profiles"
+    },
+    "firstRun": {
+      "type": "boolean",
+      "default": true,
+      "description": "Display the Git and GitHub tabs on Atom launch. Reverts to false on each launch.",
+      "hidden": true
```
smashwilson: Oops, I forgot to :fire: this.

smashwilson:
> So the hidden firstRun config option won't actually be hidden for now right? I don't see any PR in atom/atom to account for this new field.

Oops, that config option shouldn't be there. I replaced the functionality with the `~/.atom/github.cson` flag file, but I neglected to ✂️ the option.

13121997:
Thank

PR #876, opened 2017-05-26
Show committer avatar as well as author in the PR commit view

### LINKED ISSUE #774: Show committer avatar as well as author in commit list
We currently only show the author of the commits in the PR commit view.

cursor_and_pr__atom_atom_13453_ use_new_text-buffer_version_with_a_more_efficient_displaylayer_ ___atom

It would be great to show the committer too, similar to how GitHub.com does it:

cursor_and_use_new_text-buffer_version_with_a_more_efficient_displaylayer_by_maxbrunsfeld_ _pull_request__13453_ _atom_atom

ahmed-taj:
Fixes #774 

**Todo list**

- [x] Retrieve committer info 
- [x] Show committer's avatar in the commit view
- [x] Fix design (e.g. pull avatars to the right as [GitHub](https://cloud.githubusercontent.com/assets/189606/25895776/a5eb61cc-3581-11e7-9007-ae6c4530ea5f.png) does)
- [x] Add tests

### DIFF
--- a/lib/containers/timeline-items/commit-container.js
+++ b/lib/containers/timeline-items/commit-container.js
@@ -9,16 +9,33 @@ export class Commit extends React.Component {
     item: PropTypes.object.isRequired,
   }
 
+  authoredByCommitter(commit) {
+    if (commit.authoredByCommitter) {
+      return true;
+    }
+    // If you commit on GitHub online the committer details would be:
+    //
+    //    name: "GitHub"
+    //    email: "noreply@github.com"
+    //
+    // Do we need to compare the name too ?
+    if (commit.committer.email === 'noreply@github.com') {
+      return true;
+    }
+    return false;
+  }
+
   renderCommitter(commit) {
-    if (!commit.authoredByCommitter) {
+    if (!this.authoredByCommitter(commit)) {
       return (
         <img
           className="author-avatar" src={commit.committer.avatarUrl}
           title={commit.committer.user ? commit.committer.user.login : commit.committer.name}
         />
       );
+    } else {
+      return null;
     }
-    return null;
   }
   render() {
     const commit = this.props.item;
--- a/lib/containers/timeline-items/commit-container.js
+++ b/lib/containers/timeline-items/commit-container.js
@@ -42,11 +42,13 @@ export class Commit extends React.Component {
     return (
       <div className="commit">
         <Octicon className="pre-timeline-item-icon" icon="git-commit" />
-        <img
-          className="author-avatar" src={commit.author.avatarUrl}
-          title={commit.author.user ? commit.author.user.login : commit.author.name}
-        />
-        {this.renderCommitter(commit)}
+        <span className="commit-author">
+          <img
+            className="author-avatar" src={commit.author.avatarUrl}
+            title={commit.author.user ? commit.author.user.login : commit.author.name}
+          />
+          {this.renderCommitter(commit)}
+        </span>
         <span
           className="commit-message-headline"
           title={commit.messageHeadline}
--- a/lib/containers/timeline-items/commit-container.js
+++ b/lib/containers/timeline-items/commit-container.js
@@ -17,11 +17,15 @@ export class Commit extends React.Component {
     //
     //    name: "GitHub"
     //    email: "noreply@github.com"
+    //    user: null
     //
-    // Do we need to compare the name too ?
     if (commit.committer.email === 'noreply@github.com') {
       return true;
     }
+    if (commit.committer.name === 'GitHub' && commit.committer.user === null) {
+      return true;
+    }
+
     return false;
   }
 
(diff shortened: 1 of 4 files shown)


ahmed-taj:
@BinaryMuse I can't see why Travis is failing above. However, everything is OK in my [fork Travis test result](https://travis-ci.org/ahmed-taj/atom-github)

simurai:
> Fix design (e.g. pull avatars to the right as GitHub does)

Switched to a table layout so that the avatars line up. [Example](https://github.com/atom/github/pull/726#commits-pushed-cfcd149):

screen shot 2017-05-26 at 5 45 05 pm

ahmed-taj:
> Switched to a table layout so that the avatars line up

@simurai I was trying to do the same locally, but I wouldn't make it perfect like your work, good CSS lesson ;)

> but the failing was probably unrelated.

Good to know, thanks

ahmed-taj:
Oops, looks like we broke something, just noticed it when testing against [atom/atom#13453](https://github.com/atom/atom/pull/13453)

screenshot from 2017-05-26 14-42-38

I will fix it this time :smile:

kuychaco:
Yeah, apologies about the Travis failure. We have some flakey tests. We can restart the failing job for you if needed

Thanks for taking this on :)

ahmed-taj:
I would like to add some tests for [`Commit`](https://github.com/atom/github/blob/master/lib/containers/timeline-items/commit-container.js) container, but I can't figure out how!

I tried this code (after I dug into `commits-container.test.js`) to check how things work:

```jsx
describe('CommitContainer', function() {
  it('renders a header with one user name', function() {
    const item = {
      author: {name: 'FirstName', avatarUrl: '', user: {login: 'FirstLogin'}},
      committer: {name: 'GitHub', avatarUrl: '', user: null},
      authoredByCommitter: true,
      oid: 'xxx',
      messageHeadline: ':smile: it\'s fun',
      messageHeadlineHTML: 'fake html',
    };
    const app = <Commit item={item} />;
    const instance = shallow(app);
    assert.match(instance.text(), /FirstLogin/);
  });
});
```
But the assertion fails with:
```
AssertionError: assert.match(instance.text(), /FirstLogin/): expected '<Octicon />xxx' to match /FirstLogin/
```
Any hint ? It's my first time to play with Relay and Enzyme

ahmed-taj:
Ping @kuychaco :)

BinaryMuse:
Sorry I haven't had a chance to dig into this — first day back after a holiday in the US.

> I would like to add some tests for Commit container

I don't know if it's relevant to your failure or not, but the containers are responsible for reaching into the Relay store and such to fetch data from GraphQL. It's likely we'll want to separately export the non-containerized React component and test that directly instead.

ahmed-taj:
Oh, my bad I meant Commit component.

> It's likely we'll want to separately export the non-containerized React component and test that directly instead. 

Isn't it already done in [Commits](https://github.com/atom/github/blob/master/test/containers/commits-container.test.js) component @BinaryMuse? Actually, I was trying to do the same with the Commit component.

I think I will give it another shot and see if I can manage it to work

Thanks

BinaryMuse:
@ahmed-taj Ah I see, I'm sorry for the misinformation.

In the example you posted above, `FirstLogin` will never actually appear in the component's text because it's not rendered as text — it's rendered as an attribute of one of the images. For this test, you'll likely want to use one of Enzyme's other methods; here's a list of the ones provided by `shallow` instances: https://github.com/airbnb/enzyme/blob/master/docs/api/shallow.md

For example, to ensure that there's an image with the title attribute 'FirstLogin', you could use:

```
assert.isTrue(instance.containsMatchingElement(<img title="FirstLogin" />));
```

ahmed-taj:
I think it's ready for review now!

lib/containers/timeline-items/commit-container.js
```diff
@@ -9,15 +9,50 @@ export class Commit extends React.Component {
...
+    //    name: "GitHub"
+    //    email: "noreply@github.com"
+    //    user: null
+    //
+    if (commit.committer.email === 'noreply@github.com') {
+      return true;
+    }
+    if (commit.committer.name === 'GitHub' && commit.committer.user === null) {
+      return true;
+    }
```
BinaryMuse: Nice! I actually had no clue this was how it worked.

I dug into the GItHub.com source code a bit and found that the committer name will be `"GitHub Enterprise"` if we're running on GitHub enterprise. The noreply address is also based off the domain of the Enterprise install. I'm going to see if I can get a field added called `authoredViaWeb` or something that matches the internal check we do, but until then this works great. 👍

ahmed-taj: > I'm going to see if I can get a field added called authoredViaWeb or something that matches the internal check we do

This would be perfect :+1:

PR #877, opened 2017-05-26
Display warning notification when pull results in merge conflicts

### LINKED ISSUE #821: Make expected error cases friendlier such as pulling with merge conflicts
I tried to pull a branch from github.com into my local repo with Atom. There were some merge conflicts.

The error message I got was:

```
From https://github.com/Haacked/todo-app
 * branch            haacked-tries-to-code -> FETCH_HEAD
   4234f31..808d71d  haacked-tries-to-code -> origin/haacked-tries-to-code
```

It was reported using the standard red error message box. It made me feel like I made a mistake or that I broke Atom.

Since pulling branches with conflicts is an expected use case, we should provide a friendlier message about what just happened and perhaps a link to more documentation on resolving merge conflicts.

kuychaco:
Fixes #821

I started implementing a fix for #821 that involves using a warning notification rather than an error notification when there are merge conflicts as a result of a pull. Here's what it looks like:

pull-merge-conflicts

I'm thinking that a notification might not be that helpful, and that it might be more of an annoyance. In my opinion it's pretty clear what's happening based on the fact that merge conflicts appear and the commit box is populated with a message:

```
Merge branch 'master' of https://github.com/owner/repo

# Conflicts:
#	file.txt
```

Another alternative I'm considering is simply removing the notification.

### DIFF
--- a/lib/controllers/status-bar-tile-controller.js
+++ b/lib/controllers/status-bar-tile-controller.js
@@ -274,8 +274,10 @@ export default class StatusBarTileController extends React.Component {
         } else if (/Automatic merge failed; fix conflicts and then commit the result./.test(error.stdOut)) {
           return {
             type: 'Warning',
-            message: 'Automatic merge after pull failed',
-            description: 'Please fix conflicts and then commit the result.',
+            message: 'Your pull resulted in merge conflicts',
+            description: `Your last pull resulted in merge conflicts between your local changes
+              and the changes on the remote. You can resolve the conflicts with the Git panel
+              and commit them to continue.`,
           };
         }
 
--- a/lib/controllers/status-bar-tile-controller.js
+++ b/lib/controllers/status-bar-tile-controller.js
@@ -194,17 +194,26 @@ export default class StatusBarTileController extends React.Component {
     });
   }
 
-  async attemptGitOperation(operation, errorTransform = message => ({message})) {
+  async attemptGitOperation(operation, errorTransform = error => ({message: error.stdErr})) {
     const operationPromise = operation();
     try {
       return await operationPromise;
     } catch (error) {
       if (!(error instanceof GitError)) { throw error; }
-      const {message, description} = errorTransform(error.stdErr);
-      this.props.notificationManager.addError(
-        message || 'Cannot complete remote interaction',
-        {description, dismissable: true},
-      );
+      const {type, message, description} = errorTransform(error);
+      if (type === 'Error') {
+        this.props.notificationManager.addError(
+          message || 'Cannot complete remote interaction',
+          {description, dismissable: true},
+        );
+      } else if (type === 'Warning') {
+        this.props.notificationManager.addWarning(
+          message || 'Cannot complete remote interaction',
+          {description, dismissable: true},
+        );
+      } else {
+        throw new Error(`Notification type ${type} not currently supported.`);
+      }
       return null;
     }
   }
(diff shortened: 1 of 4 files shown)


simurai:
> Another alternative I'm considering is simply removing the notification.

- I think that would be ok if the Git panel is already opened
- But if the Git panel is closed, show the notification. Otherwise it's hard to guess why that ⬇️  `1` doesn't turn into `0`.
  - Or just open the Git panel?

Actually, maybe better to temporarily replace the ⬇️  1 ⬆️  1 with a :warning: `Merge conflict`. Clicking it would open the Git panel. Then people know what's going on and don't get annoyed by a notification or auto-opening the Git panel.

50Wliu:
:+1: on switching the push/pull status to a merge conflict warning.

kuychaco:
> Actually, maybe better to temporarily replace the ⬇️ 1 ⬆️ 1 with a ⚠️ Merge conflict. Clicking it would open the Git panel (if not already opened).

Oooh interesting idea. After thinking a bit about how this would look I have a couple of concerns around tradeoffs - mainly inconsistency, loss of data and access to functionality. 

**Concerns:**
I feel like the ⬇️ 1 ⬆️ 1 and ⚠️ `Merge conflict` are independent and convey different information to the user. They each have meaning outside of this particular scenario where the user pulls and there is a resulting merge conflict and they just happen to intersect.

If we add a ⚠️ `Merge conflict` then I think it would make sense to have this displayed whenever there is a merge conflict, otherwise it might seem inconsistent to only have it during some conflict situations. But with this I think about possibly cluttering up the status bar unnecessarily... is it really that helpful or necessary to have a status bar indication of merge conflicts?

And I hesitate to replace the ahead/behind count because that would again make for an inconsistent experience where sometimes it's there, sometimes it's not, and perhaps it's not immediately obvious to the user that it was replaced by the ⚠️ `Merge conflict` indicator. We also take away information about the ahead/behind count and make the `fetch` button inaccessible by doing this. 

So if anything I'd advocate for having both shown independently, but I'm not confident that it's worth it to add the ⚠️ `Merge conflict` indicator. And even then I think that the notification conveys additional useful information. After testing things out a bit I personally don't feel the notification is too obtrusive.

**Notification good enough for now?**
In my opinion the notification is the clearest indicator of what's happening. It explicitly describes what happened (automatic merge after pull failed) and what you need to do next (fix conflicts and then commit results). For super new Git users this might be helpful and informative. 

file2_txt_ ___src_test-repo

So I think I'm inclined to leave this as-is for now. Happy to revisit it if people feel strongly about it.

simurai:
> For super new Git users this might be helpful and informative.

Hmm.. not sure if **super** new Git users would know what to do. 😜 But yeah, since it's already an improvement, we can revisit in another PR.

BinaryMuse:
I'm personally not sure a status-bar icon is enough for this scenario. If the Git panel is closed, and the user pulls and gets merge conflicts, and the *only* thing that happens is a small icon appears in the status bar, it's very likely they'll continue with their work never realizing there's a merge conflict.

BinaryMuse:
@kuychaco Maybe the language in the notification could be improved? Perhaps something like:

| Your pull resulted in merge conflicts |
|:---|
| Your last pull resulted in merge conflicts between your local changes and the changes on the remote. You can resolve the conflicts with the Git panel and commit them to continue. If you need help, check out [this article on github.atom.io](https://github.atom.io/resolving-merge-conflicts). |

(where we would need to create the associated help article — or maybe even embed it in the package as a pane item?)

kuychaco:
All great suggestions. I think for now I'll do the following:

* add ⚠️ icon to status bar changed file count tile when there are merge conflicts
* use improved language in notification (but punt on the help article to tackle later)
* toggle Git panel open when pull results in conflicts

lib/controllers/status-bar-tile-controller.js
```diff
@@ -251,18 +262,28 @@ export default class StatusBarTileController extends React.Component {
...
               '<br>Please commit your changes or stash them before you merge.',
           };
+        } else if (/Automatic merge failed; fix conflicts and then commit the result./.test(error.stdOut)) {
+          this.props.ensureGitTabVisible();
+          return {
+            type: 'Warning',
+            message: 'Your pull resulted in merge conflicts',
+            description: `Your last pull resulted in merge conflicts between your local changes
+              and the changes on the remote. You can resolve the conflicts with the Git panel
+              and commit them to continue.`,
```
smashwilson: To bikeshed this message a little more, how about:

```js
{
  message: 'Merge conflicts',
  description: `Your local changes conflicted with changes made on the remote branch. Resolve the conflicts
    with the Git panel and commit to continue.`
}
```

Same basic message content, but (a) slightly shorter and (b) imperative tense rather than second-person.

I don't know, maybe that's too terse. What do you think?

kuychaco: I actually brevity makes more sense for a notification. Easier to scan, and not feel overwhelmed by text. Thanks for the great suggestions :)

PR #878, opened 2017-05-26
Allow user to discard all changes and fix context menus

### LINKED ISSUE #611: Would like a Discard All option
Unstaged Changes title bar where `Stage All` lives is a bit crowded already. Maybe either right-clicking Stage All to see a Discard All menu button or holding down alt to see the text change from Stage All to discard all?  Thoughts @simurai

### LINKED ISSUE #751: Choosing "Stage" or "Unstage" from context menu can do the wrong thing
To reproduce:

1. Select a file in the "Unstaged Changes" list
2. Right-click the empty space in the "Staged Changes" list
3. Select "Unstage"

Expected:

Nothing happens as nothing in the Staged Changes list was selected

Observed:

The selected file from the Unstaged Changes list is staged.

github-package-weird-list-menus

This happens because we dispatch `core:confirm` for both the "Stage" and "Unstage" context menu items.

kuychaco:
Fixes #611
Fixes #751 

* ~~Adds a 'Discard All' button in Unstaged Changes list header (does this take up too much space? should we just use an icon? or remove and have users to use command or context menu)~~
* Adds command 'github:discard-all-changes' to the workspace
* Adds context-menu items for Unstaged/Staged Changes list headers
  * Unstaged Changes: 'Discard All Changes' and 'Stage All Changes'
  * Staged Changes: 'Unstage All Changes'
* Makes it so that context-menu items in list show appear only when an item is right-clicked (not just empty space)

### DIFF
--- a/menus/git.cson
+++ b/menus/git.cson
@@ -115,20 +115,20 @@
   ]
   '.github-UnstagedChanges .github-StagingView-header': [
     {
-      'label': 'Stage All'
-      'command': 'github:stage-all'
+      'label': 'Stage All Changes'
+      'command': 'github:stage-all-changes'
     }
     {
       'type': 'separator'
     }
     {
-      'label': 'Discard All'
-      'command': 'github:discard-all'
+      'label': 'Discard All Changes'
+      'command': 'github:discard-all-changes'
     }
   ]
   '.github-StagedChanges .github-StagingView-header': [
     {
-      'label': 'Unstage All'
-      'command': 'github:unstage-all'
+      'label': 'Unstage All Changes'
+      'command': 'github:unstage-all-changes'
     }
   ]
--- a/menus/git.cson
+++ b/menus/git.cson
@@ -113,3 +113,22 @@
       'command': 'github:open-link-in-browser'
     }
   ]
+  '.github-UnstagedChanges .github-StagingView-header': [
+    {
+      'label': 'Stage All'
+      'command': 'github:stage-all'
+    }
+    {
+      'type': 'separator'
+    }
+    {
+      'label': 'Discard All'
+      'command': 'github:discard-all'
+    }
+  ]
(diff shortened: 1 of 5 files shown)


simurai:
> should we just use an icon?

Yes, that sounds like a good idea. And also move it right of "Stage All". Then it's the same as in the diff view:

screen shot 2017-05-27 at 2 54 21 pm

I can style the icon to make it a bit stronger.

50Wliu:
Not a huge fan of the icon.  Looks dangerous and I'm not sure it fits in with Stage Selection not being an icon.  Since discarding changes is probably less likely to happen than staging all, maybe it could be moved into a context menu (like if you right-clicked Unstaged Changes).

Just my 💭s.

kuychaco:
Yeah I think I'm a fan of no button. It makes the feature less discoverable, but @50Wliu is right that it's not a super common action and it seems like it'd be fine if it was only tucked away in a context-menu. Plus there's the extra layer of friction for accidentally discarding everything when you mean to "Stage All" instead

simurai:
> Looks dangerous 

Ahh.. yeah, the :wastebasket: `trashcan` should fit better. But 👍  on context menu only.

What about the hunk view?

screen shot 2017-05-30 at 10 13 18 am

For now, I changed it to the trash can icon:

screen shot 2017-05-30 at 10 49 26 am

kuychaco:
I thought the 🔥 was pretty cute, but I do agree that 🗑 is more appropriate. Thanks for all the feedback!

menus/git.cson
```diff
@@ -38,7 +38,7 @@
       'command': 'github:open-file'
     }
   ]
-  '.github-UnstagedChanges .github-FilePatchListView': [
+  '.github-UnstagedChanges .github-FilePatchListView .github-FilePatchListView-item': [
```
smashwilson: Ah, yep. :+1:

PR #879, opened 2017-05-26
Open diff from editor

### LINKED ISSUE #46: Highlight changed lines in the editor and stage them
next-2-2

### LINKED ISSUE #888: Filter changed file lists
As requested in #845 to make working with lots of files easier

kuychaco:
Closes #888 
Closes #46

This PR adds commands for easily viewing the diff associated with the active editor: 
* `github:view-unstaged-changes-for-current-file`
* `github:view-staged-changes-for-current-file`

The RootController class 
1. grabs the active editor
1. determines the cursor location 
1. selects the associated item in the list view
1. opens the staged or unstaged file diff associated with the file
1. selects the closest `added` hunk line based on cursor position and scrolls it into view if necessary 

As part of this the FilePatchSelection API was extended to include a `goToDiffLine(lineNumber)` method which selects the closest `added` line based on the passed in line number. This can be used for a `github:go-to-diff-line` command similar to Atom's go-to-line functionality in editors. This is outside of the scope of the current PR, and may be implemented separately at a later point.

### DIFF
--- a/lib/controllers/git-tab-controller.js
+++ b/lib/controllers/git-tab-controller.js
@@ -273,13 +273,15 @@ export default class GitTabController {
   @autobind
   viewUnstagedChangesForCurrentFile() {
     const editor = this.props.workspace.getActiveTextEditor();
-    const lineNumber = editor.getCursorBufferPosition().row;
     const absFilePath = editor.getPath();
     const repoPath = this.getActiveRepository().getWorkingDirectoryPath();
     if (absFilePath.startsWith(repoPath)) {
       const filePath = absFilePath.slice(repoPath.length + 1);
       this.quietlySelectItem(filePath, 'unstaged');
-      return this.props.showFilePatchForPath(filePath, 'unstaged', {activate: true});
+      return this.props.showFilePatchForPath(filePath, 'unstaged', {
+        activate: true,
+        lineNumber: editor.getCursorBufferPosition().row + 1,
+      });
     } else {
       throw new Error(`${absFilePath} does not belong to repo ${repoPath}`);
     }
@@ -288,13 +290,16 @@ export default class GitTabController {
   @autobind
   viewStagedChangesForCurrentFile() {
     const editor = this.props.workspace.getActiveTextEditor();
-    const lineNumber = editor.getCursorBufferPosition().row;
     const absFilePath = editor.getPath();
     const repoPath = this.getActiveRepository().getWorkingDirectoryPath();
     if (absFilePath.startsWith(repoPath)) {
       const filePath = absFilePath.slice(repoPath.length + 1);
       this.quietlySelectItem(filePath, 'staged');
-      return this.props.showFilePatchForPath(filePath, 'staged', {activate: true, amending: this.props.isAmending});
+      return this.props.showFilePatchForPath(filePath, 'staged', {
+        activate: true,
+        lineNumber: editor.getCursorBufferPosition().row + 1,
+        amending: this.props.isAmending,
+      });
     } else {
       throw new Error(`${absFilePath} does not belong to repo ${repoPath}`);
     }
--- a/lib/controllers/git-tab-controller.js
+++ b/lib/controllers/git-tab-controller.js
@@ -3,7 +3,7 @@
 
 import path from 'path';
 import etch from 'etch';
-import {Disposable} from 'event-kit';
+import {CompositeDisposable, Disposable} from 'event-kit';
 
 import GitTabView from '../views/git-tab-view';
 import ModelObserver from '../models/model-observer';
@@ -33,7 +33,14 @@ export default class GitTabController {
     etch.initialize(this);
 
     this.element.addEventListener('focusin', this.rememberLastFocus);
-    this.subscriptions = new Disposable(() => this.element.removeEventListener('focusin', this.rememberLastFocus));
+    this.subscriptions = new CompositeDisposable();
+    this.subscriptions.add(
+      new Disposable(() => this.element.removeEventListener('focusin', this.rememberLastFocus)),
+      this.props.commandRegistry.add('atom-workspace', {
+        'github:view-unstaged-changes-for-current-file': this.viewUnstagedChangesForCurrentFile,
+        'github:view-staged-changes-for-current-file': this.viewStagedChangesForCurrentFile,
+      }),
+    );
   }
 
   serialize() {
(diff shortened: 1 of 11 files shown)


simurai:
Great addition. 😍 

I wonder if this replaces the "stage from editor" https://github.com/atom/github/issues/46? For very small changes, it's ok to "blindly" stage without seeing a diff. But otherwise it's nice be able to take a quick look what you're going to stage.

I think eventually, we should still offer some kind of "quick commit" option, but that's for another PR.

lib/controllers/git-tab-controller.js
```diff
@@ -364,7 +364,7 @@ export default class GitTabController {
 
   @autobind
   quietlySelectItem(filePath, stagingStatus) {
-    return this.refs.gitTab.quitelySelectItem(filePath, stagingStatus);
+    return this.refs.gitTab.quietlySelectItem(filePath, stagingStatus);
```
smashwilson: Oops 🙈

lib/controllers/root-controller.js
```diff
@@ -43,7 +43,7 @@ export default class RootController extends React.Component {
   static propTypes = {
     workspace: PropTypes.object.isRequired,
     commandRegistry: PropTypes.object.isRequired,
-    notificationManager: PropTypes.object.isRequired,
+    notificationManager: PropTypes.object.isReuired,
```
smashwilson: :point_up: It looks like this one was spelled right the first time? Caught a stray backspace keypress maybe?

lib/controllers/root-controller.js
```diff
@@ -574,6 +587,44 @@ export default class RootController extends React.Component {
...
+        pane.splitRight();
+      } else if (splitDirection === 'down') {
+        pane.splitDown();
+      }
+      return this.diveIntoFilePatchForPath(filePath, stagingStatus, {
+        lineNumber: editor.getCursorBufferPosition().row + 1,
+        amending: stagingStatus === 'staged' && this.state.isAmending,
+      });
+    } else {
+      throw new Error(`${absFilePath} does not belong to repo ${repoPath}`);
```
smashwilson: Thinking this through: because the active text editor sets the active repository context, the only time that this could happen is if it's triggered in the window between opening an editor on a file not in the active repository and the render context change completing. Performance captures indicate that window is only in the tens of milliseconds range, so it's unlikely that anyone could hit this manually.

_Eventually_ we might want to add logic to the effect of "wait on a Promise that resolves when the active context is what we expect it to be, then trigger `diveIntoFilePatchForPath`." Not worth holding up an initial merge for though I think.

lib/controllers/root-controller.js
```diff
@@ -574,6 +587,44 @@ export default class RootController extends React.Component {
...
+    }
+  }
+
+  @autobind
+  viewUnstagedChangesForCurrentFile() {
+    this.viewChangesForCurrentFile('unstaged');
```
smashwilson: I'd :heart: to turn `'unstaged'` and `'staged'` into Symbols or another more enum-ish, non-String type.

lib/views/file-patch-selection.js
```diff
@@ -324,4 +324,29 @@ export default class FilePatchSelection {
...
+        if (newDistance < closestLineDistance) {
+          closestLineDistance = newDistance;
+          closestLine = line;
+        } else {
+          return this.selectLine(closestLine);
+        }
+      }
+    }
+
+    return this.selectLine(closestLine);
```
smashwilson: :+1: This is pretty cool. :smile:

lib/controllers/root-controller.js
```diff
@@ -574,6 +587,44 @@ export default class RootController extends React.Component {
...
+
+  @autobind
+  viewStagedChangesForCurrentFile() {
+    this.viewChangesForCurrentFile('staged');
+  }
+
+  @autobind
+  goToDiffLine(lineNumber) {
+    return this.filePatchController.goToDiffLine(lineNumber);
+  }
```
smashwilson: Passing actions down the component hierarchy like this really feels like a "we should start looking at introducing Redux" smell to me 😆

lib/views/git-tab-view.js
```diff
@@ -182,7 +182,7 @@ export default class GitTabView {
   }
 
   @autobind
-  quitelySelectItem(filePath, stagingStatus) {
+  quietlySelectItem(filePath, stagingStatus) {
```
smashwilson: Haha oh wow. How many times have my eyes just glossed riiight over this one 😆

menus/git.cson
```diff
@@ -113,3 +113,13 @@
...
   ]
+  'atom-text-editor': [
+    {
+      'label': 'View Unstaged Changes',
+      'command': 'github:view-unstaged-changes-for-current-file'
+    }
+    {
+      'label': 'View Staged Changes',
+      'command': 'github:view-staged-changes-for-current-file'
+    }
```
smashwilson: Any opinions on keybindings to roll into #790 for these events?

lib/controllers/root-controller.js
kuychaco: Nice catch!

lib/controllers/root-controller.js
kuychaco: Yeah this felt icky to me. IIRC @BinaryMuse looked into Redux and dismissed using it for some reason. Can't remember what

lib/controllers/root-controller.js
kuychaco: Great suggestion. I'll take a look into refactoring in a separate PR :)

smashwilson:
Fixes #889 instead of #888 :grin:

lib/controllers/root-controller.js
BinaryMuse: Just adding my two cents — my gut reaction is to feel slightly uncomfortable with assuming that the active repository is the one we want to use for all operations. For example, I've wanted the ability to "pin" a repo so that it's always active for a while now. On a related note, I think that the diff views should render irrespective of the active repository; that is, they set and maintain their own context outside of the "active" one.

lib/controllers/root-controller.js
BinaryMuse: The way we manage model data doesn't suit itself well to Redux in my opinion. I also wanted to live with explicit props for a while to ensure we didn't switch to something a little more magical unless we really felt the pain and needed to — the abstraction, like any other, has a cost, and already the more meta-programmy parts of the package are the hardest to understand.

I *do* think we could use the other half of Redux — actions on the context — to good effect in the package. For example, anything that lives on the `AtomEnvironment` global could be placed on the context and components that need them could be wrapped in an HoC that accesses them (but allows them to be passed explicitly as props in tests). And other methods that perform actions that do not directly modify state and that do not affect rendering could be passed on context, as long as we're careful and selective about it.

Issue #880, opened 2017-05-27
Show the full commit message as tooltip


ahmed-taj:
### Description

I need to see the full message of a commit as a tooltip. Currently, the package only shows the message headline as a tooltip, which isn't very useful (at least for me)

### Steps to Reproduce

1. Open Atom
2. Open a PR (e.g [this one](https://github.com/atom/github/pull/876))
3. Scroll down to a commit that has message body (e.g [this](https://github.com/atom/github/pull/876/commits/4d6bc71eed5ce92a39aa00d53bdc24dd71b0638a) or [this](https://github.com/atom/github/pull/876/commits/42d8c67dabbc2c74f8c3424c8afdc5eb1c4ed64d))
4. Hover on that commit

**Expected behavior:**

A tooltip with full commit message (not just headline)

**Actual behavior:**

A tooltip contains only the commit message headline

**Reproduces how often:**

100%

### Versions

```
# atom --version
Atom    : 1.18.0-beta0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0

# apm --version 
apm  1.18.1
npm  3.10.10
node 6.9.5 x64
python 2.7.12
git 2.7.4
```

### Additional Information
![selection_007](https://cloud.githubusercontent.com/assets/12673605/26520456/9541324a-42db-11e7-8e09-42a2ab8e3c78.png)

### Proposed solution

We can [retrieve](https://github.com/atom/github/blob/master/lib/containers/timeline-items/commit-container.js#L42) `message` attribute instead `messageHeadline` then [set](https://github.com/atom/github/blob/master/lib/containers/timeline-items/commit-container.js#L23) it as tooltip (AFAIK, `messageHeadline` has no other usage)

I can open a PR for it in case this solution makes sense

PR #883, opened 2017-05-28
Set the full commit message as tooltip

ahmed-taj:
### Description of the Change

Removed `messageHeadline` attribute from `Commit` container, and replaced it with `message` which contains the full text of the commit message.

### Applicable Issues

Resolves: #880

### Additional info

screenshot


kuychaco:
Awesome, thanks for the contribution ✨

Issue #891, opened 2017-05-30
Opening github from the toolbar doesn't open after closing pane


ungb:
### Prerequisites

* [x] Have you tried to reproduce the problem in [Safe Mode][safe-mode]?
* [x] Have you followed all applicable steps in the [debugging guide][debugging-guide]?
* [x] Have you checked the FAQs on the [message board][message-board] for common solutions?
* [x] Have you checked that your issue [isn't already filed][already-filed]?

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

### Description

Opening Github or Git Pane after closing the pane adds the pane back to the dock, but doesn't open the dock until you click again.

### Steps to Reproduce

1.  Open Atom.
2. If github package is open, close it by right clicking and clicking `close pane`
3. On the bottom right click on the `[0-9]+ files`

**Expected behavior:**
[What did you expect to happen?]
Opens the git panel

**Actual behavior:**
[What actually happened instead?]
Nothing happens or it seems that way, but actually the git panel gets added to the dock on the right but it's closed.

 
**Reproduces how often:**
[What percentage of the time does this happen?]
100%

### Versions

You can get this information from copy and pasting the output of `atom --version` and `apm --version` from the command line. Also, please include the OS and what version of the OS you're running.
```
Atom    : 1.19.0-dev-58313f6ab
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
```
### Additional Information
gif:

![pane](https://cloud.githubusercontent.com/assets/5624870/26608233/5c973f5e-454f-11e7-87c2-1fe173538433.gif)

Please include any additional information, configuration or data that might be necessary to reproduce the issue. Screenshots, if appropriate, are helpful as well.

If your issue involves a repository, and that repository is public, please include a link to it.

I noticed this while testing fix for https://github.com/atom/github/issues/872
/cc @smashwilson

smashwilson:
This is weird: I can reproduce, but it shows fine on the _second_ click?

smashwilson:
Okay, I think I got it :smile: It's published in 0.3.1 and included in Atom master and beta. Care to give it another spin?

ungb:
[...]

PR #895, opened 2017-05-31
Keep state and dock toggling in sync

smashwilson:
When `gitTabActive` was "false" and the toggle method was invoked, the state was being toggle to "true" correctly, but the DockItem was not being toggled correctly in the workspace. As a result, it took _two_ invocations of `github:toggle-git-tab` to show the Git tab if the dock it was in was not visible.

Probably addresses #891.

- [x] Clarify state transitions in `TabTracker.toggle()` with and without legacy panels active. With legacy panels, the `gitTabActive` component state dictates the visibility of the panel completely, so toggling should only flip the state; with the dock API, `gitTabActive` controls whether or not the tab is _rendered_, not visible, so `.toggle()` may need to reveal the dock.
- [x] Bring the RootController tests up to date. There are a few still failing because `isGitPaneDisplayed` doesn't account for the docks API.
- [x] For new projects when the docks API is available, default the Git and GitHub tabs to _rendered_ but not _activated_. Meaning, if you open Atom on a new project, the right dock is hidden, but if you open it yourself the Git and GitHub tabs are present by default.

### DIFF
--- a/lib/controllers/root-controller.js
+++ b/lib/controllers/root-controller.js
@@ -143,6 +143,9 @@ export default class RootController extends React.Component {
   componentWillReceiveProps(newProps) {
     this.repositoryObserver.setActiveModel(newProps.repository);
     this.repositoryStateRegistry.setModel(newProps.repository);
+
+    this.gitTabTracker.useLegacyPanels = newProps.useLegacyPanels;
+    this.githubTabTracker.useLegacyPanels = newProps.useLegacyPanels;
   }
 
   render() {
--- a/lib/controllers/root-controller.js
+++ b/lib/controllers/root-controller.js
@@ -884,6 +884,16 @@ class TabTracker {
     return false;
   }
 
+  reveal() {
+    const dockItem = this.getDockItem();
+    return dockItem ? dockItem.reveal() : Promise.resolve(null);
+  }
+
+  hide() {
+    const dockItem = this.getDockItem();
+    return dockItem ? dockItem.hide() : Promise.resolve(null);
+  }
+
   focus() {
     this.getControllerComponent().restoreFocus();
   }
--- a/lib/controllers/root-controller.js
+++ b/lib/controllers/root-controller.js
@@ -820,13 +820,10 @@ class TabTracker {
     this.getState = getState;
     this.getWorkspace = getWorkspace;
     this.getController = getController;
+    this.getDockItem = getDockItem;
 
     this.setStateKey = value => {
       return new Promise(resolve => setState(value, resolve));
-    };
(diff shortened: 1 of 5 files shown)


smashwilson:
> For new projects when the docks API is available, default the Git and GitHub tabs to rendered but not activated. Meaning, if you open Atom on a new project, the right dock is hidden, but if you open it yourself the Git and GitHub tabs are present by default.

:point_up: This is the bit that I'm least certain about from a "what do we actually want to happen" perspective. @BinaryMuse, @kuychaco, @simurai, any thoughts on this?

simurai:
> For new projects when the docks API is available, default the Git and GitHub tabs to rendered but not activated. Meaning, if you open Atom on a new project, the right dock is hidden, but if you open it yourself the Git and GitHub tabs are present by default.

Sounds good to me. 👌 Since there is nothing to stage/commit when opening a new project, I think it's preferable to have it closed. I guess it depends how wide your window is, but I'm a fan of starting out minimal because it's hard to know what a user intends to do next.

smashwilson:
Here's the matrix of cases that I'm testing:

- atom.config('github.useLegacyPanels', true)
  - atom.config('welcome.showOnStartup', true)
    - `~/.atom/github.cson` does not exist
      - [ ] Panel is not visible
    - `~/.atom/github.cson` exists
      - [ ] Panel is not visible
  - atom.config('welcome.showOnStartup', false)
     - `~/.atom/github.cson` does not exist
       - [ ] Panel is visible
     - `~/.atom/github.cson` exists
       - [ ] Panel is not visible
- atom.config('github.useLegacyPanels', false)
  - atom.config('welcome.showOnStartup', true)
    - `~/.atom/github.cson` does not exist
      - [ ] Dock is not visible
      - [ ] Panes are present within dock when expanded manually
    - `~/.atom/github.cson` exists
      - [ ] Dock is not visible
      - [ ] Panes are present within dock when expanded manually
  - atom.config('welcome.showOnStartup', false)
     - `~/.atom/github.cson` does not exist
       - [ ] Dock is visible with panes
     - `~/.atom/github.cson` exists
       - [ ] Dock is not visible
       - [ ] Panes are present within dock when expanded manually

smashwilson:
@ungb, this should fix the "need to click twice on the 'open git' button" issue that you found testing out #875. Once this is ✅ I'll merge, publish, update in atom/atom and cherry-pick it on to 1.18-releases and we can re-verify 😅 . Hopefully the test-case enumeration up above (and a lot of `atom-beta --clear-window-state`) might be helpful for making sure that everything is behaving as we expect it to!

PR #896, opened 2017-05-31
Edit commit message in editor

### LINKED ISSUE #818: Larger commit message editor
### Description

The commit message editor is currently (vertically) very small - and not resizable. It would be a much better experience if some or all of these were true:

* The commit message editor was vertically resizable
* You could pop open the commit message in a new pane/tab/something

### Versions

```console
$ apm-beta --version
apm  1.18.1
npm  3.10.10
node 6.9.5 x64
python 2.7.12
git 2.13.0

$ atom-beta --version
Atom    : 1.18.0-beta0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0

$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="16.04.2 LTS (Xenial Xerus)"
```

kuychaco:
Fixes #818 

TODO:
- [x] add commit grammar
- [x] activate commit editor if already open
- [x] add tests
- [x] obscure commit box when editor is open
  * per @BinaryMuse suggestion:
> If the user has the file expanded, I'd love to gray out the box or display a little overlay with an informational message. If the user clicks it (or the arrow or something) again, it'd find and focus that buffer in the workspace center. Then if the editor is closed, we re-read the text and turn it back on.
- [x] add command 
- [x] add context menu
- [x] add button somewhere
- [x] check if file is unsaved
- [x] handle case where commit box is empty and `commit` button is disabled
- [x] handle splitting commit message file into multiple panes and closing the original
- [x] delete file after committing
- [x] toggle the expanded editor instead of "revealing" it
- [x] prompt to save before destroying editor


kuychaco:
Here's what I've got so far:

commit-editor2

Could use some UX help around where to put a button to open the editor, and what to do to obscure the commit box when the editor is open. Right now I just hide the commit box using CSS. @simurai @BinaryMuse mind if I ask for your help implementing these?

simurai:
Added a subtle "expand" button:

expandbutton

> If the user clicks it (or the arrow or something) again, it'd find and focus that buffer in the workspace center. Then if the editor is closed, we re-read the text and turn it back on.

Rather than "find the expanded commit message editor", should it close it instead? Than it would be more like a toggle:

image

simurai:
@kuychaco Ok, added the expanded styles. Still needs wiring up:

1. On `github-CommitView-editor` replace `hidden` class with `is-expanded`
2. Make the expand button clickable
3. **Optional**: Move the text into the `<button>` element. Currently it's just a [pseudo element](https://github.com/atom/github/blob/bf69ee75babb75b253ae844b1a575ab173d26e5b/styles/commit-view.less#L97-L99).

image

simurai:
> Wire up button

Nice!

Do you like the behaviour? I still think it should toggle the expanded editor instead of "revealing" it.

kuychaco:
> I still think it should toggle the expanded editor instead of "revealing" it.

Agreed 👍

test/controllers/commit-view-controller.test.js
```diff
@@ -128,5 +147,126 @@ describe('CommitViewController', function() {
...
+
+        editor.setText('message in editor');
+        editor.save();
+
+        commandRegistry.dispatch(atomEnvironment.views.getView(editor), 'pane:split-right-and-copy-active-item');
+        await assert.async.notEqual(workspace.getActiveTextEditor(), editor);
+
+        sinon.spy(controller.refs.commitView, 'update');
+        editor.destroy();
+        await assert.async.isTrue(controller.refs.commitView.update.called);
```
BinaryMuse: 🤔 I wonder if this line's intention would be more clear as `await until(() => controller.refs.commitView.update.called)`

test/views/commit-view.test.js
```diff
@@ -41,45 +41,53 @@ describe('CommitView', function() {
     });
   });
 
-  it('displays the remaining characters limit based on which line is being edited', async function() {
+  it.only('displays the remaining characters limit based on which line is being edited', async function() {
```
BinaryMuse: Not sure if this is just here for debugging but :fire:

smashwilson:
:tada:

atomi:
Can we get a shortcut to git push after committing the git message. Something like ctrl+shift+p?

BinaryMuse:
@atomi There's already a keybind for the Push command; you can see it by looking next to the "GitHub: Push" command in the command palette, and you can always bind your own keybind if you don't like the included one. See http://flight-manual.atom.io/behind-atom/sections/keymaps-in-depth/ for more.

atomi:
@BinaryMuse Thank you for the fast reply! I was thinking more along the lines of a quick push while within the commit input similar to `git gui` a commit would be ctrl+enter and a push would be something like ctrl+p all within the git commit input.

PR #905, opened 2017-06-05
Display status bar info for pending pane items

jasonrudolph:
As one part of resolving https://github.com/atom/status-bar/issues/194, this PR updates this package to expect any git-backed pane items (e.g., text editors, image files, etc.) to exist in the workspace center.

### Alternate Designs

See https://github.com/atom/atom/pull/14695

### Benefits

The package will be able to display the git branch and push/pull status for pending pane items as the user navigates the tree-view.

### Possible Drawbacks

See https://github.com/atom/atom/pull/14695

### Applicable Issues

- https://github.com/atom/atom/issues/14648
- https://github.com/atom/status-bar/issues/194

#### Before

In the demo below, we start with an _active_ text editor for the `.gitignore`, and then we open another text editor for the `.gitmodules` file (in a different project) in a pending pane. When opening the pending text editor, the status bar shows the git branch and push/pull status for the previous editor. We have to take the additional step of clicking on the pane item in order to get the status bar to reflect the file's correct git branch and push/pull status.

github-before

#### After

In the demo below, we follow the same steps as above. This time, when opening the pending text editor, the status bar immediately shows the correct git branch and push/pull status for the pending pane item.

github-after

### DIFF
--- a/lib/github-package.js
+++ b/lib/github-package.js
@@ -141,7 +141,7 @@ export default class GithubPackage {
         this.rerender();
       }),
       this.project.onDidChangePaths(this.scheduleActiveContextUpdate),
-      this.workspace.onDidChangeActivePaneItem(this.scheduleActiveContextUpdate),
+      this.workspace.getCenter().onDidChangeActivePaneItem(this.scheduleActiveContextUpdate),
       this.styleCalculator.startWatching(
         'github-package-styles',
         ['editor.fontSize', 'editor.fontFamily', 'editor.lineHeight'],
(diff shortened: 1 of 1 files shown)


lib/github-package.js
```diff
@@ -360,10 +359,9 @@ export default class GithubPackage {
     };
 
     const items = [
-      this.workspace.getActivePaneItem(),
       this.workspace.getCenter && this.workspace.getCenter().getActivePaneItem(),
     ];
-    const [active, center = {}] = await Promise.all(items.map(fromPaneItem));
+    const [active = {}] = await Promise.all(items.map(fromPaneItem));
```
smashwilson: :point_up: This can be simplified now that `items` only contains a single element:

```js
const active = await fromPaneItem(this.workspace.getCenter().getActivePaneItem());
```

jasonrudolph: Thanks, @smashwilson. Done in 9596e4456660767a3f9f9b8cd5583948537142ce. ✨

PR #909, opened 2017-06-06
[WIP] Add Clone button

### LINKED ISSUE #907: "Clone" button on inactive repository view
💡 from @BinaryMuse: Let's add a "clone" button to the git panel when there's no active repository.

/cc @simurai for design 🌈

simurai:
### Description of the Change

This adds a "Clone" button to the `no-repository` view:

image

It also changes the "Create repository" to "Initialize". Then the two buttons only have a single word for fast recognizing and it also mimics `git init/clone` commands.

### TODO

- [x] Style button
- [ ] Wire up button

### Benefits

Improves discoverability of the "Clone" feature.

### Possible Drawbacks

None, except maybe a bit more cluttered UI

### Applicable Issues

Closes #907


bddicken:
@simurai Is there a reason why this has not been merged? I think a button like this would be great. It is not initially obvious that atom even supports this feature without such a button.

I just figured out that one can cmd-shift-p and the search for "github clone" to do this, but that isn't an obvious set of steps, especially for a beginner.

simurai:
> Is there a reason why this has not been merged?

This PR isn't functional yet and only adds the button. Still needs wiring up and run the command. I'll add a "help wanted" label.

bddicken:
Cool, thanks. This would be great to have as a button. The current way of cloning from github (cmd-shift-p, search "github clone", select match) is not idea.

simurai:
Closing since this got a bit out of date and has conflicts.

PR #846, opened 2017-05-18
Custom pinentry wrapper for GPG

smashwilson:
Providing passwords to gpg non-interactively via `--passphrase-fd` is fragile and somewhat counter to gpg's architecture. As I understand it, gpg2 [expects secret key material to only be handled by the `gpg-agent` process](https://lists.gt.net/gnupg/devel/69635):

> The reason why you need to enable the loopback mode is that this breaks a design goal of only allowing the gpg-agent (+scdaemon) to handle the private keys and the passphrases for it.

This _only_ works if the agent was configured with `--allow-loopback-pinentry` when it was started and, in my version of gpg at least, if `--pinentry-mode loopback` is provided on the gpg command line, which has the side-effect of preventing user-configured pinentry programs from being attempted at all. At least it looks like [aren't the only ones to run into this](https://github.com/keybase/keybase-issues/issues/1712) :wink:

This pull request replaces the `--passphrase-fd` approach with a `gpg-wrapper` helper that:

1. Attempts to run the gpg application as-is. If the user has a [custom pinentry application](https://github.com/GPGTools/pinentry-mac) that works without a tty, the commit will succeed.
2. Otherwise, creates a temporary configuration directory within the one we already create for GitPromptServer and copies any existing gpg configuration and data into it.
3. Launches a new, independent gpg-agent daemon using the temporary directory as a `--homedir` and specifying `gpg-pinentry.sh` as a pinentry program.
4. Re-attempts the gpg operation with that `--homedir` so that it will connect to our isolated agent.
5. `gpg-pinentry.sh` adapts the [pinentry Assuan protocol](http://info2html.sourceforge.net/cgi-bin/info2html-demo/info2html?(pinentry)Protocol) to `git-askpass-atom.sh`, which connects to the GitPromptServer socket and requests a password just like it does for SSH connections.

### Compatibility

The trickiest part here is ensuring that we have reasonably compatibility across both gpg versions and platforms. gpg uses a "home directory" to isolate its keychain, configuration, and so on, and it _really_ doesn't like it if you share a home directory among different gpg versions.

To help with this a bit, I've created a [test matrix repository](https://github.com/smashwilson/gpg-matrix) that automates the testing of our driver scripts across gpg versions on macOS and Linux with isolated build prefixes and home directories. Calling gpg and calling our scripts are a bit tricky so that's a work in progress as well. To keep the combinatorial explosion in check somewhat I've limited it to the latest patch release of each minor version.

[...]

PR #925, opened 2017-06-14
Respect GIT_AUTHOR_*/GIT_COMMITTER_* environment variables

BinaryMuse:
`GIT_AUTHOR_*` and `GIT_COMMITTER_*` environment variables weren't being sent to the Git worker process, so pairing information (such as set by git-pear) didn't work.

/cc @smashwilson 
/cc @jasonrudolph for info

### DIFF
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -20,6 +20,11 @@ import WorkerManager from './worker-manager';
 const LINE_ENDING_REGEX = /\r?\n/;
 
 const GPG_HELPER_PATH = path.resolve(getPackageRoot(), 'bin', 'gpg-no-tty.sh');
+const ENV_VARS_TO_COPY = [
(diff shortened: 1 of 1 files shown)


lib/git-shell-out-strategy.js
```diff
@@ -80,6 +85,12 @@ export default class GitShellOutStrategy {
         PATH: process.env.PATH || '',
       };
 
+      ENV_VARS_TO_COPY.forEach(envVar => {
+        if (process.env[envVar] !== undefined) {
+          env[envVar] = process.env[envVar];
+        }
+      });
+
```
smashwilson: I feel like I've done this a few times now. It might be worth refactoring into a helper somewhere 🤔

smashwilson: Oh I think the other times are in #846 which isn't :shipit: yet. And which would be run from a process that can't import helpers easily.

Issue #854, opened 2017-05-19
ux: Using keyboard to stage a file should show the next file's diff


benogle:
* Have multiple files with changes
* Select a file in the unstaged changes area, note that it shows the diff
* Hit enter to stage it
* Note that the file is staged, and another file in the unstaged changes list is selected, but the newly selected file's diff is not shown. If I hit the up/down key or click on the file, the diff is shown.

simurai:
❤️  Welcome back to the team.

BinaryMuse:
This has come up a couple times in various contexts — https://github.com/atom/github/issues/606,  https://github.com/atom/github/issues/645 — and the latest conversation ended in a decision to keep the diff view open, but empty, with an easy way to get to the next file. Thoughts?

benogle:
> ❤️ Welcome back to the team.

❤️  lol, my feels were flaring up when using the git stuff. It's _so nice_ to not need to leave the editor.

Keeping the diff open sounds good. Seems like keeping it open whenever the diff view or git tab are focused would be legit. I guess when to hide it is kind of hard. My gut is that I want to see the diff view the whole time I deal with git stuff, then when I ctrl-shift-9 hide, I want to see no more git stuff.

Anyway feel free to close this as a dupe if you wanna. And nice job on the package all y'all.

kuychaco:
👋 @benogle, thanks for the feedback! Captured it in #864.

Issue #645, opened 2017-04-05
minor usability nitpick


technoweenie:
Hey all, I ran into something mildly annoying in the commit view of the github plugin. When I stage the hunks for a file, the "Unstaged Changes" tab turns into a regular file tab. If I'm crushing some diff hunks to make a commit, I'd like it to display the unstaged changes for the next file in the list.

![release_asset_rb_ ___github_github](https://cloud.githubusercontent.com/assets/21/24712883/573ca786-19e1-11e7-8600-b1040d14379e.png)

### Versions

Atom Version (`atom --version`):

```
Atom    : 1.15.0
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
```

GitHub Package Version (`git --git-dir ~/.atom/packages/github/.git rev-parse head`):

dce9960042816f969d893f2841263ee61530d4d8

BinaryMuse:
Since the solution here is the same as in #606 I'm gonna close this and track it there ^_^ Thanks!

PR #970, opened 2017-06-22
After staging file via core:confirm, show the next diff

jasonrudolph:
### Description of the Change

As requested in https://github.com/atom/github/issues/854, when you have multiple files present in the staging view, hitting enter stages the file and shows the diff for the next file in the list. See before-and-after demo below.

### Alternate Designs

I didn't consider any other designs at this point. For now, I'm opening this PR to get some initial feedback on whether other people appreciate this UX change as much as I do. :innocent:

### Benefits

Streamlined workflow for people using keyboard to stage/unstage multiple files in succession.

### Possible Drawbacks

I'm not aware of any.

### Applicable Issues

Currently: https://github.com/atom/github/issues/864

Previously: https://github.com/atom/github/issues/854, https://github.com/atom/github/issues/645

#### Before

before

#### After

after

### DIFF
--- a/lib/views/staging-view.js
+++ b/lib/views/staging-view.js
@@ -218,7 +218,10 @@ export default class StagingView {
 
   confirmSelectedItems() {
     const itemPaths = Array.from(this.selection.getSelectedItems()).map(item => item.filePath);
-    return this.props.attemptFileStageOperation(itemPaths, this.selection.getActiveListKey());
+    this.props.attemptFileStageOperation(itemPaths, this.selection.getActiveListKey());
+    this.selection.coalesce();
+    this.debouncedDidChangeSelectedItem();
+    return etch.update(this);
   }
 
   getNextListUpdatePromise() {


simurai:
👍  I'm in favor of having this the default behaviour.

The only thing that feels a bit jarring: After staging a file, the diff view closes and a new diff view opens for the next file. Would it be possible to "re-use" the same diff view pane? So that there is no "flicker".

kuychaco:
@jasonrudolph thanks so much for this PR. I'm totally onboard with this UX change 👍

I've picked up from where you've left off, to handle other cases (such as staging/unstaging via mouse click) and making it such that we don't close the diff before displaying the next one (as @simurai mentioned). I also made it such that if the user closes the diff view, then we don't open a new diff view when a new file is selected (see below).

What I have so far is still super WIP. I'll pick it up and push it forward next week. 

unwanted-diff-opening

Appreciate your input!

jasonrudolph:
> I've picked up from where you've left off, to handle other cases (such as staging/unstaging via mouse click) and making it such that we don't close the diff before displaying the next one (as @simurai mentioned).

@kuychaco: Thank you! :bow::zap:

jasonrudolph:
This PR is superceded by https://github.com/atom/github/pull/1016, so I'm closing this out.

@kuychaco: Thanks for all your work in #1016. 🌟

PR #999, opened 2017-07-02
⬆ dependencies

BinaryMuse:
This updates a bunch of dependencies, including Relay, and additionally adds a `package-lock.json`.


BinaryMuse:
So the new Babel Relay plugin is giving us some issues because it loads the schema relative to `process.cwd()` — need to find a workaround.

Ref https://github.com/facebook/relay/issues/1755

BinaryMuse:
Opened a PR for a fix at https://github.com/facebook/relay/pull/1933

BinaryMuse:
The Travis builds have both passed, although one of the builds doesn't seem to want to turn green.

PR #1031, opened 2017-07-13
Fix `Repository#getGitDirectoryPath()` for absent or loading repos

BinaryMuse:
This is the cause of the failing main process tests in https://github.com/atom/atom/pull/14959


BinaryMuse:
macOS builds are backed up; merging with other tests passing.

PR #1034, opened 2017-07-14
Always use URIs to manage Git/GitHub dock items

BinaryMuse:
Opening in a separate PR to facilitate review.

/cc @kuychaco 

- [x] `DockItem` uses `StubItem`s directly
- [x] Opening the Git/GitHub panel uses `Workspace#open`
- [x] Git and GitHub tab controllers deserialize with correct URIs
- [x] `TabTracker` is updated as necessary
- [ ] Git/GitHub tabs can *either* be split *or* this is explicitly disallowed


BinaryMuse:
Soo~oo I would love to take this opportunity to allow there to be *multiple* Git/GitHub tabs, each of which could be *pinned* to a specific project repository, but that's a bit larger undertaking because of focus management and because of UX work. So I'm going to punt on it and attempt to make them un-split/copyable for now.

Issue #1041, opened 2017-07-18
The commit message box and commit button should be disabled during a commit operation


Xapphire13:
### Description
Commit message and commit button remain enabled during the execution of a commit. This is likely not noticeable usually, but my project has a long running pre-commit hook so I have a large window to observe.

### Steps to Reproduce

1. Modify a file
2. Stage it
3. Type a commit message
4. Hit commit

**Expected behavior:**
The "commit message" box and the commit button should become disabled until the operation completes.

**Actual behavior:**
Both the message box and commit button are enabled. I can click commit as many times as I wish and all but the first lead to an error message.

### Versions
**OS**: Windows 10 x64
**Atom**: 1.18.0 x64
**atom/github**: 0.3.3

### Additional Information

Successful commit, UI elements remaining enabled:
![commit](https://user-images.githubusercontent.com/3392349/28295397-331104a8-6b15-11e7-8df9-6852ecf709ad.gif)

Clicking commit multiple times (ignore the output in the error, that's the preamble in my pre-commit hook):
![commit fail](https://user-images.githubusercontent.com/3392349/28295434-67ef7a56-6b15-11e7-8119-eee789486e54.gif)

BinaryMuse:
I'm going to call this closed as of #1043 even though there's not a great way to disable the actual message editor box. We can tackle additional UI improvements there in the future.

PR #1043, opened 2017-07-18
Blocking interaction with the commit button during a commit operation

Xapphire13:
### Description of the Change
This PR addresses part of #1041 and prevents users from hitting the bug of being able to issue a commit operation multiple times for the same changes.

### Benefits
Inability to hit commit multiple times during the commit operation.

### Additional info
Giving it a quick look over, I couldn't find any API for disabling the message box (which uses the [TextEditor](https://atom.io/docs/api/v1.18.0/TextEditor) class from Atom). Seems [others](https://github.com/atom/atom/issues/9237) have tried to find this also.


lib/views/commit-view.js
```diff
@@ -124,7 +124,13 @@ export default class CommitView {
   @autobind
   async commit() {
     if (await this.props.prepareToCommit() && this.isCommitButtonEnabled()) {
-      await this.props.commit(this.editor.getText());
+      await this.update({commitInProgress: true});
```
BinaryMuse: In general, we'll want to avoid calling `update` with custom objects, since `update` is how props propagation works in Etch/React. In this case, I think it would be appropriate to add an instance variable to the *parent* component, `CommitViewController`, since that's the component that actually controls the committing (via the function it passes down to this component in the `commit` prop), and then pass its value down to this component via a property (and re-render with `etch.update(this)` in the parent as appropriate).

Xapphire13: Sounds good to me =]

BinaryMuse:
Thanks so much!

Xapphire13:
No worries, thank you both!

PR #1058, opened 2017-07-27
Copy the CircleCI template, but pull from my branch

smashwilson:
Let's see if this works... ! 🤞


smashwilson:
(Spoiler: it didn't)

PR #1075, opened 2017-08-02
Pass all environment vars to spawned git process

### LINKED ISSUE #898: bash_profile not executed when committing
### Prerequisites

* Have you tried to reproduce the problem in [Safe Mode][safe-mode]? yes
* Have you followed all applicable steps in the [debugging guide][debugging-guide]? yes
* Have you checked the FAQs on the [message board][message-board] for common solutions? yes
* Have you checked that your issue [isn't already filed][already-filed]? yes

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

### Description

When attempting to commit I'm receiving an error from our pre-commit linter due to an environment variable not being set, this environment variable is set via my bash_profile. Committing via terminal or my other GUI (Tower) works fine so it appears to be an issue isolated to github for atom.

### Steps to Reproduce

1. Set an environment variable within your bash_profile
2. Depend upon this variable within a pre-commit script and fail if it's not set
3. Attempt to commit a change via github for atom

**Expected behavior:**

Pre-commit step executes as expected using the environment variable from bash_profile.

**Actual behavior:**

An error is displayed stating that the environment variable is missing.

**Reproduces how often:**

100%

### Versions

Atom    : 1.17.2
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0

apm  1.17.0
npm  3.10.5
node 6.9.5 x64
python 2.7.10
git 2.12.2

macOS Sierra 10.12.5

### LINKED ISSUE #1051: Repositories using overcommit fail to commit
### Description

Using the `overcommit` tool for managing git hooks does not allow Git to commit through the github package. When committing, this error message appears:

> This repository contains hooks installed by Overcommit, but the `overcommit` gem is not installed.
> Install it with `gem install overcommit`.

While there is [an issue with overcommit working with editors and RVM](https://github.com/brigade/overcommit/issues/402), after extensive testing, I believe that this is actually an Atom/github package issue. I am using RVM, but even installing overcommit with the system gems and confirming that it works without special `PATH` settings, committing still fails through Atom. 

I also managed to load the PATH variable using [atom-project-shell-env](https://github.com/ddiachkov/atom-project-shell-env) but the Git commit still fails when done through the package.

### Steps to Reproduce

1. Create a sample repository
2. `gem install overcommit`
3. In sample repository, `overcommit --install`
4. Commit the `.overcommit.yml` file before proceeding (important)
5. Make changes to the sample repository
6. Attempt to commit sample changes through atom-github
7. See error message

**Expected behavior:**

It should commit without an issue, unless a git hook fails.

**Actual behavior:**

This error message:

image

> This repository contains hooks installed by Overcommit, but the `overcommit` gem is not installed.
> Install it with `gem install overcommit`.

**Reproduces how often:**

100%

### Versions

Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0

apm  1.18.1
npm  3.10.10
node 6.9.5 x64
python 2.7.13
git 2.13.1

MacOS 10.12.5 (16F73)

kuychaco:
Fixes #1051
Fixes #898 

This assumes that the main Atom process inherits the user's environment variables, which is the intended behavior.

### DIFF
--- a/test/git-prompt-server.test.js
+++ b/test/git-prompt-server.test.js
@@ -80,7 +80,7 @@ describe('GitPromptServer', function() {
       assert.isFalse(await fileExists(path.join(server.tmpFolderPath, 'remember')));
     });
 
-    it('preserves a provided username', async function() {
+    it.only('preserves a provided username', async function() {
       this.timeout(10000);
 
       let queried = null;
--- a/test/git-prompt-server.test.js
+++ b/test/git-prompt-server.test.js
@@ -80,8 +80,9 @@ describe('GitPromptServer', function() {
       assert.isFalse(await fileExists(path.join(server.tmpFolderPath, 'remember')));
     });
 
-    it.only('preserves a provided username', async function() {
+    it('preserves a provided username', async function() {
       this.timeout(10000);
+      this.retries(5);
 
       let queried = null;
 
@@ -115,6 +116,7 @@ describe('GitPromptServer', function() {
 
     it('parses input without the terminating blank line', async function() {
       this.timeout(10000);
+      this.retries(5);
 
       function queryHandler(query) {
         return {
--- a/test/git-prompt-server.test.js
+++ b/test/git-prompt-server.test.js
@@ -169,8 +169,9 @@ describe('GitPromptServer', function() {
   });
 
   describe('askpass helper', function() {
-    it('prompts for user input and writes the response to stdout', async function() {
+    it.only('prompts for user input and writes the response to stdout', async function() {
       this.timeout(10000);
+      this.retries(5);
 
       let queried = null;
 
--- a/test/git-strategies.test.js
+++ b/test/git-strategies.test.js
@@ -1035,6 +1035,7 @@ import {fsStat, normalizeGitHelperPath, writeFile, getTempDir} from '../lib/help
       });
 
       it('falls back to Atom credential prompts if credential helpers are present but explode', async function() {
+        this.retries(5);
         let query = null;
         const git = await withHttpRemote({
           prompt: q => {
--- a/test/git-strategies.test.js
+++ b/test/git-strategies.test.js
@@ -28,6 +28,36 @@ import {fsStat, normalizeGitHelperPath, writeFile, getTempDir} from '../lib/help
   };
 
   describe(`Git commands for CompositeGitStrategy made of [${strategies.map(s => s.name).join(', ')}]`, function() {
+    // https://github.com/atom/github/issues/1051
+    // https://github.com/atom/github/issues/898
+    it('passes all environment variables to spawned git process', async function() {
+      const workingDirPath = await cloneRepository('three-files');
+      const git = createTestStrategy(workingDirPath);
+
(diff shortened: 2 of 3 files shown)


BinaryMuse:
@smashwilson I'm curious if you can think of any environment variables we *shouldn't* copy over. I think all the ones we care about we override later if necessary?

smashwilson:
> @smashwilson I'm curious if you can think of any environment variables we shouldn't copy over. I think all the ones we care about we override later if necessary?

Yeah, `GIT_CONFIG_PARAMETERS` is the big one to prevent infinite recursion of credential handlers and such (it's how git passes flags like `-c credential.helper=...` to its own subprocess trees) and we handle that one internally.

There _may_ be other environment variables [that modify git behavior](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables) in ways that we don't expect that would break our integration. At a glance, `GIT_EXTERNAL_DIFF` looks like the only one that would really conflict (we need to override that config setting too), because it'll interfere with our ability to parse the diff output. `GIT_FLUSH` might also cause performance issues I suppose, but I'd assume that if someone set that it's because they need it for some reason (on a weird-ass filesystem maybe?)

test/git-prompt-server.test.js
```diff
@@ -82,6 +82,7 @@ describe('GitPromptServer', function() {
 
     it('preserves a provided username', async function() {
       this.timeout(10000);
+      this.retries(5);
```
smashwilson: :+1: Yesss. Let the deflaking begin

smashwilson: Also: I was hoping that #1068 would make these fail more informatively at least 🤔

test/git-strategies.test.js
```diff
@@ -28,6 +28,36 @@ import {fsStat, normalizeGitHelperPath, writeFile, getTempDir} from '../lib/help
...
+
+      const hookPath = path.join(workingDirPath, '.git', 'hooks', 'pre-commit');
+      await writeFile(hookPath, hookContent);
+      fs.chmodSync(hookPath, 0o755);
+
+      delete process.env.ALLOWCOMMIT;
+      await assert.isRejected(git.exec(['commit', '--allow-empty', '-m', 'commit yo']), /ALLOWCOMMIT/);
+
+      process.env.ALLOWCOMMIT = 'true';
+      await git.exec(['commit', '--allow-empty', '-m', 'commit for real']);
```
smashwilson: Maybe worth asserting that the promise is resolved here? Maybe even that `HEAD` has moved... ?

BinaryMuse: If this promise rejects, the test will fail. Maybe asserting with `isResolved` is more intention revealing? Not sure.

smashwilson: Yah. Not a big deal

PR #924, opened 2017-06-14
More PR timeline events

BinaryMuse:
- [x] Force pushes
- [ ] PR cross-references (pending https://github.com/github/github/issues/74424)
- [x] Commit comment threads (pending https://github.com/github/platform-interface/issues/478)
- [ ] Labels
- [ ] Base ref changed (pending GraphQL API changes)

PR #1084, opened 2017-08-04
Show PR build statuses

kuychaco:
This PR implements the query and display of build statuses for pull requests.


BinaryMuse:
As we started on this PR, it grew in scope a little bit:

* Merged in #999, which brought in the latest version of Relay
* Merged in #924 early, which included some changes to the way we group timeline events of the same type that we wanted to include
* Updated the entire application to use Relay modern (first switching to `relay/compat` and then moving to the modern core)

BinaryMuse:
@kuychaco What do you think about merging this now and taking care of the pagination issue in a future PR?

PR #1101, opened 2017-08-15
Fix regression for updating the discard history after loading

### LINKED ISSUE #1001: Git removed/deleted all of my files from desktop
I accidentally (without fully understanding what I was doing) clicked the "Create Repository" button while viewing my projects, and a git was created on my desktop. Because I had no idea how a git was used with Atom, I thought I was simply going to link Atom to an already created git. Of course, this was my fault for not reading what it said directly above the button "Initialize this project directory with a git repository"... I got way to ahead of myself, and didn't understand the possible dangers or what it meant.

After Atom tried to add everything on my desktop to the unstaged changes list, I clicked out of atom; this is when I noticed that every file on my desktop was gone, and the only thing in its place was a .git folder. I lost pictures, project files, scripts and pieces of software that was on my desktop.. 

The git (as far as I'm aware, as I am new to using git), did not push my content to an online repo; but then again, where are all my files then? They should have still been left untouched on my desktop, shouldn't they? But they're all gone. Is it possible the files have been uploaded somewhere? or were they deleted?

The .git folder contains about 1.75GB of content.

After opening Atom again, the files that were on my desktop and were previously shown in Atom are no longer shown anywhere under the git bars or in the project view. The only files shown are leftover folders in white, and 1 is green (a .exe that starts with the pc, so it couldn't be parsed).. 

If there are any suggestions that can help me out and recover my files, I'd very very much appreciate it.

Btw, I highly recommend the git creation feature be removed/or at least relocated or improved. I can see linking an already created git, but what Atom did was completely unnecessary and should not have happened with a single click. There should be two buttons at least, one that includes "find git repository folder" and or "Create Repository". But, there should always be a warning or notice first asking if your sure you want to create the git in that location, with the locations name listed in a pop up window... I didn't even realize what directory it was going to be created in until after the fact.

And yes, I understand this primarily occurred because of my lack of knowledge in Github or Atom, but it was all too easy for what happened to actually happen anyways. and I believe more safety precautions need to be in place for people like myself. I lost a lot of content, pics, project files, text documents... a lot of stuff that I thought was somewhat safe on my desktop.

Again, I thank anyone who can help me recover my files; months of work has just disappeared, and I am lost now.

### LINKED ISSUE #1104: "Undo Discard" button not displayed after reload
Reference https://github.com/atom/github/issues/1001#issuecomment-312608775

As a result discards from a previous session cannot be restored

kuychaco:
Fixes #1104 

This PR fixes a regression that was likely introduced during the refactor of the Repository model into a state machine. The result is that when Atom is reloaded, the discard history is updated appropriately in the model layer, and the "Undo discard" buttons appear in the UI so that previous states can be restored easily.

### DIFF
--- a/lib/models/repository-states/present.js
+++ b/lib/models/repository-states/present.js
@@ -54,6 +54,10 @@ export default class Present extends State {
       this.workdir(),
       {maxHistoryLength: 60},
     );
+
+    if (history) {
+      this.discardHistory.updateHistory(history);
+    }
   }
 
   isPresent() {


smashwilson:
Ohhh nice 🌠

PR #1110, opened 2017-08-17
Exec GPG wrapper script from temp directory

### LINKED ISSUE #1109: GPG support is broken by Atom's ASAR bundling
Oops :boom:

```
15:02:46.969395 git.c:328               trace: built-in: git 'commit' '-m' 'Trust in RVO'
15:02:46.971734 run-command.c:626       trace: run_command: '/Applications/Atom.app/Contents/Resources/app.asar/node_modules/github/bin/gpg-no-tty.sh' '--status-fd=2' '-bsau' '04F69D2A'
fatal: cannot exec '/Applications/Atom.app/Contents/Resources/app.asar/node_modules/github/bin/gpg-no-tty.sh': Not a directory
error: gpg failed to sign the data
fatal: failed to write commit object
```

I'm guessing we just need to do the same "copy to temp directory, pass path as env var" trick we use for the other scripts.

smashwilson:
Copy and execute the GPG wrapper script from the git-operation temporary directory rather than directly from the package source, so that it will work correctly when bundled in the Atom `.asar` file. This is mostly extracted from #846 so I'll have some nice, satisfying merge conflicts once I get back to that PR.

Fixes #1109.


lib/git-prompt-server.js
```diff
@@ -25,25 +26,41 @@ export default class GitPromptServer {
     // TODO: [mkt] Windows?? yes.
     this.promptForInput = promptForInput;
     const windows = process.platform === 'win32';
-    this.tmpFolderPath = await getTempDir('github-');
+    this.tmpFolderPath = await getTempDir({
+      dir: windows ? os.tmpdir() : '/tmp',
```
damieng: Why is this hard-coded outside of Windows instead of just using os.tmpdir?  Seems it would ignore the environment https://github.com/nodejs/node-v0.x-archive/blob/master/lib/os.js#L47

smashwilson: Oh, yeah, there's a long story here. The short version is that we create a Unix domain socket as a subdirectory of this, and on macOS and Linux, [domain socket paths must be less than 104 or 108 bytes long](https://stackoverflow.com/questions/34829600/why-is-the-maximal-path-length-allowed-for-unix-sockets-on-linux-108). On macOS, `os.tmpdir()` returns a path beneath `/private/` that eats up enough of the available space that (sometimes!) the domain socket creation will fail. So, we use `os.tmpdir()` on Windows and hard-code it to the nice, short `/tmp` on other OSes.

Note that this PR didn't actually introduce this, it just lost and restored the call by accident while I was doing `git checkout -p` from my other branch. The `windows` ternary is there on master:

https://github.com/atom/github/blob/7835bc46d9f8dec07b6072293b788866db191bff/lib/git-prompt-server.js#L29-L33

Also, this is absolutely a hack, and I'd love to find a better way to ensure that the socket can be created successfully.

/cc @BinaryMuse who explained this to _me_ the first time I saw this code :wink:, in case she has anything to add

PR #1129, opened 2017-08-22
Add line-height to sync icon animation to center

jontewks:
### Description of the Change
Right now the sync animation icon in the bottom toolbar rotates off center and moves up and down a bit rather than spinning in place. Adding this line height resolves that.

Current:
current

Updated line-height:
with line height

### Alternate Designs

I considered where else to modify the css, or how else besides line height, but this seemed the most straight forward with the lowest likelihood of unintended side-effects

### Benefits

Animation will now look as intended

### Possible Drawbacks

Possible the css has other effects I couldn't see or find, but it seems very unlikely

### Applicable Issues

None found

### DIFF
--- a/styles/status-bar-tile-controller.less
+++ b/styles/status-bar-tile-controller.less
@@ -39,6 +39,7 @@
       100% { transform: rotate(360deg); }
     }
     animation: github-StatusBarSync-animation 2s linear 30; // limit to 1min in case something gets stuck
+    line-height: 0.9;
   }
 
   // Merge conflict icon


BinaryMuse:
I looked at this today and for a while I thought the "wiggle" is caused by the 0.25em `margin-right` set in `.status-bar .icon::before`, presumably from core styles.

github-package-spinner-oddity

But, removing it didn't seem to help. Deferring to @simurai about the best solution to this.

simurai:
@jontewks What theme are you using? With One dark the icon isn't centered by 1 pixel:

screen shot 2017-08-25 at 10 08 55 am

But I have been [fiddling with icon alignments](https://github.com/atom/one-dark-ui/pull/216) before, so current stable version is probably different. Not quite sure what changed. Maybe how Chromium renders text?

It seems every theme needs to manually nudge the icon to find the right center point. 🤔  Other options might be:

- Animate the arrows up/down to show progress
- Switch to a custom SVG icon where alignment is more predictable

jontewks:
Hello @simurai. I am using One dark on Atom 1.19.

simurai:
Ok, I think this is impossible to fix with the current font icons. For example, if the icon gets nudged to be in the center with 2px space around it:

screen shot 2017-08-26 at 9 14 39 am
screen shot 2017-08-26 at 9 15 09 am

As soon as something changes in the status-bar, for example if there are only "9 updates", the horizontal centering is off by a bit and makes the icon wiggle when animated:

screen shot 2017-08-26 at 9 17 48 am
screen shot 2017-08-26 at 9 17 15 am

I guess fonts now get sub-pixel rendering, which is great, but not good for icons that should stay sharp. Another reason to switch our icons to SVG.

simurai:
It should be "fixed", or at least improved with: https://github.com/atom/one-light-ui/pull/108

@jontewks Thanks for this PR. 🙇

Issue #830, opened 2017-05-17
UX: Pinning repo feature is a somewhat confusing concept


PandelisZ:
### Description

The auto pinning feature feels slightly confusing initially. I sometimes just browse through a PR and read the comments and discussions. This is a case where I wouldn't want to automatically be pinned to the PR. This default behaviour makes me feel stuck in this PR

### Steps to Reproduce

1. Click into a PR within the GitHub pane

**Expected behavior:**

Now browsing this PR and looking through discussion with the option to return to the all the PRs

Proposal to add a back button as shown below

<img width="405" alt="screen shot 2017-05-17 at 12 21 46 copy" src="https://cloud.githubusercontent.com/assets/8436717/26151882/99c64b0c-3afc-11e7-8193-16d44d8b1cc0.png">

**Actual behavior:**

User needs to click on unpin to return to the PR list

<img width="405" alt="screen shot 2017-05-17 at 12 21 46" src="https://cloud.githubusercontent.com/assets/8436717/26151907/b70f2af8-3afc-11e7-886c-90ebcf9bb979.png">

**Reproduces how often:**

Every time 100%

### Versions

github 0.0.9
atom 1.18.0-beta0
Mac OS - 10.12.3

kuychaco:
@simurai for 💭

simurai:
> Click into a PR within the GitHub pane

It sounds like you're seeing a list of PRs, like in this [screenshot](https://github.com/atom/github/pull/714#issuecomment-298240407). That's probably because you're on `master` branch and there isn't a PR for it. The current GitHub panel is more a "here is the PR for your **current** branch, but in case we got it wrong or a PR can't be found, you can pin another PR".

> I sometimes just browse through a PR

Yes, that is something we were thinking too. An idea is to have a search/autocomplete dropdown, see https://github.com/atom/github/pull/714#issuecomment-298145826. Or maybe once we have a list of branches, we could show which ones have a PR. And it depends if we wanna add notifications like on github.com. I usually just browse PRs if they are new or I get pinged or asked to review.

simurai:
Here an exploration of a mini GitHub browser: https://github.com/atom/github/issues/1019

It wouldn't just be for PRs, but anything that the API allows to be searched for.

annthurium:
we have different github package behavior now that doesn't include pinning repos.  Going to close this out.  😄

PR #1138, opened 2017-08-23
Create a new pull request from unsuccessful PR selection views

BinaryMuse:
When a PR can't be found for a branch, or many PRs are found for a branch, offer a link to the create PR page for the branch on GitHub.com.

Also adds a new button to re-do the search, useful if the user creates the PR in another window and wants to display it in Atom.

cursor_and_github__preview__ ___atom

### DIFF
--- a/lib/action-pipeline.js
+++ b/lib/action-pipeline.js
@@ -1,0 +1,76 @@
+function partial(fn, ...args) {
+  return function wrapped() {
+    return fn(...args);
+  };
+}
+
+export class ActionPipeline {
+  constructor(action) {
+    this.action = action;
+    this.middleware = [];
+    this.middlewareNames = [];
+  }
+
+  run(fn, ...args) {
+    // eslint-disable-next-line prefer-spread
+    return this.middleware
+      .map(middleware => middleware.fn)
+      .reduceRight((acc, nextFn) => partial(nextFn, acc, ...args), partial(fn, ...args))
+      .apply(undefined, args);
+  }
+
+  addMiddleware(name, fn) {
+    if (!name) {
+      throw new Error('Middleware must be registered with a unique middleware name');
+    }
+
+    if (this.middleware.some(mw => name === mw.name)) {
+      throw new Error(`A middleware with the name ${name} is already registered`);
+    }
+
+    this.middleware.push({name, fn});
+  }
+
+  removeMiddleware(name) {
+    this.middleware = this.middleware.filter(mw => mw.name !== name);
+  }
+
+  insertMiddlewareBefore(needle, name, fn) {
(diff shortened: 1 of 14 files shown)


smashwilson:
Random, drive-by bikeshedding:

* I feel like "open a new pull request" should be the first, more visually dominant option. It seems like it'd be the most common, happy-path next step here; create a branch, create a PR.
* Maybe change the text "Submit" to "Pin"? "Submit" sounds a bit like "submit PR", especially right next to the actual create PR action.

(As always, feel free to disregard entirely :smile:)

simurai:
> @smashwilson (As always, feel free to disregard entirely 😄)

I regard entirely 😄 . Here how a flip around version could be:

image

50Wliu:
I think in the case where a branch hasn't been pushed yet, only one button is needed that could read "Push and open new pull request".  Because you can already push through the status bar.

BinaryMuse:
@smashwilson @simurai Totally agree, good call. Love the mockup, I'll work on that.

I punted on pushing from this UI at first because the GitHub tab components don't have the active repository passed down through it like the Git tab does, but this is something I can remedy. :)

BinaryMuse:
@simurai We've gotten feedback before (#830) that "pinning" is somewhat confusing. I think the language is probably problematic. What do you think about changing this language to something like "linking a pull request with this branch," or "associating" or something else similar?

simurai:
> @50Wliu only one button is needed that could read "Push and open new pull request".

Yeah, that would be even better.

> @BinaryMuse I punted on pushing from this UI at first because the GitHub tab components don't have the active repository passed down through it like the Git tab does, but this is something I can remedy. :)

Ah, you mean that the GitHub tab doesn't know anything about the status (pushed or not) of the current repo? Also, I guess opening the browser with the github.com compare page would need to be delayed until the push is done, right? So that it can show the changes.

simurai:
> @BinaryMuse We've gotten feedback before (#830) that "pinning" is somewhat confusing. I think the language is probably problematic. What do you think about changing this language to something like "linking a pull request with this branch," or "associating" or something else similar?

Hmm.. yeah. Maybe "you can manually link a GitHub pull request..". 👍  "Associate" is kinda mouthful.

krainboltgreene:
Would love to have this, any chance I can help?

kuychaco:
@krainboltgreene thanks for the comment and offer to help! We're currently working on a refactor to better enable this feature in #1230. So hopefully we'll pick this back up soon :)

am:
Any progress on this? Looks like the refactor is merged. Is there any plans to continue this feature?

smashwilson:
Closing in favor of #1376 since the codebase has moved on quite a bit since this was opened.

PR #1153, opened 2017-09-06
Handle repositories that are very large and/or that exist in home or root directories

### LINKED ISSUE #953: Atom crashing with OOM errors possibly related to Git usage
@joshgibs at https://github.com/atom/atom/issues/14806#issuecomment-309077338
Crash dump: https://github.com/atom/atom/files/1081430/atom.crash.txt

<details>
<summary>Crash dump sample</summary>

```
Path:                  /Applications/Atom.app/Contents/Frameworks/Atom Helper.app/Contents/MacOS/Atom Helper
Identifier:            com.github.atom.helper
Version:               1.18.0 (1.18.0)
Code Type:             X86-64 (Native)
Parent Process:        Atom [2953]
Responsible:           Atom [2953]
User ID:               501

Date/Time:             2017-06-16 12:44:04.404 -0400
OS Version:            Mac OS X 10.12.5 (16F2073)
Report Version:        12
Anonymous UUID:        FEB24DAE-B179-8357-7B7F-B63CA05D15B6


Time Awake Since Boot: 15000 seconds

System Integrity Protection: enabled

Crashed Thread:        0  CrRendererMain  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000

Thread 0 Crashed:: CrRendererMain  Dispatch queue: com.apple.main-thread
0   com.github.electron.framework       0x0000000101ac9344 0x101a06000 + 799556
1   com.github.electron.framework       0x0000000101ac92d6 0x101a06000 + 799446
2   libnode.dylib                       0x000000010578280a 0x10540f000 + 3618826
3   libnode.dylib                       0x00000001057827ae v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) + 526
4   libnode.dylib                       0x0000000105aafc17 v8::internal::Factory::NewFixedArray(int, v8::internal::PretenureFlag) + 455
5   libnode.dylib                       0x0000000105bf549f 0x10540f000 + 8283295
6   libnode.dylib                       0x0000000105bee26c v8::internal::SeededNumberDictionary::AddNumberEntry(v8::internal::Handle<v8::internal::SeededNumberDictionary>, unsigned int, v8::internal::Handle<v8::internal::Object>, v8::internal::PropertyDetails, bool) + 140
7   libnode.dylib                       0x0000000105a95262 0x10540f000 + 6840930
8   libnode.dylib                       0x0000000105bcb283 v8::internal::JSObject::AddDataElement(v8::internal::Handle<v8::internal::JSObject>, unsigned int, v8::internal::Handle<v8::internal::Object>, v8::internal::PropertyAttributes, v8::internal::Object::ShouldThrow) + 1283
9   libnode.dylib                       0x0000000105d1ba4f v8::internal::Runtime::SetObjectProperty(v8::internal::Isolate*, v8::internal::Handle<v8::internal::Object>, v8::internal::Handle<v8::internal::Object>, v8::internal::Handle<v8::internal::Object>, v8::internal::LanguageMode) + 191
10  libnode.dylib                       0x0000000105d1efa1 v8::internal::Runtime_SetProperty(int, v8::internal::Object**, v8::internal::Isolate*) + 129
11  ???                                 0x000008725cb084e7 0 + 9287274366183
12  ???                                 0x000008725dac49a7 0 + 9287290866087
13  ???                                 0x000008725d6c49bc 0 + 9287286671804
14  ???                                 0x000008725cb086f5 0 + 9287274366709
15  ???                                 0x000008725d6e97e2 0 + 9287286822882
16  ???                                 0x000008725d98a3db 0 + 9287289578459
17  ???                                 0x000008725cb6cb55 0 + 9287274777429
18  ???                                 0x000008725d96034a 0 + 9287289406282
19  ???                                 0x000008725d963da1 0 + 9287289421217
20  ???                                 0x000008725d6deee6 0 + 9287286779622
21  ???                                 0x000008725d6deb57 0 + 9287286778711
22  ???                                 0x000008725cb74ca3 0 + 9287274810531
23  ???                                 0x000008725cb6496f 0 + 9287274744175
24  libnode.dylib                       0x0000000105aa694b 0x10540f000 + 6912331
25  libnode.dylib                       0x0000000105aa674b v8::internal::Execution::Call(v8::internal::Isolate*, v8::internal::Handle<v8::internal::Object>, v8::internal::Handle<v8::internal::Object>, int, v8::internal::Handle<v8::internal::Object>*) + 331
26  libnode.dylib                       0x0000000105aa6b99 v8::internal::Execution::TryCall(v8::internal::Isolate*, v8::internal::Handle<v8::internal::Object>, v8::internal::Handle<v8::internal::Object>, int, v8::internal::Handle<v8::internal::Object>*, v8::internal::MaybeHandle<v8::internal::Object>*) + 121
27  libnode.dylib                       0x0000000105b959e7 v8::internal::Isolate::RunMicrotasksInternal() + 759
28  libnode.dylib                       0x0000000105b9552a v8::internal::Isolate::RunMicrotasks() + 42
29  libnode.dylib                       0x00000001054ef900 node::RunMicrotasks(v8::FunctionCallbackInfo<v8::Value> const&) + 16
```

</details>

Also:

> It definitely crashes when doing any auto complete (like Lorem or python
methods), but also crashes while saving or sometimes while sitting idle.

and

> Disabled github package and it has been stable so far.

---

@cthomaschase is having the same issue at https://github.com/atom/atom/issues/14806#issuecomment-309198058
Crash dump: https://github.com/atom/atom/files/1082404/Atom.Helper_2017-06-16-211121_bullitt.crash.zip

BinaryMuse:
Repositories that have output for our `status` call that is greater than a given threshold (set to 10 MB here) are considered too large for Atom to handle and are moved to the `TooLarge` state, which we detect and use to show a friendly message to users.

cursor_and__

Additionally, if we detect the repository is in the user's home directory, we show slightly different messaging.

cursor_and_git_ __

Messaging is WIP; ping @jasonrudolph for awesome Atom voice 💭 s. ❤️ 

Fixes #953 
Fixes atom/atom#15443
Fixes atom/atom#15493

TODO:

- [x] Messaging
- [x] ~Create Discuss posts to link to~
- [x] Styling

### DIFF
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -19,11 +19,20 @@ import GitTimingsView from './views/git-timings-view';
 import WorkerManager from './worker-manager';
 
 const LINE_ENDING_REGEX = /\r?\n/;
+const MAX_STATUS_OUTPUT_LENGTH = 1024 * 1024 * 10;
 
 let headless = null;
 let execPathPromise = null;
 
 export class GitError extends Error {
+  constructor(message) {
+    super(message);
+    this.message = message;
+    this.stack = new Error().stack;
+  }
+}
+
+export class LargeRepoError extends Error {
   constructor(message) {
     super(message);
     this.message = message;
@@ -357,6 +366,10 @@ export default class GitShellOutStrategy {
   async getStatusBundle() {
     const args = ['status', '--porcelain=v2', '--branch', '--untracked-files=all', '--ignore-submodules=dirty', '-z'];
     const output = await this.exec(args);
+    if (output.length > MAX_STATUS_OUTPUT_LENGTH) {
+      throw new LargeRepoError();
+    }
+
     const results = await parseStatus(output);
 
     for (const entryType in results) {
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -1,5 +1,6 @@
 import path from 'path';
 import fs from 'fs-extra';
+import os from 'os';
 import temp from 'temp';
 import {ncp} from 'ncp';
 
@@ -113,6 +114,14 @@ export function firstImplementer(...targets) {
       }, Object.prototype);
     },
   });
+}
+
+function isRoot(dir) {
+  return path.resolve(dir, '..') === dir;
+}
+
+export function isValidWorkdir(dir) {
+  return dir !== os.homedir() && !isRoot(dir);
 }
 
 export function readFile(absoluteFilePath, encoding = 'utf8') {
--- a/lib/models/repository-states/present.js
+++ b/lib/models/repository-states/present.js
@@ -2,6 +2,7 @@ import path from 'path';
 
 import State from './state';
 
+import {LargeRepoError} from '../../git-shell-out-strategy';
 import {deleteFileOrFolder} from '../../helpers';
 import {FOCUS} from '../workspace-change-observer';
 import FilePatch from '../file-patch';
@@ -388,13 +389,27 @@ export default class Present extends State {
 
   getStatusBundle() {
     return this.cache.getOrSet(Keys.statusBundle, async () => {
-      const bundle = await this.git().getStatusBundle();
-      const results = await this.formatChangedFiles(bundle);
-      results.branch = bundle.branch;
-      if (!results.branch.aheadBehind) {
-        results.branch.aheadBehind = {ahead: null, behind: null};
+      try {
+        const bundle = await this.git().getStatusBundle();
+        const results = await this.formatChangedFiles(bundle);
+        results.branch = bundle.branch;
+        if (!results.branch.aheadBehind) {
+          results.branch.aheadBehind = {ahead: null, behind: null};
+        }
+        return results;
+      } catch (err) {
+        if (err instanceof LargeRepoError) {
+          this.transitionTo('TooLarge');
+          return {
+            branch: {},
+            stagedFiles: [],
+            unstagedFiles: [],
+            mergeConflictFiles: [],
+          };
+        } else {
+          throw err;
+        }
       }
-      return results;
     });
   }
 
--- a/lib/models/workdir-cache.js
+++ b/lib/models/workdir-cache.js
@@ -1,7 +1,7 @@
 import path from 'path';
 import fs from 'fs';
 
-import {readFile} from '../helpers';
+import {readFile, isValidWorkdir} from '../helpers';
 
 /**
  * Locate the nearest git working directory above a given starting point, caching results.
@@ -56,6 +56,10 @@ export default class WorkdirCache {
       let currentDir = initialDir;
 
       const check = () => {
+        if (!isValidWorkdir(currentDir)) {
+          return walk();
+        }
+
         const dotGit = path.join(currentDir, '.git');
         fs.stat(dotGit, async (statError, stat) => {
           if (statError) {
@@ -79,6 +83,7 @@ export default class WorkdirCache {
           // .git directory found! Mission accomplished.
           return resolve(currentDir);
         });
+        return null;
       };
 
       const walk = () => {
--- a/lib/views/git-tab-view.js
+++ b/lib/views/git-tab-view.js
@@ -67,8 +67,11 @@ export default class GitTabView {
     } else if (this.props.repository.showGitTabInit()) {
       const inProgress = this.props.repository.showGitTabInitInProgress();
       const message = this.props.repository.hasDirectory() ?
-        'Initialize this project directory with a git repository' :
-        'Initialize a new project directory with a git repository';
+        (
+          <span>Initialize <strong>{this.props.repository.getWorkingDirectoryPath()}</strong> with a
+          Git repository</span>
+        ) :
+          <span>Initialize a new project directory with a Git repository</span>;
 
       return (
         <div className="github-Panel is-empty" tabIndex="-1">
--- a/lib/views/git-tab-view.js
+++ b/lib/views/git-tab-view.js
@@ -1,8 +1,6 @@
 /** @jsx etch.dom */
 /* eslint react/no-unknown-property: "off" */
 
-import os from 'os';
-
 import etch from 'etch';
 import {autobind} from 'core-decorators';
 import cx from 'classnames';
@@ -10,6 +8,7 @@ import cx from 'classnames';
 import StagingView from './staging-view';
 import GitLogo from './git-logo';
 import CommitViewController from '../controllers/commit-view-controller';
+import {isValidWorkdir} from '../helpers';
 
 export default class GitTabView {
   static focus = {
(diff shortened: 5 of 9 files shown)


simurai:
Re. messaging: Might be helpful to have a title like "Too many changes". So that when you keep seeing these kind of messages, it's quick to know what they are about and you don't have to read through them every time.

lib/views/git-tab-view.js
```diff
@@ -32,7 +34,30 @@ export default class GitTabView {
...
+        <div className="github-Panel is-empty" tabIndex="-1">
+          <div ref="noRepoMessage" className="github-Panel no-repository">
+            <div className="large-icon">
+              <GitLogo />
+            </div>
+            {workingDir === os.homedir()
+              ? <div className="initialize-repo-description">
+                  The number of changed files in the repository at <strong>{workingDir}</strong> is
+                  too large for Atom to manage. In addition, the repository is located at your
+                  home directory, which may be unintended. <a href="https://gist.github.com/BinaryMuse/something">More details</a>
```
jasonrudolph: The best error messages help the user understand how to resolve the error, and it would be difficult to explain resolution steps in sufficient detail here. I :heart: that you're linking to a resource that can provide more details. :zap:

In the scenario where we detect a repository in the user's home directory, am I correct in thinking that the external resource will recommend that the user consider deleting that repository (assuming they haven't intentionally created that repository)?

lib/views/git-tab-view.js
```diff
@@ -32,7 +34,30 @@ export default class GitTabView {
...
+              </div>
+              : <div className="initialize-repo-description">
+                  The number of changed files in the repository at <strong>{workingDir}</strong> is
+                  too large for Atom to manage. Ensure that the repository path is correct and
```
jasonrudolph: Can you help me understand what it means to "ensure that the repository path is correct?" Is there a scenario where the repository path would be incorrect?

lib/views/git-tab-view.js
jasonrudolph: > Might be helpful to have a title like "Too many changes". So that when you keep seeing these kind of messages, it's quick to know what they are about and you don't have to read through them every time. -- https://github.com/atom/github/pull/1153#issuecomment-327656181

I agree.

---

> The number of changed files in the repository at **/Users/mtilley/** is too large for Atom to manage.

If we reorder this a bit, we can shorten this sentence. I also think the reordering increases the simplicity of the sentence, which increases the readability:

> The repository at **/Users/mtilley/** has too many changed files to display in Atom.

---

Putting this all together, I think this does a nice job of conveying the situation:

> ### Too many changes
> 
> The repository at **/Users/mtilley/** has too many changed files to display in Atom. In addition, the repository is located at your home directory, which may be unintended. [More details](#)

kuychaco: Thanks @jasonrudolph! Love what you came up with ✨ 

> In the scenario where we detect a repository in the user's home directory, am I correct in thinking that the external resource will recommend that the user consider deleting that repository (assuming they haven't intentionally created that repository)?

yup! 👍

lib/views/git-tab-view.js
kuychaco: Yeah I agree that this wording is a bit unclear. I _think_ it was mostly referring to the case when the user accidentally had a .git folder in their home directory. Which is addressed in the other case so we can probably just drop this wording here. @BinaryMuse 💭s?

BinaryMuse:
I've been working on crafting a message or two this morning for the "read more" links, but having a particularly tough time coming up with a good way to describe the problem and potential solutions without it being very technical. Wondering if anyone has any thoughts on the best way to communicate this.

BinaryMuse:
After discussing things with @kuychaco, we decided to fully enumerate the common possible states and have distinct messaging for each, which removes the need to explain things to the user. In addition, we special case the home directory (and root directories) so that it's impossible for the user to get into a weird state based on having a repo there.

Here's the matrix of new possibilities:

* User has a repo in their home directory
  * User opens home directory specifically 👉 show an "unsupported directory" message
  * User opens a subfolder that does not contain a repo 👉 when we walk up the tree to find a repo in parent folders, we skip over the home (and root) directories, ensuring a Git repo in the home directory is never chosen
* User does not have a repo in their home directory
  * User opens home directory specifically 👉  show an "unsupported directory" message (instead of the "initialize a repo here" button)
  * User opens a subfolder 👉 same as always (get the Git UI or the "create repo" button as appropriate)

We also kept the " too large" state as Atom will choke on big repos, and this will keep us from crashing.

BinaryMuse:
@smashwilson I think we're pretty happy with the state of this, but we'd love to get your opinions on the most recent decisions (and also a technical review :bowing_woman:)

lib/helpers.js
```diff
@@ -115,6 +116,14 @@ export function firstImplementer(...targets) {
   });
 }
 
+function isRoot(dir) {
+  return path.resolve(dir, '..') === dir;
```
smashwilson: :+1:

lib/helpers.js
```diff
@@ -115,6 +116,14 @@ export function firstImplementer(...targets) {
...
+}
+
+export function isValidWorkdir(dir) {
+  return dir !== os.homedir() && !isRoot(dir);
```
smashwilson: Note that this won't handle things like relative paths or symlinks.

That's probably okay for this, though, since we just want to explicitly exclude `~` and `/` mistakes, and if someone slips one of them through with filesystem trickery the TooLarge state will catch it.

lib/models/repository-states/present.js
```diff
@@ -388,13 +389,27 @@ export default class Present extends State {
...
+        return results;
+      } catch (err) {
+        if (err instanceof LargeRepoError) {
+          this.transitionTo('TooLarge');
+          return {
+            branch: {},
+            stagedFiles: [],
+            unstagedFiles: [],
+            mergeConflictFiles: [],
+          };
```
smashwilson: Ah, so we bail out entirely if a repo is too large? I was wondering if we'd be able to do some kind of partial status in that case with the streaming API. That's probably too error-prone though.

lib/models/workdir-cache.js
```diff
@@ -56,6 +56,10 @@ export default class WorkdirCache {
       let currentDir = initialDir;
 
       const check = () => {
+        if (!isValidWorkdir(currentDir)) {
+          return walk();
+        }
```
smashwilson: :+1: I like that we're skipping `~` and `/` when we're traversing directories. Why even run stat if we're just going to throw it out.

lib/git-shell-out-strategy.js
```diff
@@ -19,6 +19,7 @@ import GitTimingsView from './views/git-timings-view';
 import WorkerManager from './worker-manager';
 
 const LINE_ENDING_REGEX = /\r?\n/;
+const MAX_STATUS_OUTPUT_LENGTH = 1024 * 1024 * 10;
```
smashwilson: Any idea what this roughly translates to in terms of file count?

We wouldn't want to accidentally lock out .com or Linux kernel devs 😅

BinaryMuse: Unfortunately it depends on path lengths. So deeply nested projects will be penalized more.

lib/helpers.js
50Wliu: :heart: this just gave me an idea for one of the bugs I've been trying to fix!

smashwilson:
:tada: :tada: :tada:

matzipan:
Hi, not sure if this is the appropriate place to leave this comment in (should I open an issue?). It seems to be triggering into a fairly average repository (.git of size 37M) that Atom was perfectly able to handle before I did the update. 

Furthermore, if the issue is that the repository it's to big, then the displayed message is a bit misleading, since it says "Too many chages".

PR #1165, opened 2017-09-09
Check for deep equality of selected items after update

### LINKED ISSUE #1163: Pending diff view opens on saving file when panes are split
Repro steps:

1. Have two panes open
2. In left pane make a **pending** diff view the active item
3. In right pane, save a file

Expected: the file simply saves
Actual: a diff view opens in the right pane for the file selected in the changed files list.

Note: if the diff view in the left pane is NOT pending, then all works as expected.

/cc @maxbrunsfeld @BinaryMuse

kuychaco:
When there is a file system change, the repo emits a change event which results in a an update to `StagingView`. As part of this, the changed file information is fetched from git and updated in the UI. We check to see if the changed files selection has changed, and if it has, we call `didChangeSelectedItems` which opens the diff view associated with the newly selected file.

Previously we were checking for object reference equality to see if the selection had changed, but we need to be checking for deep equality since we're fetching new changed file objects. This was causing funky behavior described in #1163. This PR uses `lodash`'s `isEqual` to check for deep equality before calling `didChangeSelectedItems`.

Fixes #1163

### DIFF
--- a/lib/views/staging-view.js
+++ b/lib/views/staging-view.js
@@ -143,9 +143,7 @@ export default class StagingView {
       await this.resolutionProgressObserver.setActiveModel(this.props.resolutionProgress);
     }
 
-    const previousSelection = Array.from(previouslySelectedItems).sort();
-    const newSelection = Array.from(this.selection.getSelectedItems()).sort();
-    const selectionChanged = !isEqual(previousSelection, newSelection);
+    const selectionChanged = !isEqual(previouslySelectedItems, this.selection.getSelectedItems());
     if (selectionChanged) {
       this.debouncedDidChangeSelectedItem();
     }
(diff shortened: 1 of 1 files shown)


lib/views/staging-view.js
```diff
@@ -143,8 +143,10 @@ export default class StagingView {
       await this.resolutionProgressObserver.setActiveModel(this.props.resolutionProgress);
     }
 
-    const {added, removed} = compareSets(previouslySelectedItems, this.selection.getSelectedItems());
-    if (added.size > 0 || removed.size > 0) {
+    const previousSelection = Array.from(previouslySelectedItems).sort();
+    const newSelection = Array.from(this.selection.getSelectedItems()).sort();
+    const selectionChanged = !isEqual(previousSelection, newSelection);
```
BinaryMuse: Do we need to convert these to arrays?

```javascript
a = new Set([1, 2, 3])
//=> Set(3) {1, 2, 3}
b = new Set([2, 1, 3])
//=> Set(3) {2, 1, 3}
_.isEqual(a, b)
//=> true
```

kuychaco: nooope, totally unnecessary. Nice catch :)

PR #1167, opened 2017-09-12
Add text clipping for 'Create Repo' button

kentywang:
### Description of the Change

Hi there! First time contributing to Atom :)

The `Create repository` button on the Github panel when there is no repository open allows for the text inside to leak out when the panel is resized to small dimensions. It could be improved by adding two rules for the button element's style: `overflow: hidden` and `text-overflow: ellipses`. This will replace the overflown text with ellipses when horizontally shrunk and will clip the text when vertically shrunk.

### Alternate Designs

Before: | After: 
--- | ---
 | 
 | 

### Benefits

No more text leaking from the button :)

### Possible Drawbacks

There should not be any other side-effects; the rule is specific to this one element.

Thank you, Atom team, for all your hard work!

### DIFF
--- a/styles/git-panel.less
+++ b/styles/git-panel.less
@@ -43,6 +43,11 @@
     .git-logo-path {
       fill: mix(@base-background-color, @text-color, 66%); // 2/3 of bg color
     }
+
+    .btn.btn-primary {
+      overflow: hidden;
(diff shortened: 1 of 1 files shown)


BinaryMuse:
Thanks very much for contributing! :bowing_woman:

PR #1174, opened 2017-09-15
Update to React 16

BinaryMuse:
Gets us portals, error boundaries, performance and more.

- [x] Update to React 16
- [x] Update to Enzyme 3
- [ ] Update all Portal-based components
- [ ] Fix Enzyme tests
- [ ] Verify snapshotting works


BinaryMuse:
It's likely we can completely get rid of our `<Portal />` wrapper container and replace it entirely with `ReactDOM.createPortal` + refs.

BinaryMuse:
Blocked by https://github.com/airbnb/enzyme/issues/928

BinaryMuse:
Enzyme v3 is out and some stuff changed. You can find [the full migration guide here](https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md), but in particular we're gonna be bit by the "for `mount`, updates are sometimes required when they weren't before" bit.

> enzyme v3 architecturally created a layer where React would create an intermediate representation of the render tree at an instance in time and pass that to enzyme to traverse and inspect. This has many advantages, but one of the side effects is that now the intermediate representation does not receive automatic updates.

This means that any test where some async operation asynchronously calls a `setState` will *not* automatically result in the Enzyme wrapper being updated unless we call `wrapper.update()` to trigger to Enzyme that an update occurred.

Luckily, `wrapper.update()` returns the wrapper instance, so the changes should mostly look like

```diff
diff --git a/test/controllers/file-patch-controller.test.js b/test/controllers/file-patch-controller.test.js
index f7e63f1c..5d6505e3 100644
--- a/test/controllers/file-patch-controller.test.js
+++ b/test/controllers/file-patch-controller.test.js
@@ -230,7 +230,7 @@ describe('FilePatchController', function() {
       getFilePatchForPath.returns(filePatch);
 
       wrapper.instance().onRepoRefresh(repository);
-      await assert.async.isTrue(wrapper.find('HunkView').exists());
+      await assert.async.isTrue(wrapper.update().find('HunkView').exists());
       assert.isTrue(wrapper.find('HunkView').text().includes('@@ -0,1 +0,1 @@'));
     });
```

but I wanted to document this here in case others run into it.

BinaryMuse:
Realized while working this that there's a chance React 16 will break snapshotting; something to look into before merging.

BinaryMuse:
Found some odd behaviors that seem to be related to https://github.com/airbnb/enzyme/issues/1163; now blocked on that issue.

BinaryMuse:
For notes about Babel 7 in relation to React 16.2 (for Fragment JSX support), see https://babeljs.io/blog/2017/09/12/planning-for-7.0. Seems this is still pretty early so it probably doesn't make sense to create an Atom transpiler package for Babel 7 just yet (if we need to at all).

smashwilson:
Done in #1390. :tada:

PR #1182, opened 2017-09-20
Align changed files count

### LINKED ISSUE #1198: Incorrect styles on github-ChangedFilesCount causes status bar tile to appear unaligned
### Description

It looks like there's a few issues with the `github-ChangedFilesCount` status bar tile.  First of all, it's outside the `span` that the other two GitHub tiles are contained in (branch select, push/pull).  Then it's an `a` instead of a `div`.  But more importantly, it has the `icon` class assigned to the tile itself rather than a sub-span, causing the tile to be ~2 pixels lower than the rest of the tiles.

github-changedfilecount-issues

/cc @simurai maybe when you're back

sethlu:

### Requirements

### Description of the Change

This PR addresses a slight vertical visual offset by introducing a div wrapper around the anchor object preventing `vertical-align` inherited from `.icon`, preserving the same alignment standards.

### Alternate Designs

- Override `vertical-align: middle` inherited from `.icon` for the anchor object. However, this may result in weird text aligning.

### Benefits

It allows the background color on hover to span the whole column vertically, resolving a light margin above the clickable element.

Before change:

screenshot 2017-09-19 23 48 40

After change:

screenshot 2017-09-20 00 00 50

### Possible Drawbacks

N/A

### Applicable Issues

Fixes #1198

### DIFF
--- a/styles/changed-files-count-view.less
+++ b/styles/changed-files-count-view.less
@@ -4,8 +4,15 @@
 
 .github-ChangedFilesCount {
 
-  &.icon-diff::before {
-    margin-right: .2em;
+  &.icon.icon-diff {
+    display: inline-block;
+    vertical-align: top;
+    line-height: 2.5;
+
+    &::before {
+      margin-right: .2em;
+      vertical-align: text-bottom; // Visually adjusted
+    }
   }
 
 }
--- a/styles/push-pull-view.less
+++ b/styles/push-pull-view.less
@@ -7,6 +7,12 @@
   &-icon.icon.icon.icon {
     margin-right: 0;
     text-align: center;
+    vertical-align: top;
+    line-height: 2.5;
+
+    &::before {
+      vertical-align: text-bottom; // Visually adjusted
+    }
   }
 
   &-label.is-pull {
--- a/styles/status-bar-tile-controller.less
+++ b/styles/status-bar-tile-controller.less
@@ -46,6 +46,17 @@
     margin-left: @component-padding / 2;
   }
 
+  .github-branch {
+    .icon.icon-git-branch {
+      line-height: 2.5;
+      vertical-align: top;
+
+      &::before {
+        vertical-align: text-bottom; // Visually adjusted
+      }
+    }
+  }
+
   .github-branch-detached {
     color: @text-color-subtle;
     font-style: italic;


Xapphire13:
In your screenshot it looks like it doesn't touch the right border, it should probably behave the same as the update reminder:

image

sethlu:
@Xapphire13 The screenshot's taken from an earlier version. I'll update the PR with some changes later.

sethlu:
New screenshots following recent changes are here:

Before:

screen shot 2017-11-06 at 6 41 31 pm

After:

screen shot 2017-11-06 at 6 42 02 pm

*And with baseline highlighted...*

Before:

before

After:

after

---

🚧  Do not merge this PR just yet, I just realized the layouts for a few other tiles have yet been investigated.

With the ones shown, a scrolling overflow is resolved.

sethlu:
Seems that the issue's already fixed in the latest release.

50Wliu:
@sethlu Hmm, still seems like a problem to me:
github-sunken

I think the issue is that that status bar tile is instantiated incorrectly.

sethlu:
Oh I just realized I’m running the dev version. 🙈

sethlu:
Just inspected the different views and think this patch should be ready to address the existing layout problem.

simurai:
Should we move this PR over to One light/dark?

Each theme might have different styling for the status bar items. For example, this PR would make it uneven in the `atom-light-ui` theme:

image

sethlu:
@simurai Good question... And I'm not sure 😕 since `atom-light-ui` overwrites the default styling over these tile items.

My original intention for this PR is to address an issue with the background color for the "21 files" part when mouse on hover (there's a gap between the top border of the status bar and the tile item). In a recent change with Electron the issue is *fixed* with an `overflow: hidden` on the status bar so the overflowing background won't cause a scrolling behavior.

I can probably modify PR so the changes only address this background fill overflow for the files changed section? ... as showed in comment https://github.com/atom/github/pull/1182#issuecomment-345091954

50Wliu:
I just looked at this diff and I think this PR is taking the wrong approach.  As I detailed in #1198 I think the root cause is that the DOM markup for the changed files count is incorrect, not the styling.

simurai:
> @50Wliu I think the root cause is that the DOM markup for the changed files count is incorrect

I tried this alternative approach by change the DOM markup: https://github.com/sethlu/github/pull/1. And yeah, it seems to fix the issue with the `1px` gap. 👍 

@sethlu If you like that approach, you can merge https://github.com/sethlu/github/pull/1 and it should update this one.

PR #1230, opened 2017-10-24
WIP centralize git operation side effects

kuychaco:
The purpose of this PR is to centralize all of the side effects associated with git operations, such as error-handling, showing notifications, prompting, and setting progress state. Previously this logic resided in controllers and views which coupled the side effect to a particular piece of the UI. This refactor ensures that a "push" will do the same thing regardless of whether it's triggered from the `PushPullMenuView` or from the GitHub panel as part of creating a PR.

The implementation includes the creation of an `ActionPipeline` for each git operation (push, pull, fetch, commit, checkout) followed by middleware registration for each side effect ('confirm-force-push', 'set-push-in-progress', etc). The pipelines are passed to the `Repository`, which executes the pipeline actions associated with each git operation.

Relies on https://github.com/atom/atom/pull/16294
Supersedes some of the work in #1138

TODO:
- [x] in progress spinner doesn't stop after error message appears
- [x] :fire: ModelStateRegistry and move amending/commit message state to repo
- [x] fix flicker when checking out existing branch (takes time or `currentBranch` to update)
- [x] add progress indicator for checkout
- [x] operationStates is not defaulting to empty object (set as defaultProp in the meantime)
- [x] convert remote interactions
- [x] convert checkout
- [x] convert commit
- [x] extract pipeline setup from `GithubPackage`
- [x] extract action to set in-progress state


lib/github-package.js
```diff
@@ -80,6 +83,110 @@ export default class GithubPackage {
...
+          console.error(error);
+          this.notificationManager.addError('Unable to push', {
+            description: `<pre>${error.stdErr}</pre>`,
+            dismissable: true,
+          });
+        }
+        return error;
+      }
+    });
+    pushPipeline.addMiddleware('set-push-in-progress', async (next, repository, branchName, options) => {
```
BinaryMuse: 👍

lib/models/repository-states/state.js
```diff
@@ -438,6 +449,11 @@ export default class State {
     return this.repository.state;
   }
 
+  // pipeline
+  executePipelineAction(...args) {
+    return this.repository.executePipelineAction(...args);
+  }
+
```
BinaryMuse: Does this need to be here on `State`?

lib/github-package.js
```diff
@@ -263,6 +370,7 @@ export default class GithubPackage {
 
     ReactDom.render(
       <RootController
+        pipelineManager={this.pipelineManager}
```
BinaryMuse: Since the pipelines live in the Repo now, we don't need to pass these down through the JSX, right?

kuychaco: oh yeah, good catch 👍

lib/models/repository-states/state.js
kuychaco: Yeah, it's either this, or in the `Present` class we call `this.repository.executePipelineAction(...)` directly. I decided to put a method on `State` for convenient access. Open to other thoughts if you have them.

test/controllers/git-tab-controller.test.js
```diff
@@ -105,30 +107,26 @@ describe('GitTabController', function() {
   });
 
   it('displays the staged changes since the parent commit when amending', async function() {
-    const didChangeAmending = sinon.spy();
     const workdirPath = await cloneRepository('multiple-commits');
     const repository = await buildRepository(workdirPath);
     const ensureGitTab = () => Promise.resolve(false);
+    // TODO: check that git tab opens in root controller when mrege conflict
```
smashwilson: Ooh I like that this is a thing we can do, now.

lib/github-package.js
```diff
@@ -415,9 +419,12 @@ export default class GithubPackage {
   @autobind
   async cloneRepositoryForProjectPath(remoteUrl, projectPath) {
     const context = this.contextPool.getContext(projectPath);
+    // TODO: do we need to pass pipelineManager? or is this discarded immediately & recreated later in workdir context
```
smashwilson: The `Repository` created here is indeed just discarded immediately; the "real" one will be created later by `getNextContext()`, either triggered by the `project.addPath()` call down below, or when you open a file in the new project.

But, I'd still give it a pipelineManager just the same. It makes it easier to reason about because we never have to worry about whether or not a given `Repository` has a manager or not.

lib/github-package.js
```diff
@@ -415,9 +419,12 @@ export default class GithubPackage {
...
     const repository = context.isPresent() ? context.getRepository() : new Repository(projectPath);
 
     await repository.clone(remoteUrl);
+    // TODO: ask Ash
+    // repository.destroy()
```
smashwilson: Hah, this is tricky because we only _need_ to call `.destroy()` here if the `: new Repository(projectPath)` branch is followed up above. Otherwise, we already had a `WorkdirContext` in the pool for the path we're trying to clone to, and it'll be destroyed by the pool when `getNextContext()` fires. We _can't_ destroy it if the `context.getRepository()` branch is followed and that Repository is in a state like Present, because we'd still need it for further operations in that git repo... but, if it's Present, the `.clone()` call should throw anyway. So I think this line is safe 🤞

lib/controllers/commit-view-controller.js
```diff
@@ -118,60 +105,54 @@ export default class CommitViewController {
...
-    const choice = this.props.confirm({
-      message: 'One or more text editors for the commit message are unsaved.',
-      detailedMessage: 'Do you want to commit and close all open commit message editors?',
-      buttons: ['Commit', 'Cancel'],
-    });
-    return choice === 0;
+  @autobind
+  async commit(message) {
+    // TODO: better way? can this logic just reside in the pipeline registry
+    if (this.getCommitMessageEditors().length > 0) {
```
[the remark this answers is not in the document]
kuychaco: Ooh yeah that's an interesting thought. I was more wondering how to best handle the fact that `commit` gets passed different arguments depending on if we're committing from the commit box or an editor. So it's either have the conditional logic here, or put it elsewhere and pass both the file path and message on from here. It's a small sort of nit-picky thing though. At least having the conditional here, we're very explicit around what argument will be used.

lib/get-repo-pipeline-manager.js
```diff
@@ -0,0 +1,204 @@
...
+
+  const fetchPipeline = pipelineManager.getPipeline(pipelineManager.actionKeys.FETCH);
+  fetchPipeline.addMiddleware('set-fetch-in-progress', async (next, repository) => {
+    repository.getOperationStates().setFetchInProgress(true);
+    try {
+      await next();
+    } finally {
+      repository.getOperationStates().setFetchInProgress(false);
+    }
+  });
```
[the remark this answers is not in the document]
kuychaco: Yeah, it would make things much less repetitive and I do like "clever". My small concern is that would make it harder to grep through the code to find all places where `setThingInProgress` is used. And if this file gets separated into multiple files for each operation (say after it grows a bunch), then we'd have to pull in `createInProgressMiddleware` from a different file and it's less immediately clear from glancing at the code what is happening.

lib/models/repository-states/state.js
```diff
@@ -421,6 +421,43 @@ export default class State {
     return null;
   }
 
+  // Atom repo state
+
+  @shouldDelegate
+  getOperationStates() {
+    return null;
```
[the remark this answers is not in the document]
kuychaco: ooh yeah 👍

lib/get-repo-pipeline-manager.js
```diff
@@ -0,0 +1,204 @@
...
+            ' Try pulling before pushing again. Or, to force push, hold `cmd` or `ctrl` while clicking.',
+          dismissable: true,
+        });
+      } else {
+        notificationManager.addError('Unable to push', {
+          description: `<pre>${error.stdErr}</pre>`,
+          dismissable: true,
+        });
+      }
+      throw error;
```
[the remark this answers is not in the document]
kuychaco: Yes exactly, to propagate to the rest of the middleware so they can act accordingly. And in the case of `checkout`, the BranchMenuView relies on catching an error to restore controls (we disable the UI when checkout is in progress). 

I'm leaning towards adopting a pattern of having pipeline middleware always re-throw caught errors and having the views be responsible for making the final determination of what to do with them. 

```
  pull() {
    try {
      return this.props.pull();
    } catch (error) {
      if (error instanceof GitError) {
        console.warn('Non-fatal', error);
      } else {
        throw error;
      }
      return null;
    }
  }
```

test/controllers/git-tab-controller.test.js
Arcanemagus: Also `mrege` -> `merge`.

lib/controllers/git-tab-controller.js
```diff
@@ -269,26 +267,8 @@ export default class GitTabController {
...
-        throw e;
-      }
-    }
-  }
-
-  @autobind
-  setAmending(isAmending) {
-    this.props.didChangeAmending(isAmending);
+  commit(message) {
+    return this.getActiveRepository().commit(message, {amend: this.isAmending()});
```
BinaryMuse: Since `this.isAmending()` calls out to `this.getActiveRepository()` itself, could we move the boolean into the `Repository#commit` method instead of passing it in as an option?

lib/views/branch-menu-view.js
```diff
@@ -79,73 +94,78 @@ export default class BranchMenuView extends React.Component {
...
+    if (this.editorElement) {
+      this.editorElement.getModel().component.setInputEnabled(false);
+      this.editorElement.classList.remove('is-focused');
+    }
+  }
+
+  @autobind
+  restoreControls() {
+    if (this.editorElement) {
+      this.editorElement.getModel().component.setInputEnabled(true);
```
BinaryMuse: Is this a public API? Can we count on it remaining stable?

ungb:
I went through and tested the following on mac(I'll test on linux and windows this week):
-Commit new changes
- Push
- Force Push
- Pull 
- Fetch 
- Create new branch and push 
- Switching branches

kuychaco:
Quick status update: This PR is temporary blocked on https://github.com/atom/atom/pull/16294 (adding a text editor `readonly` state). Currently waiting on input.

PR #1238, opened 2017-10-30
Never lock the index

### LINKED ISSUE #961: Disable locking (observe only by default)
I prefer doing my commits and pushes from the terminal, because for instance I can use my template for commits in terminal, where the plugin just adds simple commit messages. But when the plugin is running, even if I only use it to observe, or not visible at all, it creates a .lock file in the .git folder which prevents me from doing anything in the commandline. When I remove the lockfile, it gets recreated by the plugin. Since Atom now installs this plugin by default this is a problem. Simple solution for now was to disable the plugin completely, because closing atom for a commit doesn't work for me. A good solution would be to be able to observe by default and only when you want to use its staging, committing, pushing, whatever feature that requires that lockfile, that you actually need to tell the plugin you want to use it. I don't want to be forecd to soly use this plugin. I don't mind it being installed and running in the background and use it for simple commits.

BinaryMuse:
git/git@27344d6a6c8056664966e11acf674e5da6dd7ee3 has finally landed in Git 2.15.0 and in Dugite 1.49.0! This updates dependencies as appropriate, utilizes the new option via the `GIT_OPTIONAL_LOCKS` environment variable, and removes `DeferredCallbackQueue` (as it was introduced to help alleviate index lock contention in #726).

I verified this on my own computer using this test plan:

1. Open a project in Atom with `atom -d .`
2. Set Atom as my commit editor with `export EDITOR="atom -d --wait"`
3. Begin an interactive rebase with `git rebase -i head~5`
4. In the Atom window that opens, change the first and third item to `r` to rewrite the commit message.
5. Save and close the window.
6. In the window that next opens (to allow editing the commit), change the commit text, save, and close the window.

Prior to this PR, the rebase would immediately stop, because the window opened in step 1 would receive focus and run `status`, locking the index. After this PR, I was able to edit multiple commit messages with no problems. 🎉 

Fixes #961 
Fixes akonwi/git-plus#693
Ref platformio/platformio-atom-ide-terminal#297 (may fix some reported issues)


BinaryMuse:
Well, I was able to design a test that failed on my machine with the env var unset, but it caused a lot of issues on CI so I've decided to remove it — this is not likely to regress.

PR #1252, opened 2017-11-15
Move undo discard button & prompt if head has changed since last discard

kuychaco:
Refer to comment thread in #1242


kuychaco:
Superseded by #1702

PR #1254, opened 2017-11-17
Disable all relevant color options when executing Git commands

### LINKED ISSUE #977: Git config `ui = always` breaks diff pane
I just updated atom to 1.18.0 in osx and started looking at the github package. The staging pane works fine, but when clicking on one of the modified files nothing happens.
I just get this error on the console:

`Uncaught (in promise) peg$SyntaxError {message: "Expected "* Unmerged path ", "diff ", "diff --cc ", or end of input but "\x1B" found.", expected: Array[4], found: "", location: Object, name: "SyntaxError"}`

This is my atom version:
Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0

I tried different repos and safe mode but the error always happens

BinaryMuse:
In lieu of an environment variable or some other global `--no-color` option, this PR adds command line arguments to every Git command that sets all relevant color options to `false` (which is equivalent to `never` for options that take it).

Fixes #977

### DIFF
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -41,6 +41,13 @@ export class LargeRepoError extends Error {
   }
 }
 
+const DISABLE_COLOR_FLAGS = [
+  'branch', 'diff', 'showBranch', 'status', 'ui',
+].reduce((acc, type) => {
+  acc.unshift('-c', `color.${type}=false`);
+  return acc;
+}, []);
+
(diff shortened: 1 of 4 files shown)


BinaryMuse:
An alternative approach would be to find all the invocations of the specific commands that take a `--no-color` option (like `status` and `diff`) and add the arguments there, but this strategy makes sure we don't miss any. Open to additional 💭 s.

PR #1274, opened 2018-01-04
Fix #1067: Bypass external diff tools

### LINKED ISSUE #1067: External app opens if external diff tool is defined in .gitconfig
### Description

Atom opens the external application for git diff, if defined in repository `.git/config` or global `.gitconfig`. In my opinion it should always use the built-in viewer editor. Reason I have an external tool configured is because when using command line I prefer to view diff using [Meld](http://meldmerge.org/). There's no reason to do it if already on a GUI (i.e. Atom).

This is possible in command-line using `git diff --no-ext-diff`, but I don't know how Atom handles it, if it's possible to skip git configuration from it.

### Steps to Reproduce

1. [Configure an external diff tool](https://stackoverflow.com/questions/255202/how-do-i-view-git-diff-output-with-a-visual-diff-program) in `~/.gitconfig` or `.git/config`
2. Open a git repo in Atom
3. Change something
4. Click a file in the *Unstaged changes* panel
5. The external diff tool will be opened

If you want to reproduce **more quickly** without the hassle of configuring an external tool, you can specify any invalid command as diff tool. The command will fail, but Atom will not show the diff anyway. Just replace step one with:

1. `echo -e "\n[diff]\n\texternal = missingno" >> ~/.gitconfig`

### Versions

   Ubuntu 16.04.2 LTS

    Atom    : 1.18.0
    Electron: 1.3.15
    Chrome  : 52.0.2743.82
    Node    : 6.5.0

    apm  1.18.1
    npm  3.10.10
    node 6.9.5 x64
    python 2.7.12
    git 2.7.4

daviwil:
This change fixes issue #1067 which states that a user-configured external diff tool will cause the "View [Un]Staged Changes" commands to return empty diff views because `git diff` launches an external viewer.  The fix is to add the `--no-ext-diff` parameter to the `git diff` call in `getDiffForFilePath` as this bypasses any user-configured external diff tool.

I double-checked the `diffFileStatus` path since that's another call to `git diff`, it does not seem to be affected by user configuration.

Since this is my first PR to this repo, please let me know if I'm not following the guidelines correctly!  :smile: 
 
Fixes #1067

### DIFF
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -461,7 +461,7 @@ export default class GitShellOutStrategy {
   }
 
   async getDiffForFilePath(filePath, {staged, baseCommit} = {}) {
-    let args = ['diff', '--no-prefix', '--no-renames', '--diff-filter=u'];
+    let args = ['diff', '--no-prefix', '--no-ext-diff', '--no-renames', '--diff-filter=u'];
     if (staged) { args.push('--staged'); }
     if (baseCommit) { args.push(baseCommit); }
     args = args.concat(['--', toGitPathSep(filePath)]);
--- a/test/git-strategies.test.js
+++ b/test/git-strategies.test.js
@@ -203,6 +203,18 @@ import {fsStat, normalizeGitHelperPath, writeFile, getTempDir} from '../lib/help
         assert.isUndefined(diffOutput);
       });
 
+      it('bypasses external diff tools', async function() {
+        const workingDirPath = await cloneRepository('three-files');
+        const git = createTestStrategy(workingDirPath);
+
+        fs.writeFileSync(path.join(workingDirPath, 'a.txt'), 'qux\nfoo\nbar\n', 'utf8');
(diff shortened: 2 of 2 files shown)


BinaryMuse:
@daviwil I lied! Mind running `npm run lint` and fixing the errors that pop up? Thanks!!

daviwil:
Whoops!  Should have thought of that :)  Updated the PR, thanks a bunch!

daviwil:
Hmm, looks like the macOS build timed out.  Is that a common occurrence or should I investigate?

daviwil:
My mistake, the timeout was in the Linux build which uses Atom Beta.  I noticed a similar build timeout [here](https://travis-ci.org/atom/github/jobs/313786361) so it seems like a recurrent issue.

BinaryMuse:
@daviwil Yeah, it's a weird flaky test thing we haven't tracked down. I'll keep an eye on the tests and merge when green. Thanks again!! ❤️

Arcanemagus:
> it's a weird flaky test thing

I see it rarely on several different repos, I think it's a bug with either Travis or Atom itself personally.

PR #1275, opened 2018-01-05
Handle filemode changes

### LINKED ISSUE #310: Add support for `typechange`
If a symlink changes to a regular file or a regular file changes to a symlink, Git reports the change as a `typechange` and `git diff --name-status head` shows a `T`, which we do not currently handle.

### LINKED ISSUE #942: Staging mode changes on macOS impossible from "Unstaged Changes" view.
### Description
:wave: Hey! I just noticed something interesting when trying to stage a file using the "Stage File" button in the "Unstaged Changes" diff view: when I change the file's mode from `644` to `755`, Atom appears to stage the changes to its contents but not the changes to the mode. Right-clicking the file and clicking "Stage" works, and clicking "Stage All" works, but the diff view leaves the file in a half-staged netherworld. `git diff` from the command line shows the unstaged mode change.

git

### Steps to Reproduce

1. Create a file (e.g., `test.js`).
2. Change the file's mode from `644` to `755`.
3. Open the file in the Git package's diff view.
4. Click "Stage File."

**Expected behavior:** I would expect the whole file to be staged, including changes to its metadata.

**Actual behavior:** Only the contents of the file are staged.

**Reproduces how often:** 100% of the time.

### Versions
```
~ ❯❯❯ atom --version
Atom    : 1.18.0
Electron: 1.3.15
Chrome  : 52.0.2743.82
Node    : 6.5.0

~ ❯❯❯ apm --version
apm  1.18.1
npm  3.10.10
node 6.9.5 x64
python 2.7.13
git 2.13.1

~ ❯❯❯ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.12.5
BuildVersion:	16F73
```

### Additional Information
💖  Thanks for taking a look.

### LINKED ISSUE #1273: Issues staging changes to execute permissions
The GitHub package chokes on files that have filemodes changed from 100644 to 100755 or vise versa. Repro steps:

**Scenario 1**: only changing file mode

1. Change a non-executable file in a repo to be executable: `chmod +x somefile.txt`
2. Click on the file in Atom

Atom throws an exception in the diff parser (`what-the-diff`). It *is* possible to stage the file by pressing enter on the file or right-clicking and choosing "Stage."

**Scenario 2**: text and file mode changes

1. Change a non-executable file in a repo to be executable: `chmod +x somefile.txt`
2. Change the contents of the file
3. Click on the file in Atom

In this case, Atom shows the text changes and allows the user to stage the lines/hunks effected, but once they're all staged, the same issue as the scenario above appears.

---

Pretty sure this is a duplicate but I can't seem to find the original issue. `¯\(º_o)/¯`

BinaryMuse:
This PR introduces necessary changes to handle mode changes, including handling symlinks.

- [x] Clean up file patch controller tests after refactor, add comments
- [x] UX improvements
- [x] Test symlink cases
- [x] Handle parsing diffs with no patch or binary part
- [x] Header in normal diff UI when execute bit is changed
- [x] Stage entire file instead of all lines/hunks when clicking "Stage File" from header
- [x] Use keyboard to stage mode changes
- [x] Special UI when symlink changes path
- Handle typechanges:
  - [x] Properly handle single diff case when symlink added/removed
  - [x] Properly parse/handle dual diffs
  - [x] Show header when file changes to symlink or visa versa
  - [x] Handle staging symlink mode change
  - [x] Add command for staging/unstaging symlink changes
  - [x] When unstaging contents for typechange, delete file and vice versa
  - [x] Allow staging lines and hunks for symlink, without staging symlink change
  - [x] Correctly handle staging lines that require symlink change to be staged
  - [x] Check permanent pane item behavior

Fixes #1273 
Fixes #310 
Fixes #942


BinaryMuse:
For staging *only* a filemode change (between `100644` and `100755`), we've found two strategies:

```
$ git update-index --skip-worktree --chmod=+x <file>
$ git update-index --no-skip-worktree <file>
```

and

```
$ git update-index --cacheinfo new_mode,blob_oid,filename
```

(which requires finding the current blob OID on the index via `git ls-files -s -- filename`)

kuychaco:
Notes for implementing logic for staging/unstaging symlink changes:

‘Stage/Unstage Symlink Change` button should stage/unstage whole file UNLESS
* [unstaged changes] symlink deleted and added lines
    * instead use `git rm --cached <filename>`
* [staged changes] symlink created and deleted lines
    * instead use `git rm --cached <filename>`

Edge case: for symlink change and added/changed lines, when all lines are staged/unstaged there’s no way to stage/unstage the file deletion. Consider adding “Delete file” button instead of “File has no contents” message?

lib/git-shell-out-strategy.js
```diff
@@ -487,12 +508,26 @@ export default class GitShellOutStrategy {
       // add untracked file
       const absPath = path.join(this.workingDir, filePath);
       const executable = await isFileExecutable(absPath);
+      const symlink = await isFileSymlink(absPath);
```
BinaryMuse: wondering if this will return the correct thing on Windows (e.g. could Git be configured to recognize them or have a stored symlink filemode but the fs helpers return `false`)

lib/controllers/file-patch-controller.js
```diff
@@ -172,8 +172,12 @@ export default class FilePatchController extends React.Component {
       {oldMode: fp.getOldMode(), newMode: fp.getNewMode()} :
       null;
     const symlinkChange = fp && fp.hasSymlink() ?
-      {oldSymlink: fp.getOldFile().getSymlink(), newSymlink: fp.getNewFile().getSymlink(), status: fp.getStatus()} :
-      null;
+    {
+      oldSymlink: fp.getOldFile().getSymlink(),
+      newSymlink: fp.getNewFile().getSymlink(),
```
BinaryMuse: Can we throw these on `FilePatch` directly, similar to `getOldPath` and `getNewPath`?

lib/models/file-patch.js
```diff
@@ -208,7 +225,16 @@ export default class FilePatch {
...
+      if (this.hasTypechange() && this.getStatus() === 'added') {
+        // handle special case when a file was created after a symlink was deleted.
+        // In order to unstage the file creation, we must ensure that the unstage patch has no new file,
+        // so when the patch is applied to the index, there file will be removed from the index
+        return this.getUnstagePatch().clone({
+          newFile: new File({path: null, mode: null, symlink: null}),
+        });
+      } else {
+        return this.getUnstagePatch();
+      }
```
BinaryMuse: This look great! If they're not exercised in a repo test, we should consider adding specific `FilePatch` tests for them.

test/fixtures/repo-symlinks/dot-git/hooks/applypatch-msg.sample
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
BinaryMuse: How do you feel about dropping all the sample hooks and other non-essential files from the fixture repo, just to keep things tidy?

lib/models/file-patch.js
kuychaco: Done!

BinaryMuse:
@simurai Hi! 👋 We were wondering if we could get some of your lovely design expertise on this PR. There are a few new interactions we want to have some UI for.

**Executable mode changes**

When a user changes a file to make it executable or non-executable, we want the user to be able to stage this separately from any lines that changed in the file. Currently, there is a banner under the header affording this interaction.

unstaged_changes__thing_txt_ ___github_test-project

I also wonder if we could say something more user-friendly than just the textual representation of the modes (`100644` and `100755`). Maybe "This file was made executable," etc?

**Adding, removing, or changing a symlink**

If the user adds a new file that is a symlink, deletes a file that is a symlink, or changes a symlink so that it points to a new file, then that change will have no associated hunks ever. Right now we still just show a banner, but the diff view feels a little empty, and the primary change to the file (adding/removing/changing a symlink) feels de-emphasized.

git_ ___github_test-project

unstaged_changes__symlink-changed_txt_ ___github_test-project

I considered making a larger, more front-and-center UI for this type of interaction, but wasn't quite sure what direction to take it.

**Changing a regular file to a symlink or vice versa**

When a file is deleted and a symlink added in its place, or a symlink is deleted and a file added in its place, we then have a change that includes *both* a symlink being added/removed *and* lines being added/removed.

In this case, the different types of changes can be staged independently (excluding certain edge cases, which the package handles).

unstaged_changes__symlink-removed-and-contents-added_txt_ ___github_test-project

---

If you have 💭 s here, they'd be much appreciated. Happy to help turn design vision into actual code if you so desire. Thank you very much! 🙇‍♀️

simurai:
> I also wonder if we could say something more user-friendly than just the textual representation of the modes (100644 and 100755). Maybe "This file was made executable," etc?

Yeah, that would be nice so people don't have to google the meaning. Maybe still keep the number? Like:

File changed mode from non-executable (`100644`) to executable (`100755`)

> I considered making a larger, more front-and-center UI for this type of interaction, but wasn't quite sure what direction to take it.

I think it's still good to keep all these "meta changes" (if you can call it that) consistent. But given that on average, people see these not that often, making the whole area for the "meta section" a bit bigger might be ok. Like it could have a title and a different background. Also, have red/green highlights to show the diff? I'll see if I can tweak it a bit.

---

Another question? Should people be able to discard mode and symlink changes? Maybe tricky in some cases? Like discarding the "Stage Symlink Change" here...

unstaged_changes__symlink-removed-and-contents-added_txt_ ___github_test-project

... would result in adding the symlink back, but then also having another file with the same name. Right? Hmm... 🤔  Maybe too rare of a use-case to bother, for now.

BinaryMuse:
@simurai Ahhh, it looks so good! 😍 Calls out the change without being too invasive, I love it.

BinaryMuse:
@kuychaco As far as I can tell, the last of your changes look good!

PR #1276, opened 2018-01-06
Improve wording for empty diff message, specify no "changed" contents

kuychaco:
This is so as not to confuse users who might interpret the old message as saying the file is empty of contents (ie. a blank file)
  
Previously:
```File has no contents```

Now:
```File has no changed contents```

### CODE REVIEW

**Comment by kuychaco** [comment]
Or maybe "modified" would be a better 
```File has no modified contents```

**Comment by 50Wliu** [comment]
`File has not been modified`?

**Comment by BinaryMuse** [comment]
> File has not been modified

@50Wliu That one isn't strictly true, especially now that we're handling filemode changes near the same diff UI.

File: lib/views/file-patch-view.js [hunk 1]
```diff
@@ -76,7 +76,7 @@ export default class FilePatchView extends React.Component {
   renderEmptyDiffMessage() {
     return (
       <div className="is-blank">
-        <span className="icon icon-info">File has no contents</span>
+        <span className="icon icon-info">File has no changed contents</span>
```
BinaryMuse: How do you feel about "No changes to display" or similar? It also works well if and when we decide to handle ignoring whitespace.

kuychaco:
Closing this as it is soon to be obsoleted by the mighty #1512

PR #1284, opened 2018-01-16
Fix incorrect file paths in transpiled files

BinaryMuse:
This PR updates `atom-babel6-transpiler` to 1.1.3, which fixes an issue with relative filenames during transpilation. As a result, clicking links in the `atom-mocha-test-runner` output will now lead to the correct file.

etch-wrapper_test_js_ ___github_github_and_github_package_tests_-_pid_16952_and_1__mtilley_c02r312bg8wm____github_github__zsh_

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   ],
   "dependencies": {
     "@binarymuse/relay-compiler": "^1.2.0",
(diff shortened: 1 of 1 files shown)


BinaryMuse:
Travis is having a sad day, so I'm merging without it being green.

Issue #1119, opened 2017-08-18
Stalled tests at GithubPackage / activate / uses models from preexisting projects


smashwilson:
Seen [on Travis](https://travis-ci.org/atom/github/jobs/266112361):

```
    activate()
      ✓ begins with an undetermined repository context
      ✓ uses models from preexisting projects (186ms)
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

PR #1307, opened 2018-02-09
Align changed file count in status bar

### LINKED ISSUE #1198: Incorrect styles on github-ChangedFilesCount causes status bar tile to appear unaligned

50Wliu:
No more misalignment, yay!
status-bar-file-count

Same as sethlu/github#1 by @simurai but on master.

Fixes #1198
Supersedes and closes #1182

### DIFF
--- a/lib/views/changed-files-count-view.js
+++ b/lib/views/changed-files-count-view.js
@@ -23,8 +23,9 @@ export default class ChangedFilesCountView extends React.Component {
     return (
       <a
         ref="changedFiles"
-        className="github-ChangedFilesCount inline-block icon icon-diff"
+        className="github-ChangedFilesCount inline-block"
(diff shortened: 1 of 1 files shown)


simurai:
Ok, let's get this 🚢 ed. Thanks @50Wliu 

Failed test seems to be #1119.

Issue #842, opened 2017-05-17
Auto-fetch?


danielbayley:
I recently [implimented](https://github.com/akonwi/git-plus/pull/635) this in _git-plus_, but wondered if it might be an appropriate feature here?

Tower and GitKraken clients both auto-fetch periodically.

adrian3H:
Also - clicking 'fetch' doesn't retrieve full list of branches in repository (origin). Currently there doesn't appear to be a way to check out a branch via the github package in Atom.

simurai:
Here some thoughts about enabling auto-fetching. They're mostly questions. :wink:

### How often?

- `git-plus` has an option to choose between 0-60 minutes. 0 means disabling auto-fetching
- GitHub Desktop seems to auto-fetch every 5-10 minutes???

### Auto-fetch when doing certain actions?

Auto-fetching could also be tied to doing certain actions. For example:

- After opening a repo (or adding a repo to a current project).
- After switching branches.
- After committing.
- After a computer comes back from sleep. If that's possible to detect? Otherwise maybe after being idle for a while. Like when coming back from lunch or the next morning.
- More?

Because in these situations, it would be a good time to pull (if you're behind). Periodically auto-fetch could be distracting if you're in the middle of coding and you see the spinner and then the counter update.

![image](https://user-images.githubusercontent.com/378023/38079026-fe4954a0-3378-11e8-8ac8-821cde62945d.png)

And then you're in this conflict: "Should I first finish and commit or go check on github.com what the changes are?" :thinking:

### Option to disable?

Have an option to disabling auto-fetching in case you are tethering with your phone and wanna save data. https://github.com/atom/github/pull/1308#issuecomment-377095292. There are also apps like [TripMode](https://www.https://www.tripmode.ch/.ch/) to control access outside of Atom.

mattmattmatt:
@simurai I think all of these are good points in time to auto-fetch without being distracting. 👍 I personally like a periodic auto-fetch as well, but I can definitely see that mental conflict once one realizes that "there are changes on remote", so that might not be for everyone.

From a quick web search, it seems to be [possible in Electron to detect when a computer comes back from sleep](https://github.com/electron/electron/blob/master/docs/api/power-monitor.md). I have no idea how hard it would be to get such information into the GitHub package, but in theory this possibility seems to exist.

I threw together a quick hack of what a __⚙️Settings__ sub-screen could look like:

[...]

Issue #311, opened 2016-11-10
Simplify Pull/Push tooltip


simurai:
## Problem

Now that the branch pull/push buttons are in a tooltip of the status-bar, you have to click 2 times (click to open tooltip + click button) to do these actions.

![1](https://cloud.githubusercontent.com/assets/378023/20167938/3be5ac0e-a762-11e6-99c1-7a1f67032223.png)

It just feels a bit like too much extra work.

## Possible solution

Below an exploration in adding something similar to the [Magic sync button](https://github.com/desktop/desktop/issues/598) like on Desktop.

Clicking on the status-bar items does the action immediately without showing a tooltip. Also only the actions that you can currently do are shown.

### Neither ahead or behind
![4](https://cloud.githubusercontent.com/assets/378023/20167967/65ecf7f0-a762-11e6-9bea-dee2083235d6.png)

### Behind
![2](https://cloud.githubusercontent.com/assets/378023/20167946/4ac0d906-a762-11e6-88a3-fa3ad3c04754.png)

### Ahead
![3](https://cloud.githubusercontent.com/assets/378023/20167953/531699ce-a762-11e6-90f2-4cee7728f283.png)

### Ahead and behind
![6](https://cloud.githubusercontent.com/assets/378023/20167960/5a61f26e-a762-11e6-89ab-2effa10c84a8.png)

### Unpublished
![5](https://cloud.githubusercontent.com/assets/378023/20167977/6e2ad87e-a762-11e6-9c44-fce56dc3b327.png)

This should cover 97.4% of use cases.

## Special cases

Sometimes you might wanna fetch before pulling. Maybe to make sure nobody made more changes on the remote? Or you want to force push. For these cases, we could have a context-menu show up when right-clicking. So at any time there are still available.

![group](https://user-images.githubusercontent.com/378023/33748236-b46ec8bc-dc0a-11e7-86b9-77d96860b2c8.png)

The current popup could still be used for errors or confirmations, for example:

![7](https://cloud.githubusercontent.com/assets/378023/20167987/83d92c70-a762-11e6-8eb2-29b231b2017b.png)

kuychaco:
For the "Special cases" we could additionally offer alt/cmd-clicking for fetching and force pushing to save users some clicks, per @mattmattmatt's suggestion in https://github.com/atom/github/pull/1308#issuecomment-365547324

[...]

PR #1308, opened 2018-02-11
Push and pull straight from status bar

mattmattmatt:
### Description of the Change

Instead of clicking on the PushPull icon in the status bar to bring up a tooltip that requires me to click again to actually push or pull, this PR proposes a change so that pushing and pulling is possible directly from the status bar.

A GIF is worth a thousand character description:

This PR is not in a mergeable state yet – before handling edge cases and unhappy paths, and making tests pass, I would like to see if this change is something the maintainers would be open to in the first place.

### Alternate Designs

An alternative to this solution could be to keep the PushPull tooltip as is but additionally enable users to push and pull straight from the status bar by e.g. shift-clicking one of the icons.

### Benefits

No more unnecessary clicking and mousing before being able to push and pull changes.

### Possible Drawbacks

Users used to the old behavior might accidentally push or pull. A "first time warning" for the new behavior could solve that.
Users might also not be used to perform "important" actions from the status bar.


ungb:
👋 @simurai any chance you can take a look at this?  Was curious what your thoughts were on this change?

simurai:
I agree that having to open the tooltip first is a bit of a chore. There is a similar proposal where the status-bar tile changes based on the situation: #311

Some questions about this PR that come to mind:

- How do you fetch? Maybe as an option in a right-click context menu. Or `alt`-clicking on the pull icon, but that's harder to discover.
- How do you force-push? Context menu or `alt`-clicking on the push icon?
- There are cases where a message gets shown. Like when no remote branch exists and pushing would create one. In that case the tooltip could still open or a notification could be used.

mattmattmatt:
I really like the ideas in #311, I would love to take some inspiration from them in this PR. Are these "up for grabs"?

Regarding your questions:
>  - How do you fetch? Maybe as an option in a right-click context menu. Or alt-clicking on the pull icon, but that's harder to discover.

I was imagining alt/cmd-clicking for fetching, and force pushing. Discoverability is certainly a concern there, so I was thinking of showing the blue Atom tooltips when hovering over the respective icons, similar to other items in the status bar:

image

I'm also open to other suggestions, of course.

> - How do you force-push? Context menu or alt-clicking on the push icon?

Same as fetching, an alt/cmd-click, with blue tooltip. As is the case now, there would still be the "do you really want to do this?" prompt for confirmation.

> - There are cases where a message gets shown. Like when no remote branch exists and pushing would create one. In that case the tooltip could still open or a notification could be used.

I have not looked into the message cases in detail yet, but that's where I see the existing tooltip coming in, or in the more "error"-like cases the notifications.

I'd be happy to provide more screenshots of my ideas as I go through the full scope. For now I just wanted to hear if the maintainers are even interested in this change.

simurai:
> Are these "up for grabs"?

Yes, totally up for grabs. :grin: But before sinking too much time into it, might be good to see if there is more feedback from other maintainers. /cc @atom/feedback

kuychaco:
👍 to implementing @simurai's design in #311 

Thanks for the contributions so far @mattmattmatt. We're definitely interested in this change to make pushing/pulling easier and more discoverable. 

Our main concern with what is currently implemented in this PR is that it's not immediately clear that clicking the arrow icons in the status bar will result in a push/pull action, and we want to avoid unpleasant surprises for our users. @simurai's design addresses this nicely by explicitly stating what will happen. 

Regarding message cases, if I'm not mistaken, I _think_ this should already be handled for you in our [repo pipleline manager](https://github.com/atom/github/blob/master/lib/get-repo-pipeline-manager.js) where we've centralized all of the side effect-y things for git operations. If there's something missing, feel free to add it. For now we're simply using notifications and atom's built-in confirm prompt.

Thanks for the great help!

lee-dohm:
@mattmattmatt Hey, is there any help you need on this?

mattmattmatt:
@lee-dohm Thanks for the offer, I just needed some dedicated time to continue on this.

This PR from my point of view is now "feature complete" and it seems all tests are passing too. Where do we go from here? Anything I can help with to make reviewing easier?

lee-dohm:
We'll take a look at it in our next prioritization meeting @mattmattmatt and let you know :+1:

.eslintrc
```diff
@@ -14,7 +14,8 @@
   },
   "extends": ["fbjs/opensource"],
   "rules": {
-    "linebreak-style": 0
+    "linebreak-style": 0,
+    "indent": ["error", 2]
```
smashwilson: In the future, it'd be best not to mess with linter settings in the same PR as other work, unless it's directly needed to make new source format nicely (for example). I'm fine with this one though :smile:

lib/get-repo-pipeline-manager.js
```diff
@@ -46,7 +46,7 @@ export default function({confirm, notificationManager, workspace}) {
         if (/rejected[\s\S]*failed to push/.test(error.stdErr)) {
           notificationManager.addError('Push rejected', {
             description: 'The tip of your current branch is behind its remote counterpart.' +
-            ' Try pulling before pushing again. Or, to force push, hold `cmd` or `ctrl` while clicking.',
+            ' Try pulling before pushing.<br />To force push, hold `cmd` or `ctrl` while clicking.',
```
smashwilson: 👍

lib/views/push-pull-view.js
```diff
@@ -2,37 +2,196 @@ import React from 'react';
...
+
+  onClickPush = clickEvent => {
+    if (this.props.isSyncing) {
+      return;
+    }
+    this.props.push({
+      force: clickEvent.metaKey || clickEvent.ctrlKey,
+      setUpstream: !this.props.currentRemote.isPresent(),
+    });
+  }
```
smashwilson: Is there a reason these are:

```js
methodName = () => {
  //
}
```

Rather than:

```js
methodName() {
  //
}
```

? It isn't a pattern I've seen before.

lib/views/push-pull-view.js
```diff
@@ -2,37 +2,196 @@ import React from 'react';
...
+    this.setTooltip();
+  }
+
+  setTooltip = () => {
+    this.tooltip = atom.tooltips.add(this.tileNode, {
+      title: `<div style="text-align: left; line-height: 1.2em;">${this.tooltipTarget.dataset.tooltip}</div>`,
+      delay: {show: 200, hide: 100},
+      html: true,
+    });
+  }
```
smashwilson: 🚨 I'd prefer it if we used the [`<Tooltip>` component](https://github.com/atom/github/blob/master/lib/views/tooltip.js) in render methods rather than manipulating `atom.tooltip` directly in lifecycle methods. It helps centralize the imperative Atom API logic in well-defined, centralized components.

lib/views/push-pull-view.js
```diff
@@ -2,37 +2,196 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 
+import {RemotePropType, BranchPropType} from '../prop-types';
+
+
```
smashwilson: 🚨 🎨 Looks like this file has a few extra newlines in it. (I'm surprised the linter didn't complain, actually.)

lib/views/push-pull-view.js
```diff
@@ -2,37 +2,196 @@ import React from 'react';
...
-        <span className="github-PushPull-label is-pull">
-          {this.props.behindCount ? `${this.props.behindCount}` : ''}
+    const isAhead = this.props.aheadCount > 0;
+    const isBehind = this.props.behindCount > 0;
+    const isUnpublished = !this.props.currentRemote.isPresent();
+    const isDetached = !!this.props.currentBranch.detached;
+    const isSyncing = this.props.isSyncing;
+    const hasOrigin = !!this.props.originExists;
+
+    return (<div
```
smashwilson: 🚨 :art: We prefer this style for multi-line jsx, with parenthesis on their own lines:

```js
return (
  <div
     ref={...}
     className={...}>
  </div>
)
```

menus/git.cson
```diff
@@ -115,3 +115,24 @@
...
+        'type': 'separator'
+    }
+    {
+        'label': 'Push',
+        'command': 'github:push'
+    }
+    {
+        'label': 'Force Push',
+        'command': 'github:force-push'
+    }
```
smashwilson: 👍 ✨

lib/views/push-pull-view.js
```diff
@@ -2,37 +2,196 @@ import React from 'react';
...
+      ref={node => (this.tileNode = node)}
+      className={cx('github-PushPull', 'inline-block', {'github-branch-detached': isDetached})}>
+
+      {isAhead && !isBehind && !isUnpublished && (
+        <a
+          ref={node => (this.tooltipTarget = node)}
+          onClick={this.onClickPush}
+          className="push-pull-target"
+          data-tooltip="Click to push<br />Cmd-click to force push<br />Right-click for more">
```
smashwilson: 🚨 Force-push is only "cmd-click" on macOS. Can we vary the caption to match the user's platform?

lib/views/push-pull-menu-view.js
```diff
@@ -1,148 +0,0 @@
...
-    } catch (error) {
-      if (error instanceof GitError) {
-        // eslint-disable-next-line no-console
-        console.warn('Non-fatal', error);
-      } else {
-        throw error;
-      }
-    }
-  }
-}
```
smashwilson: Deleted code makes me happy 😆

simurai:
One small thing. When running a command different to the one that is shown, should the text change while in progress? For example if I'm one commit ahead, but instead of pushing, I fetch:

fetch

_While_ the spinner rotates, should it say "Fetch" instead of "Push 1"? So that there are no panic attacks. :astonished:  "Wait wait.. did I just click the wrong thing? Is it now pushing?".

lib/views/push-pull-view.js
mattmattmatt: Absolutely! 👍

lib/views/push-pull-view.js
mattmattmatt: Since arrow functions don't change their context, `this` inside of these event handlers automatically refers to the component. So this pattern avoids having to do e.g. 
```JS
function constructor(props) {
    super(props);
    this.onClickPush = this.onClickPush.bind(this);
}
```
or 
```JSX
<a onClick={this.onClick.bind(this)}>Click me</a>
```
to keep the correct context in event handlers.

If introducing this new pattern is unwelcome, I'll be happy to replace it with on of the more traditional approaches.

mattmattmatt:
Thanks for the feedback, I'll implement it ASAP ⏰ 

Your comment about the status bar tile saying "Push" while it's fetching is also a good point, I'll provide a better experience for that. Avoiding user panic is a high priority 😉

mattmattmatt:
@smashwilson I was also considering whether to include auto-fetching for published branches every X minutes in this PR. Since fetches are non-destructive and cheap (if only fetching the current branch), it would be cool if that status bar tile always showed what's available on the remote.

What do you think, is this a good idea? And if yes, should I include it here or in a separate branch off this one?

lee-dohm:
> I was also considering whether to include auto-fetching for published branches every X minutes in this PR.

@mattmattmatt While they may be non-destructive, they may not be cheap for people that have to pay per MB for their network connection. Additionally, I think that this should be discussed as a separate change to the one you're making currently.

lib/views/push-pull-view.js
smashwilson: Oh! Neat. 👍 Elsewhere we're using the `@autobind` decorator, but honestly I kind of prefer this - it doesn't rely on is-it-or-isn't-it-deprecated Babel plugins :smile:

simurai:
> @mattmattmatt ...include auto-fetching.. ...is this a good idea?

Yes, that would be a great improvement. 💯 

> @lee-dohm I think that this should be discussed as a separate change to the one you're making currently.

I agree to leave it for another PR.

Here a relevant issue: https://github.com/atom/github/issues/842. I'll try to add some thoughts around auto-fetching there.

mattmattmatt:
@smashwilson  🏓 I addressed your feedback (hopefully without missing anything) and refactored this quite a bit. The separate `<Tooltip />` component did make life a little bit harder but I do like the idea of having the tooltip component right next to the element it describes. 👍
Let me know if anything else (or anything new) needs addressing.

@simurai 🚂 I got rid of the panic inducing in-progress state (good catch!) and while I was in there, I took the train to animation station a bit further: I included these small animation details in this PR that I'm quite fond of, but I'd like to hear your opinion.

smashwilson:
✨ Thanks for all your patience and responsiveness on this!

> The separate <Tooltip /> component did make life a little bit harder but I do like the idea of having the tooltip component right next to the element it describes. 👍

Oops, sorry about that. 🙈Usually, moving to `<Tooltip />` involves getting rid of a bunch of component lifecycle methods. Nicely done with it though!

> I included these small animation details in this PR that I'm quite fond of, but I'd like to hear your opinion.

Let's merge this with them there. We can always back them out if they turn out to be too much as we're using it 😄

simurai:
> @mattmattmatt I took the train to animation station a bit further

🚂 Hehe.. awesome! :clap:

varemenos:
Sorry for leaving a comment that doesn't contribute to the issue...
I just wanted to say thank you for spending so much time and effort into the editor's interfrace and UX.

PR #1311, opened 2018-02-15
Fix logic for automatically opening diff views based on selection change

### LINKED ISSUE #1287: Unstaged changes view pops up unexpectedly when switching panes
### Description

Sometimes, when changing focus from a diff view in one pane to a text editor in another pane, another diff view unexpectedly pops up.

### Steps to Reproduce

1. Add multiple project folders, each of which are a git repository (say `atom` and `github`).
2. Open a diff view for a file in the `atom` repository.
3. Change focus to another pane in which a file from `github` is open.
4. A diff view for the first modified file in the `github` repository opens unexpectedly.

diff-view

**Expected behavior:**

It seems like diff views should only open when I directly perform some action in the commit panel. They shouldn't open in response to other unrelated events like changing focus to a text editor.

**Actual behavior:**

Diff views spontaneously open when changing focus to a pane containing a text editor.

**Reproduces how often:**

Unfortunately, this seems to depend on some sequence of events that I don't fully understand. I can't always get it to happen.

### Versions

```
Atom    : 1.25.0-dev-3d8902d52
Electron: 1.7.10
Chrome  : 58.0.3029.110
Node    : 7.9.0
```

kuychaco:
Fixes #1287 

Problem: We’re seeing surprised diff views popping up unexpectedly, such as in the following scenarios

Scenario 1: see steps in #1287 
Scenario 2: have two repos open in project. have pending diff in pane 2 corresponding to repo 2. in pane 1 focus a file from repo 1, then focus a file from repo 2. note that new diff view opens. 
Scenario 3: only have one repo. have pending diff in pane 2. focus pane 1. focus staging view (be sure not to click on a file). use arrow key to change selection (up or down). note that new diff view pops up in pane 1. The only time we want to open a brand new pending diff view in a pane is if the user explicitly expresses intent, such as by clicking the file (see 1 below). 

These point to bugs in our logic for automatically opening diff views when the selection changes. We want to do this in two cases:

1. when navigating up/down and there is a pending diff view open in the active pane, update the diff view to reflect the new selection. This enables users to quickly glance at the change they've made
2. when there is a pending diff view open and the file associated with it no longer has changes (such as when changes are staged or deleted), then display the next diff view in its place according to what is selected in the staging view. This was a highly requested feature so that one could quickly stage using the keyboard by hitting `enter` multiple times. Or hit the “Stage File” button in a pending diff view in rapid sequence.

So let’s get clear on the circumstances under which we do want a pending diff item to be opened.

In order to get this behavior right, we need to get more specific around when/what/how exactly we want to automatically show the next file. 

**If there are stale pending pane items corresponding to a file that no
longer exists in the changed files list, update them to reflect the new
selection.**

When? The following criteria must be satisfied 
* A staging action has taken place, the repo updates, the staged file no longer appears in the staged changes list, and a new file is selected
* There is at least one pending diff view open corresponding to the file that has just been staged. Multiple pending diff views can be open across multiple panes, must check all
    * since we can’t easily know which file has been staged, we need to deduce this somehow 
        * diff view no longer has staging view file associated with it
        * diff view belongs to active repo (check to discern between staging action and repo change)

What? What parts of the workspace should be affected?
* Any pane that has a pending diff view that meets the criteria above should be updated such that its pending item corresponds to the new selection in the staging view
    * This behavior is consistent with the fact that staging lines/hunks in a diff view affects all diff views for that file open in other panes

How? Implementation details for updating pending items according to what is outlined above
* Check to see when there is a change in the selected file in the staging view
* Identify the panes that have pending diff views that no longer have staging view files associated with them and belong to the active repo
* For each pane, call workspace.open with the newly selected file and pass the pane as an option, as well as pending: true. This will cause the old diff view to be replaced

TODO: 
- [x] maintain selection across repo changes. 
repro steps for bug: open diff view for repo 1 that is not the first one listed. make repo 2 active. click diff view for repo 1 and note that it changes to be first file listed.


lib/views/staging-view.js
```diff
@@ -391,20 +391,60 @@ export default class StagingView {
...
-      if (openNew || pendingItem) {
-        const activate = pendingItem === this.props.workspace.getActivePaneItem();
-        await this.showFilePatchItem(selectedItem.filePath, this.selection.getActiveListKey(), {activate});
+      if (openNew) {
+        // User explicitly asked to view diff, such as via click
+        await this.showFilePatchItem(selectedItem.filePath, this.selection.getActiveListKey(), {activate: false});
+      } else {
+        const panesWithStaleItemsToUpdate = this.getPanesWithStalePendingFilePatchItem();
+        if (panesWithStaleItemsToUpdate.length > 0) {
+          // Staging event caused selection to change, so we need to update all old diff views to reflect new selection
```
kuychaco: Not just when staging. also when deleting files amd stashing

lib/views/staging-view.js
```diff
@@ -391,20 +391,60 @@ export default class StagingView {
...
+            });
+          }
+        }
       }
     }
   }
 
-  async showFilePatchItem(filePath, stagingStatus, {activate} = {activate: false}) {
+  getPanesWithStalePendingFilePatchItem() {
+    // "stale" meaning there is no longer a changed file associated with it due to contents being fully staged/unstaged
```
kuychaco: again, not just when staging/unstaging

test/views/staging-view.test.js
```diff
@@ -355,11 +355,11 @@ describe('StagingView', function() {
         });
         document.body.appendChild(view.element);
 
-        const getPendingFilePatchItem = sinon.stub(view, 'getPendingFilePatchItem').returns(false);
+        const getPanesWithStalePendingFilePatchItem = sinon.stub(view, 'getPanesWithStalePendingFilePatchItem').returns([]);
         await view.selectNext();
         assert.isFalse(showFilePatchItem.called);
 
-        getPendingFilePatchItem.returns(true);
+        getPanesWithStalePendingFilePatchItem.returns(['pending-file-patch-item']);
```
kuychaco: have multiple array items to improve this and test the code that runs `showFilePatchItem` for multiple panes

lib/views/staging-view.js
```diff
@@ -391,20 +391,60 @@ export default class StagingView {
...
+    return this.props.workspace.getPanes().filter(pane => {
+      const pendingItem = pane.getPendingItem();
+      if (!pendingItem) { return false; }
+      const isDiffViewItem = pendingItem.getRealItem && pendingItem.getRealItem() instanceof FilePatchController;
+      const isInActiveRepo = pendingItem.getWorkingDirectory() === this.props.workingDirectoryPath;
```
kuychaco: add comment `Diff views from repositories other than the active one do not count as stale. Ex: ...`

lib/views/staging-view.js
```diff
@@ -138,14 +139,22 @@ export default class StagingView {
...
     }
 
-    const selectionChanged = !isEqual(previouslySelectedItems, this.selection.getSelectedItems());
-    if (selectionChanged) {
-      this.debouncedDidChangeSelectedItem();
+    if (this.activeFilePatch) {
+      this.quietlySelectItem(this.activeFilePatch.getFilePath(), this.activeFilePatch.getStagingStatus());
+    } else {
+      const isRepoSame = oldProps.workingDirectoryPath === this.props.workingDirectoryPath;
+      const selectionsPresent = previouslySelectedItems.size > 0 && currentlySelectedItems > 0;
```
smashwilson: 🚨 It looks like `currentlySelectedItems` is an Array. Did you mean `currentlySelectedItems.length > 0`?

Also: the test suite is ✅ even though this is here. Does that mean we're missing a test case, or does it mean that the `currentlySelectedItems` isn't actually needed?

lib/views/staging-view.js
```diff
@@ -391,20 +400,61 @@ export default class StagingView {
         await this.showMergeConflictFileForPath(selectedItem.filePath, {activate: true});
       }
     } else {
-      const pendingItem = this.getPendingFilePatchItem();
-      if (openNew || pendingItem) {
-        const activate = pendingItem === this.props.workspace.getActivePaneItem();
-        await this.showFilePatchItem(selectedItem.filePath, this.selection.getActiveListKey(), {activate});
+      if (openNew) {
+        // User explicitly asked to view diff, such as via click
```
smashwilson: 👍 

I like explicitly separating the logic between "the user explicitly triggered this" and "we're inferring that this should happen."

lib/views/staging-view.js
```diff
@@ -391,20 +400,61 @@ export default class StagingView {
...
+              activate: false,
+              pane,
+            });
+          }));
+        } else {
+          // Selection was changed via keyboard navigation, update pending item in active pane
+          const activePane = this.props.workspace.getCenter().getActivePane();
+          const activePendingItem = activePane.getPendingItem();
+          const activePaneHasPendingFilePatchItem = activePendingItem && activePendingItem.getRealItem &&
+            activePendingItem.getRealItem() instanceof FilePatchController;
```
smashwilson: `instanceof` checks always make me feel vaguely uncomfortable. I think they raise my "OO code smell" instincts :smile:

(No useful alternatives to suggest here, though; we don't have much choice when we're dealing with arbitrary PaneItems.)

lib/views/staging-view.js
```diff
@@ -391,20 +400,61 @@ export default class StagingView {
...
+    });
+  }
+
+  async showFilePatchItem(filePath, stagingStatus, {activate, pane} = {activate: false}) {
     const encodedFilePath = encodeURIComponent(filePath);
     const encodedWorkdir = encodeURIComponent(this.props.workingDirectoryPath);
     const filePatchItem = await this.props.workspace.open(
       `atom-github://file-patch/${encodedFilePath}?workdir=${encodedWorkdir}&stagingStatus=${stagingStatus}`,
-      {pending: true, activatePane: activate, activateItem: activate},
+      {pending: true, activatePane: activate, activateItem: activate, pane},
```
smashwilson: Ah, it'll be good to control the Pane here.

... I still want to do that change in :atom: core to be able to specify "categories" of pending pane, so that there's only one pending pane of each category at once and we can keep ours independent from tree-view's.

test/views/staging-view.test.js
```diff
@@ -355,14 +355,17 @@ describe('StagingView', function() {
...
 
-        const getPendingFilePatchItem = sinon.stub(view, 'getPendingFilePatchItem').returns(false);
+        const getPanesWithStalePendingFilePatchItem = sinon.stub(view, 'getPanesWithStalePendingFilePatchItem').returns([]);
         await view.selectNext();
         assert.isFalse(showFilePatchItem.called);
 
-        getPendingFilePatchItem.returns(true);
+        getPanesWithStalePendingFilePatchItem.returns(['item1', 'item2']);
         await view.selectPrevious();
+        assert.isTrue(showFilePatchItem.calledTwice);
```
smashwilson: Since this is `.calledTwice`, shouldn't there be two `assert.isTrue(showFilePatchItem.calledWith(...))` calls below it to assert that they're both correct? Or are the other calls too annoying to assert specifically? 🤔 

Same for the next stanza there 👇

test/views/staging-view.test.js
```diff
@@ -423,15 +426,18 @@ describe('StagingView', function() {
...
-        const getPendingFilePatchItem = sinon.stub(view, 'getPendingFilePatchItem').returns(false);
+        const getPanesWithStalePendingFilePatchItem = sinon.stub(view, 'getPanesWithStalePendingFilePatchItem').returns([]);
         await view.selectNext();
         assert.isFalse(showFilePatchItem.called);
 
-        getPendingFilePatchItem.returns(true);
+        getPanesWithStalePendingFilePatchItem.returns(['item1', 'item2', 'item3']);
         await view.selectPrevious();
         await assert.async.isTrue(showFilePatchItem.calledWith(filePatches[0].filePath));
+        assert.isTrue(showFilePatchItem.calledThrice);
```
smashwilson: Hehe. Times you get to use the word "thrice" in source code++

smashwilson:
Nice ✨

PR #1315, opened 2018-02-21
Process draft

smashwilson:
This is an effort to document the way that we collaborate on the design and development of this package. It's based on an in-person meeting between @kuychaco and I where we attempted to formalize the way we work, and make some implicit conventions and practices that we use explicit. I've made a pass over our notes and done some wordsmithing; @kuychaco, please let me know if anything I've written here differs from your understanding of what we said 😉 

While a lot of this is what we're already doing, some of it is aspirational. Process is always hard to judge before it's put into practice, so we'll refine this as we go.

@simurai, one of the big things that we wanted to accomplish with this was including your input even though our timezones overlap so little. Let us know if you think we've mentioned you where you'd feel most useful, and if there's anywhere else you'd like to be involved that we haven't said :smile:

[Rendered](https://github.com/atom/github/blob/process/docs/how-we-work.md)


docs/how-we-work.md
```diff
@@ -0,0 +1,131 @@
+# How we work
+
+:sparkles: _Fabulously_ :sparkles:
+
+This is an attempt to make explicit the way that the core team plans, designs, and works day to day on the GitHub package. Our goal is to reduce ambiguity and uncertainty and improve communication, while avoiding artificial confinement and ceremony for ceremony's sake.
```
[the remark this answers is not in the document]
smashwilson: Ah, good question. We've talked about writing a separate document for external contributors ("how to get your pull request merged") but we've punted on it for the time being. I think we'd want to involve @lee-dohm in crafting that one, as well, and maybe adopt it beyond this package if the outcome is positive.

> On the other hand, might be not bad when people start to write rfcs's before making PRs?

I'd actually love to see this, especially for people tackling "big" changes, but I also don't want to mandate it for everything. Maybe modifications of existing design can be done first as a PR against an existing RFC? (That's not how "real RFCs" work, but we have commit history here...)

Long story short: that merits its own guide, content to be determined. Maybe we could research how projects like Kubernetes or Rust do this?

docs/how-we-work.md
```diff
@@ -0,0 +1,131 @@
...
+##### Examples
+
+* Porting the package from Etch to React.
+* Introducing additional diagnostics or metrics.
+* Restructuring an entire component tree.
+
+##### Process
+
+1. Propose these changes first in a conversation in Slack, a stand-up, or another synchronous channel. The decisions to be made here are:
+  * Does this change make sense to people other than me?
```
Arcanemagus: Add at least one (the correct amount is two) space(s) here (and in the other similar level bullet points) in order for them to be properly indented when rendered.

docs/how-we-work.md
```diff
@@ -0,0 +1,131 @@
...
+
+```sh
+$ apm publish minor
+version 0.11.0
+$ git branch 0.11-releases && git push -u origin 0.11-releases
+```
+
+When you merge a fix for a bug, cherry-pick the merge commit onto to the most recent release branch, then run `apm publish patch` and update `package.json` on the most recent beta release branch on the `atom/atom` repository. This will ensure bug fixes are delivered to users on Atom's stable channel as part of the next release.
+
+When you merge a fix for a **security problem**, a **data loss bug**, or fix a **crash** or a **lock-up** that affect a large portion of the user population, run `apm publish patch` and update `package.json` on the most recent beta _and_ stable release branches on the `atom/atom` repository. Consider advocating for a hotfix release of Atom to deliver these fixes to the user population as soon as possible.
```
Arcanemagus: Should this be hotfixed to the branch of the currently released `minor` version on Atom stable? As written this could involve an update of several minor versions of `github` to the `stable` branch for a "hotfix".

docs/how-we-work.md
smashwilson: 👍 Got it, thanks. I've also added a "rendered" link to the PR description too.

docs/how-we-work.md
smashwilson: Ah, good catch. I think I confused myself quite a bit when I was transferring this from our notes 😄 

Also: this is kind of cumbersome. I actually advocated for us _not_ to do this way back when we were getting started because it adds a lot of complexity and ceremony. If we end up having to do this more than once or twice I'll likely get us a script we can use to make sure the fixes go to the right branches and such. 🤔

Issue #554, opened 2017-02-23
Recent commit log


smashwilson:
Show the most recent two or three commits in beneath the CommitView. This will provide a bit of subtle feedback that creating a commit "did something" rather than disappearing into the aether.

Only render a short SHA and a (likely truncated) description line.

BinaryMuse:
Another issue talking about logging: https://github.com/atom/github/issues/823

simurai:
Now with the docks, the ~recent~ commit log could be its own panel. Then if you wanted, you could drag it under the Git panel and make it look like the mockup above (plus tab).

simurai:
Tried out the "be its own panel" in this static [prototype](https://github.com/atom/github-prototype/tree/history).

It's definitely nice be able to move the history panel anywhere that fits your window size or workflow. You can

- Switch between tabs
- Move it under the commit message
- Or side by side

![history](https://user-images.githubusercontent.com/378023/27576861-749a2b6e-5b59-11e7-9d16-13b2181f5425.gif)

A problem is that I think splitting dock items only gets serialized per project. So you would have to "split the history/log panel" in **every** project.

CharlineCouchot:
Hi there.

Don't know if this is the right place to make a suggestion or if I should open a new issue but the option to disable this log or hide it would definitely be nice.

Thanks !

PR #1318, opened 2018-02-22
Recent commit changes RFC

smashwilson:
A more formal proposal of the "recent commits" view displayed beneath the mini commit editor.

Previous discussion and design ideating are available in #554.

##### Acceptance criteria

We will merge this RFC and begin work when we accumulate ✅  reviews from @smashwilson, @kuychaco, and @simurai.

[Rendered](https://github.com/atom/github/blob/rfc-recent-commits/docs/rfcs/001-recent-commits.md)

## Implementation phases

- [x] Convert GitTabController and GitTabView to React. #1319
- [x] List read-only commit information. #1322
- [ ] Replace the amend checkbox with the "undo" control. #1328
- [ ] Context menu with actions. #1328
- [ ] Balloon with action buttons and additional information.
- [ ] Show which commits have been pushed.
- [ ] Show information about other refs.


docs/rfcs/001-recent-commits.md
```diff
@@ -0,0 +1,87 @@
...
+* Provide a discoverable launch point for an eventual log feature to explore the full history.
+* Achieve greater consistency with GitHub desktop:
+
+![desktop](https://user-images.githubusercontent.com/7910250/36570484-1754fb3c-17e7-11e8-8da3-b658d404fd2c.png)
+
+## Explanation
+
+If the active repository has no commits yet, display a short panel with a background message: "Make your first commit".
+
+Otherwise, display a **recent commits** section containing a sequence of horizontal bars for each of the top three commits reachable from the current `HEAD`, with the most recently created commit on top. The user can resize the recent commits section. As it is expanded or shrunk, the number of visible commits is changed responsively.
```
kuychaco: After creating new branch, only display commit that was branched off of. This may be technically challenging to implement (how do we know when to do this? last reflog entry is a checkout?).

docs/rfcs/001-recent-commits.md
```diff
@@ -0,0 +1,87 @@
...
+
+Each **recent commit** within the recent commits section summarizes that commit's metadata, to include:
+
+* GitHub avatar for both the committer and (if applicable) author. If either do not exist, show a placeholder.
+* The commit message (first line of the commit body) elided if it would be too wide.
+* A relative timestamp indicating how long ago the commit was created.
+
+On the most recent commit, display an "undo" button. Clicking "undo" performs a `git reset` and re-populates the commit message editor with the existing message.
+
+If any of the recent commits have been pushed to a remote, display a divider after the most recently pushed commit that shows an octocat icon. On hover, show the name of the remote tracking branch.
```
kuychaco: Additionally add dividers for other refs (branches and tags)

docs/rfcs/001-recent-commits.md
```diff
@@ -0,0 +1,87 @@
...
+- Add a "view diff from this commit" option to the recent commit context menu.
+- Integration with and navigation to "git log" or "git show" pane items when they exist.
+
+## Implementation phases
+
+1. Convert `GitTabController` and `GitTabView` to React.
+2. List read-only commit information.
+3. Replace the amend checkbox with the "undo" control.
+4. Context menu with actions.
+5. Tooltip with action buttons and additional information.
```
kuychaco: * Show which commits have been pushed
* Surface info about other refs

docs/rfcs/001-recent-commits.md
kuychaco: Just show last three commits? Or allow seeing more via scrolling or dragging to make view larger?

kuychaco: Also show commits that are on remote branch but not local (as a result of auto-fetch). Punt to show in full-featured log? Accordion style (expand/collapse). @simurai make mock-up to feel it out

docs/rfcs/001-recent-commits.md
kuychaco: @simurai would you mind making some quick mockups showing the dividers for remote tracking branch, local branch, tag with icon and ref name visible 🙏

kuychaco: consider when there are multiple refs for a given commit

docs/rfcs/001-recent-commits.md
```diff
@@ -0,0 +1,87 @@
...
+
+* For the most recent commit only, an "Amend" option. "Amend" is enabled if changes have been staged or the commit message mini-editor contains text. Choosing this applies the staged changes and modified commit message to the most recent commit, in a direct analogue to using `git commit --amend` from the command line.
+* A "Revert" option. Choosing this performs a `git revert` on the chosen commit.
+* A "Hard reset" option. Choosing this performs a `git reset --hard` which moves `HEAD` and the working copy to the chosen commit. When chosen, display a modal explaining that this action will discard commits and unstaged working directory context. Extra security: If there are unstaged working directory contents, artificially perform a dangling commit, disabling GPG if configured, before enacting the reset. This will record the dangling commit in the reflog for `HEAD` but not the branch itself.
+* A "Soft reset" option. Choosing this performs a `git reset --soft` which moves `HEAD` to the chosen commit and populates the staged changes list with all of the cumulative changes from all commits between the chosen one and the previous `HEAD`.
```
smashwilson: Plus: "mixed reset," consistent with the tool-tip actions :notes:

docs/rfcs/001-recent-commits.md
```diff
@@ -0,0 +1,87 @@
...
+
+On click, reveal a tool-tip containing:
```
kuychaco: In the future when we have a multi-file diff view, we can consider just opening a pane item with this information

docs/rfcs/001-recent-commits.md
```diff
@@ -0,0 +1,87 @@
...
+
+- Display tracking branch in separator that indicates which commits have been pushed. This could make the purpose of the divider more clear. Drawback is that this takes up space.
+
+## Unresolved questions
+
+- Allow users to view the changes introduced by recent commits. For example, interacting with one of the recent commits could launch a pane item that showed the full commit body and diff, with additional controls for reverting, discarding, and commit-anchored interactions.
+- Providing a bridge to navigate to an expanded log view that allows more flexible and powerful history exploration.
+- Show an info icon and provide introductory information when no commits exist yet.
+- Add a "view diff from this commit" option to the recent commit context menu.
+- Integration with and navigation to "git log" or "git show" pane items when they exist.
```
kuychaco: How do we surface that we've made a commit for you before doing a `git reset --hard`? So you never actually lose any unstaged changes

kuychaco: Right-click context menu on history? "Undo reset" until next commit is made.

docs/rfcs/001-recent-commits.md
smashwilson: "Balloon" instead of "tooltip" 🎈

docs/rfcs/001-recent-commits.md
smashwilson: Oh hey there's a [StackOverflow question](https://stackoverflow.com/questions/14848274/git-log-to-get-commits-only-for-a-specific-branch) about this:

```sh
git log mybranch --not $(git for-each-ref --format='%(refname)' refs/heads/ | grep -v "refs/heads/mybranch")
```

simurai:
Here a few dividers:

#### With label

divider1

#### Icon only

divider2

#### Tag

divider3

Hmm.. not sure. They use quite some space for such a small area and it's not clear if they are referring to the commits above or below a divider. Might fit better for the more fully-featured history log panel.

Alternatively we could have up/down arrows on the commits, matching the ahead/behind in the status-bar:

arrows

And tags could also be attached to a commit:

tag

docs/rfcs/001-recent-commits.md
smashwilson: Thanks to @simurai magic ✨ scrolling is a go. If we can, I'd really like to make it resizable by dragging too. If it's too technically challenging we can drop it, just like the "only show commits reachable by the current HEAD" bit up above, but I want to leave it in the RFC as a behavior goal if that's okay. I think I just feel like it'd be too frustrating not to be able to do 😁

smashwilson:
Ohh, fancy ✨ 

Looking over the options, some thoughts:

* I'd agree that the octocat divider and the dedicated tag line both occupy a little too much vertical real estate for the information they give.
* The up and down arrows are interesting! I like that they preserve the context of which commits are only local and only remote even if when the remote tracking branch ref isn't visible, and the symmetry with the push and pull buttons.
* I really like the bubble-ref in your last .gif - in fact, I think I prefer that to any of the other full-line dividers. We do need to be sure that it scales correctly when many refs refer to a single commit. Maybe after we reach some low ref threshold (two?) they get their own horizontal row above the message? Something like this:

  image

smashwilson:
I've updated the writeup based on the review comments. @kuychaco, @simurai, care to have another look and review when you get the chance? 🙇

ampinsk:
Hey all! @donokuda asked me to take a look here because of the overlap in problem space with some things we're working on at GHD.

Taking a look at the options you have here, I really like the icons direction. I think it does the best job working within the small timeline here. But, I think there might be a clearer icon to use than the up/down arrow. To me at least, the up/down arrow is pretty strongly associated with the push/pull _action_, not necessarily with the commit itself in terms of "this commit is ahead of my branch" 

It would be awesome to come up with a new icon that we can use (here and in GHD!) to signal "this commit is ahead/behind your branch" Maybe something like `|→` and `←|` ? 

Hope this is helpful! Let me know if I can clarify anything :)

simurai:
> @smashwilson We do need to be sure that it scales correctly when many refs refer to a single commit. 

Here a few refs below and aligned with the message:

refs

If they would be above, it might look like they belong to the commit above. We could also add divider lines for each commit, but then it quickly looks more heavy.

Hmm... also I think having these refs on a 2nd line makes messages less scannable because you get interrupted with something different when scrolling. How important is it to see these refs in the recent commits? I can see two options:

1. Only show tags (in the same line after the message). Their length (no more than `vX.X.X`) is somewhat predictable.
2. Show other refs on a 2nd line, but then disable scrolling.

simurai:
> @ampinsk Maybe something like `|→` and `←|` ?

Yeah, a hint of their meaning could come from how branches show the [ahead/behind](https://github.com/atom/github/branches):

image

The up/down arrows would be more from a local/remote perspective. Like this cheesy way of thinking that the remote is somewhere up in the cloud. Kinda like downloading and uploading.

```
commit ⬇︎ (you're behind: pull this down)
---- remote / local -----
commit ⬆︎ (you're ahead: push this up)
commit ⬆︎ (you're ahead: push this up)
commit
```

Looking at the up/down arrows, people might try to click them to actually push/pull? 🤔

docs/rfcs/001-recent-commits.md
```diff
@@ -0,0 +1,114 @@
...
+- Can we surface the commit that we make on your behalf before performing a `git reset --hard` with unstaged changes? Add an "Undo reset" option to the context menu on the recent commit history until the next commit is made? Show a notification with the commit SHA after the reset is complete?
+
+## Implementation phases
+
+1. Convert `GitTabController` and `GitTabView` to React. [#1319](https://github.com/atom/github/pull/1319)
+2. List read-only commit information. [#1322](https://github.com/atom/github/pull/1322)
+3. Replace the amend checkbox with the "undo" control.
+4. Context menu with actions.
+5. Balloon with action buttons and additional information.
+6. Show which commits have been pushed.
```
simurai: I think it should say ~~have been~~ "can be" pushed. And also show commits that can be pulled (after fetching).

And also move it up in the priority list, maybe as far as 3rd place?

simurai:
Here more feedback from @ampinsk + @donokuda. ❤️

1. When reloading Atom, it shows the "Make your first commit" for 1-2 seconds before showing the recent commits.
2. An alternative to the ahead/behind icons -> Have greyed out/bold text, different background color, or some other kind of highlight
3. Controversial: Instead of showing avatars -> show commit icon, with a vertical line like .com. This allows to have different states or icons like ahead/behind.
4. Don't show refs! Takes up too much space makes it harder to scan the messages.

And here the 2nd suggestion visualized:

screen shot 2018-03-02 at 5 36 11 pm

kuychaco:
We've completed Phases 1 and 2 with Phase 3 almost complete. With that we've reached our near-term goals for this feature 🎉 

The rest is icing and we'll hold off until future quarters in order to prioritize progress on other roadmap items. I'll go ahead and merge this now so that we can save all of the goodness to refer to when we pick this back up again. 

Thanks everyone for your fabulous input!

PR #1322, opened 2018-02-24
Add read only recent commits view

### LINKED ISSUE #554: Recent commit log

kuychaco:
Adding model changes and static markup for the recent commits view.

RFC: #1318 

### TODO

- [x] Fix layout in bottom dock
- [x] Make avatar dynamic
- [x] Make "time ago" dynamic
- [x] Format date (in title)
- [x] Display full message and body on hover

### Nice to have

- [x] Responsive list (grows with window height)
- [x] Show multiple authors (avatar accordion like Desktop)
- [ ] Render emoji in messages
- [ ] ~~Animate when commits appear~~ This is blocked until we have a "add a class only to new commits".

### Applicable issues

Closes #554


kuychaco:
@simurai if you like you can take a pass and give our markup and CSS a makeover. We still have to add some logic to convert the date to relative time stamps and grab GitHub avatars based on email addresses. But there's enough for you to work with right now if you wish to :). Otherwise we'll be picking up and adding more on Monday

simurai:
Made commits look more "compact" so they fit on a single line. Also, currently it has a max height of 3 commits, after it starts scrolling.

commits

Maybe ok like that? Well, depends a bit on these dividers (unpushed, unpulled, tags). Then it's harder to predict what height it should have.

lib/views/recent-commits-view.js
```diff
@@ -0,0 +1,69 @@
...
+    return (
+      <li className='github-RecentCommit'>
+        <img className='github-RecentCommit-avatar'
+          src="https://avatars3.githubusercontent.com/u/7910250?v=4&s=32"
+          title={`@kuychaco - ${this.props.commit.getAuthorEmail()}`}
+        />
+        <span className='github-RecentCommit-message'>{this.props.commit.getMessage()}</span>
+        <time className='github-RecentCommit-time'
+          title={this.props.commit.getAuthorDate()}>
+          1h
```
simurai: For the "time ago" format, to keep it narrow as possible, how about this:

```
<1m, 1m, 1h, 1d, 1w, 1M, 1y
```

So they are all lower case, except for the **month** to differentiate from **minutes**. When hovering and waiting for a bit, the title could show the date like `Feb 20, 2018` and maybe without time.

lib/views/recent-commits-view.js
```diff
@@ -0,0 +1,69 @@
...
+class RecentCommitView extends React.Component {
+  static propTypes = {
+    commit: PropTypes.object.isRequired,
+  };
+
+  render() {
+    return (
+      <li className='github-RecentCommit'>
+        <img className='github-RecentCommit-avatar'
+          src="https://avatars3.githubusercontent.com/u/7910250?v=4&s=32"
```
simurai: Default avatar (when there is no GitHub avatar) can be added with

```
src="atom://github/img/avatar.svg"
```

Should look like

screen shot 2018-02-27 at 3 11 07 pm

lib/models/avatar-cache.js
```diff
@@ -0,0 +1,38 @@
+import {Emitter} from 'event-kit';
+
+const LOADING_AVATAR_URL = 'https://github.com/simurai.png';
```
smashwilson: @simurai: 😇 

Think you could replace this with a suitable placeholder that we can show while we're waiting for the real avatar URL to come back from the GitHub API?

simurai: Maybe it could be the same as the default avatar: https://github.com/atom/github/pull/1322#discussion_r170824537

I'll try to add that.

simurai:
Bottom dock should be ok:

dock

simurai:
The list is now responsive and shows more commits (5, 10) if the window height gets larger.

responsive

This doesn't work well when you add a split pane, but maybe ok until there are persistent docks.

PR #1326, opened 2018-03-01
Shorter time ago in recent commits

simurai:
### Description of the Change

This shortens the "time ago". E.g. `2 hours ago` to just `2h`.

Before | After
--- | ---
screen shot 2018-03-01 at 2 30 59 pm | screen shot 2018-03-01 at 2 29 47 pm

- For "a few seconds" it says "Now".
- For months an uppercase `M` is used to differentiate from minutes.

### Alternate Designs

We could also use [twitter](https://momentjs.com/docs/#/plugins/twitter/) or [short date formatter](https://momentjs.com/docs/#/plugins/shortformat/) plugin. These only have a short form under 1 week.

### Benefits

Less space for the time ago, more space for the message + status icon or maybe refs.

### Possible Drawbacks

Might not be too obvious what the short form means. But maybe still fine, Twitter and other apps use it too.

### Applicable Issues

This PR is on top of #1322.

### DIFF
--- a/lib/views/recent-commits-view.js
+++ b/lib/views/recent-commits-view.js
@@ -4,25 +4,6 @@ import moment from 'moment';
 
 import Timeago from './timeago';
 
-moment.updateLocale('en', {
-  relativeTime : {
-    future: "%s",
-    past:   "%s",
-    s  : 'Now',
-    ss : '<1m',
-    m:  "1m",
-    mm: "%dm",
-    h:  "1h",
-    hh: "%dh",
-    d:  "1d",
-    dd: "%dd",
-    M:  "1M",
-    MM: "%dM",
-    y:  "1y",
-    yy: "%d y"
-  }
-});
-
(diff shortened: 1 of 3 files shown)


smashwilson:
Nice ✨ 

screen shot 2018-03-01 at 9 41 46 am

The `Timeago` component is already being used elsewhere in the package (specifically in the GitHub PR timeline) so I've added a `displayStyle` prop that we can use to toggle which format is used. I've also changed it to localize the short diffs in a custom locale derived from `en` rather than configuring Moment globally (_shakes fist at global state_) and added tests for both styles.

Thanks for getting this started and doing the legwork for customizing Moment there 😄

simurai:
> The `Timeago` component is already being used elsewhere in the package (specifically in the GitHub PR timeline) so I've added a `displayStyle` prop that we can use to toggle which format is used.

👍 Awesome. Yeah, in the pane, with more space, the longer form is nicer.

PR #1327, opened 2018-03-01
Remember Me?

### LINKED ISSUE #861: Allow for caching of Username and PW with GitHub Package
Each time you go to interact with a GitHub repository through Atom, you are required to enter Username and PW.  For those using a token due to 2FA, this is painful.  Please allow for caching/saving user login information.

screen shot 2017-05-23 at 5 47 52 pm

smashwilson:
Add a "remember me" checkbox to the authentication dialog that's launched by the bundled git credential helper. When checked, if authentication is successful (as detected by git invoking `git credential approve`), store the provided credentials in your OS keychain by way of [keytar](https://www.npmjs.com/package/keytar). If authentication fails, delete them.

##### Remaining

- [x] Refactor the `KeytarStrategy` implementations out of `github-login-model` so I can reuse them from the credential helper.
- [x] Introduce a `FileStrategy` that fakes keytar by reading and writing a JSON file. This is insecure as hell, so put in extra guards against it handling actual credentials.
- [x] Pre-transpile `keytar-strategy.js` and get its transpiled path so that it can be executed from the credential helper subprocess (from specs).
- [x] Attempt to read credentials from keytar within the credential handler before prompting from the Atom dialog. If credentials are found, short-circuit and reply.
- [x] Store the username as a per-host default as well if it wasn't provided in the original query.
- [x] Store a successfully used credential pair on `git credential approve`.
- [x] Delete an unsuccessfully used credential pair on `git credential reject`.
- [x] Re-use the token from the GitHub panel if it's been set. The interesting wrinkle here is that we don't have a username, so we'll need to pull it from the GitHub API.
- [x] Add the actual "remember me" text to the authentication dialog.
- [x] Test the pre-transpiled `keytar-strategy.js` in (a) dev mode and (b) when bundled within an Atom build. Find an alternative if it doesn't work.
- [x] Investigate and correct test failures
- [x] Give it a spin on Windows
- [x] Give it a spin on Linux

Fixes #861.

### DIFF
--- a/bin/git-credential-atom.js
+++ b/bin/git-credential-atom.js
@@ -285,11 +285,16 @@ async function toKeytar(query) {
 }
 
 /*
- *
+ * Remove credentials that failed authentication.
  */
-// async function deleteFromKeytar(query) {
-//   //
-// }
+async function deleteFromKeytar(query) {
+  const strategy = await createStrategy();
+
+  const gitService = `atom-github-git @ ${query.host}`;
+  log(`removing account "${query.username}" from service "${gitService}"`);
+  await strategy.deletePassword(gitService, query.username, query.password);
+  log('success');
+}
 
 async function get() {
   const query = await parse();
@@ -329,6 +334,7 @@ async function erase() {
   try {
     const query = await parse();
     await withAllHelpers(query, 'reject');
+    await deleteFromKeytar(query);
     log('success');
     process.exit(0);
   } catch (e) {
--- a/bin/git-credential-atom.js
+++ b/bin/git-credential-atom.js
@@ -3,6 +3,7 @@ const readline = require('readline');
 const url = require('url');
 const fs = require('fs');
 const path = require('path');
+const https = require('https');
 const {execFile} = require('child_process');
 const {GitProcess} = require(process.env.ATOM_GITHUB_DUGITE_PATH);
 const {createStrategy, UNAUTHENTICATED} = require(process.env.ATOM_GITHUB_KEYTAR_STRATEGY_PATH);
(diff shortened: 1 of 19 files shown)


smashwilson:
@simurai: When you get the chance, would you mind styling the "remember me" box I've added to the credentials dialog? It's not what you'd call "pretty" right now:

screen shot 2018-03-02 at 4 47 06 pm

To get it to pop up:

1. Clone a private repository with an `https` remote URL.
2. Disable any credential helpers you have installed. You can see which ones are installed by running:
    ```sh
   git config --get-all credential.helper
    ```
    I had to comment out `osxkeychain` from my system git configuration. The easiest way to do that is to run:
    ```sh
    EDITOR='atom --wait' git config --edit --system
    ```
3. Open the "Keychain Access" app and delete an `atom-github` key if you have one.
4. Open the repository in Atom and fetch with the branch controls.

Feel free to mess with the CSS classes, markup, verbiage, and anything - I'll get the tests ✅ again after you're done :smile:

jlsam:
This is probably something very trivial, but I installed Atom only last week and I keep seeing the dialog for authentication every time I try to push something. I don't see any check box there.

I was trying to follow the instructions to make it appear, but I got stuck at 'Open the "Keychain Access" app'. I have no idea what this could be, as I wasn't able to find anything like that within Atom. Can you explain better how to get there?

Also, I added a GitHub token authorization to Atom, is this important? I imagined this token would allow for authentication without login/pass, but that is not the case. What is the point of this token then?

smashwilson:
@jlsam This shipped in Atom 1.26 [just yesterday](http://blog.atom.io/2018/04/18/atom-1-26.html), so it's possible that your Atom installation just hasn't picked up the upgrade yet. If you _are_ on Atom 1.26 and you're not seeing the checkbox I'd love to know.

> I was trying to follow the instructions to make it appear, but I got stuck at 'Open the "Keychain Access" app'. I have no idea what this could be, as I wasn't able to find anything like that within Atom. Can you explain better how to get there?

Oh, those instructions were meant specifically for @simurai - he needed to be able to consistently reach the dialog to apply CSS magic. They aren't the general instructions needed to see the authorization dialog; you see it any time a git operation needs to interact with a remote and doesn't have credentials cached.

> Also, I added a GitHub token authorization to Atom, is this important? I imagined this token would allow for authentication without login/pass, but that is not the case. What is the point of this token then?

Previously, we used the token to authenticate to the GitHub API to fetch things like pull request data, but not to push. This PR changed that, so once you upgrade, the same token will be used for both.

smashwilson:
Ah, no, Atom doesn't update itself on Linux, we prefer to use the host distribution's mechanisms if possible. Rather than installing the .deb directly, you could try [installing it through aptitude](https://flight-manual.atom.io/getting-started/sections/installing-atom/#debian-and-ubuntu-debapt) to receive updates through a regular `apt-get upgrade`.

The `apm` commands update _packages_ that you have installed; the GitHub package is bundled with Atom, so you'll only receive updates for it when Atom itself is updated.

jlsam:
I added the PPA and reinstalled Atom. No password is even asked now, stuff just gets sent, I guess that's because of the certificate. Cool. What happens though, is Atom now says in 'About' that my system does not support automatic updates. I didn't run `sudo apt-get install atom-beta`, only `sudo apt-get install atom`. This is off-topic, perhaps I should start another ticket for this?

smashwilson:
That just means that Atom doesn't handle automatic updates internally - now you'll get updates through apt. You should be good now!

PR #1352, opened 2018-03-14
Add atom-ts-transpiler 

kuychaco:
So that we can start adding some typescript goodness :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,28 @@
           "graphql/schema.graphql"
         ]
       }
+    },
+    {
+      "transpiler": "atom-ts-transpiler",
+      "glob": "{lib,test}/**/*.tsx",
+      "options": {
+        "compilerOptions": {
+          "module": "commonjs",
+          "moduleResolution": "node",
+          "target": "es6",
+          "noImplicitReturns": true,
+          "noFallthroughCasesInSwitch": true,
+          "noUnusedLocals": true,
+          "sourceMap": true,
+          "jsx": "react",
+          "strict": true
+        },
+        "cacheKeyFiles": [
+          "package.json",
+          "assert-messages-plugin.js",
+          "graphql/schema.graphql"
+        ]
+      }
     }
   ],
   "dependencies": {


package.json
```diff
@@ -27,6 +27,28 @@
...
+        "compilerOptions": {
+          "module": "commonjs",
+          "moduleResolution": "node",
+          "target": "es6",
+          "noImplicitReturns": true,
+          "noFallthroughCasesInSwitch": true,
+          "noUnusedLocals": true,
+          "sourceMap": true,
+          "jsx": "react",
+          "strict": true
```
smashwilson: I'm not sure if setting `"strict": true` will cause problems while we're mid-transition? Like, I think interfacing with JavaScript is easier with some of the strict checks enabled. I'm not sure about specifics, though.

Let's default to keeping it enabled until we hit problems though :smile:

package.json
```diff
@@ -27,6 +27,28 @@
...
+        },
+        "cacheKeyFiles": [
+          "package.json",
+          "assert-messages-plugin.js",
+          "graphql/schema.graphql"
```
smashwilson: Are these accurate? I'd guess we'd want to keep only `package.json` here. I don't think the extra entries will hurt anything, though.

smhxx: A few minor "insights" on `cacheKeyFiles` that I can offer based on my experience with the custom transpilers system:

1. Atom already feeds a couple values from `package.json` into `crypto` when determining the currently valid cache key for a file, specifically the name and version number of the transpiling package, and the contents of the "options" property in the transpiler config. (This is an Atom thing, not an `atom-ts-transpiler` thing.) As such, doing something like changing the compiler options or bumping to a newer version of `atom-ts-transpiler` will invalidate the cache regardless of whether `package.json` is explicitly added to the `cacheKeyFiles.` The one case that I can think of where you would want to explicitly include `package.json` is if you want the cache to be invalidated when you bump the version of `typescript`, even if the source code and compiler options haven't changed.

2. As of v1.3.0, `atom-ts-transpiler` will include the contents of the appropriate `tsconfig.json` file in the cache key data (including values from any other configs extended by that config,) so if you do choose to add one or more tsconfigs in the future, they don't need to be explicitly added.

3. Any files added to `cacheKeyFiles` will be read from disk in their entirety the first time each transpiled file is required (per-session,) meaning that there *is* a minuscule penalty to package load time associated with each entry. I don't know exactly how long it would be in this case, but the impact of including a file with ~8000 LOC like `graphql/schema.graphql` might be noticeable. In general, it's better to omit any entries that won't have a direct effect on transpilation output.

smashwilson:
Alas, we likely shouldn't do this until `apm` gets some attention. Because of the way that bundled packages are currently transpiled, we would need to include typescript as a non-dev dependency, which would add _30MB_ to the Atom download. Sadly that's enough of a bump that we probably shouldn't take that leap. 😢

PR #1359, opened 2018-03-19
Use unambiguous branch name when pushing

### LINKED ISSUE #1358: Atom fails to push: 'src refspec matches more than one'
### Description

Atom pops up error in red box when pushing a branch:

> error: src refspec my_branch matches more than one.
> error: failed to push some refs to '...'

### Steps to Reproduce

1. Have a repo with a remote
2. Have a tag that is the same name as the current branch ('my_branch')
3. in atom ui try and push

**Expected behavior:**

pushes without displaying error

**Actual behavior:**

gives git error message

**Reproduces how often:**

100%

### Additional Information

Doesn't seem to effect checkout/fetch/pull, they all succeed (in atom ui or on cmd line)

### Versions

Atom        : 1.25.0
Electron    : 1.7.11
Chrome      : 58.0.3029.110
Node        : 7.9.0
atom/github : 0.10.3

win10

iPherian:
### Description of the Change

When pushing, use unambiguous "refs/heads/branch_name" instead of "branch_name"

### Benefits

Avoids ambiguity with other kinds of git refs like tags. git push will fail if a ref string refers to more than one thing. Fixes #1358

### Applicable Issues

#1358

### DIFF
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -658,7 +658,7 @@ export default class GitShellOutStrategy {
   }
 
   push(remoteName, branchName, options = {}) {
-    const args = ['push', remoteName || 'origin', branchName];
+    const args = ['push', remoteName || 'origin', `refs/heads/${branchName}`];
     if (options.setUpstream) { args.push('--set-upstream'); }
     if (options.force) { args.push('--force'); }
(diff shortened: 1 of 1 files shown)


smashwilson:
👍 Cool, nice catch.

PR #1362, opened 2018-03-22
Fix flakey GitTabController test 

kuychaco:
Async operations were causing a race condition. The solution was to await promises for the staging operation and selection update.


kuychaco:
@annthurium congrats on your first! 🚢 :shipit: 🚀

PR #1355, opened 2018-03-16
Commit together with co-authors

kuychaco:
Details to come ;) 

commit-with-friends

TODO:
- [x] update user store after initial loading (what if new commits are pulled with new authors?)
- [x] cache invalidation for `getAuthors`

In separate PRs:
- [ ] use `%x1F` delimiter for `getRecentCommits`
- [ ] hide behind feature flag
- [ ] add new author (use prompt similar to CredentialDialog)
- [ ] get username and avatar url from github graphql API
- [ ] get rid of web committer ([like so](https://github.com/desktop/desktop/blob/76216ec81d638eb7c0333e5934dec03ea9de5469/app/src/models/commit.ts#L93-L126))
- [ ] consider sorting based on most recent authors or # of commits rather than alphabetically
- [ ] get rid of amend checkbox and  implement undo and “amend” right-click instead
    * as is, clicking the amend checkbox will not strip co-author trailers and you could end up with duplicate co-authors. so we'll want to make this change before shipping to beta

/cc @niik @simurai

PR #1374, opened 2018-04-03
Add dialog for new co-author

kuychaco:
This PR builds upon #1355 and adds UI for users to add a co-author who is not in the list of suggested co-authors (which is based on who has previously committed to the repository).

TODO: 
- [x] change 'New User' to be 'New author'
- [x] have "New User" always in list
- [x] validate email address
- [x] disable "Add co-author" button if input is missing or invalid email address is provided (type="email")
- [x] auto populate "Name" field
- [x] determine the right amount of labeling
- [x] @simurai ✨ 
- [x] Closing toggle should empty coauthor list
- [x] Tooltip says "add" and "remove" coauthors instead of "toggle" on the toggle toggle
- [x] Italic or so treatment of "new user" list item
- [x] When the co-author input field is shown, the co-author icon should look "active"
- [x] Make toggle icons in commit box hover with hand (cursor: pointer)
- [x] When toggling coauthors on, focus input field
- [x] When focusing from editor to input field, input blurs and staged changes receive focus
- [x] ensure `tab` focuses correct input fields or button
- [x] after `esc` and after `return` from new author form, focus co-author input list
- [x] Get rid of noreply@github.com emails from GitHub web commits
- [x] Esc button in input should close email suggestion list
- [x] ~Tab key from input field should focus commit button if it doesn’t select an entry.~ If it does select an entry, it should not change field focus. For committing via keyboard, add a tooltip for commit button that says you can `cmd-enter` to commit
- [x] Editing commit message seems to reset the used-filter of the suggestion list
- [x] Fix flaky UserStore test
- [x] Styling of box does not distinguish it enough from background. add border around co-author input in all themes
- [x] Filter self from suggestion list
- [x] What happens if you use a private email address?
  - [x] As per help center documentation, users must change their local git settings to use private email addresses from the command line.  Tested this flow by changing my email to private, making a commit, and then changing my email back to the public one.  After that, the private email address showed up in suggested co author menu, along with my name.  Instead of the user avatar, the default "no avatar" icon was shown. I believe that's reasonable behavior.

Lower priority TODOs:
- [ ] Add github avatars to suggestion list (based on email address and usercontent url)
  - [ ] and show "no user" for email addresses with no github user (maybe generic gravatar or octocat or identicon?)
  - [ ] add "new user" icon for "New User" selection
- [ ] Remove github commands for navigating co-author suggestion list
- [ ] List component has rendering performance issues when e.g. holding down `arrow up` for continuous scrolling
- [ ] use `%x1F` delimiter for `getRecentCommits`
- [ ] Add “Copy SHA” link in recent commits box when hovering over a commit (out of scope?)
- [ ] get username and avatar url from github graphql API (separate PR)

### DIFF
--- a/lib/controllers/commit-controller.js
+++ b/lib/controllers/commit-controller.js
@@ -245,6 +245,10 @@ export default class CommitController extends React.Component {
   hasFocus() {
     return this.domNode && this.domNode.contains(document.activeElement);
   }
+
+  hasEditorFocus() {
+    return this.refCommitView.hasEditorFocus();
+  }
 }
 
 function wrapCommitMessage(message) {
--- a/lib/controllers/commit-controller.js
+++ b/lib/controllers/commit-controller.js
@@ -178,7 +178,6 @@ export default class CommitController extends React.Component {
     }
   }
 
-  @autobind
   isCommitMessageEditorExpanded() {
     return this.getCommitMessageEditors().length > 0;
   }
--- a/lib/controllers/commit-controller.js
+++ b/lib/controllers/commit-controller.js
@@ -246,8 +246,8 @@ export default class CommitController extends React.Component {
     return this.domNode && this.domNode.contains(document.activeElement);
   }
 
-  hasEditorFocus() {
-    return this.refCommitView.hasEditorFocus();
+  hasFocusEditor() {
+    return this.refCommitView.hasFocusEditor();
   }
 }
 
--- a/lib/models/repository-states/present.js
+++ b/lib/models/repository-states/present.js
@@ -648,9 +648,18 @@ export default class Present extends State {
     return bundle.branch.aheadBehind.behind;
   }
 
+  // config
+
   getConfig(option, {local} = {local: false}) {
     return this.cache.getOrSet(Keys.config.oneWith(option, {local}), () => {
       return this.git().getConfig(option, {local});
+    });
+  }
+
+  getCommitter() {
+    // todo (tt, 4/2018): figure out the best way to just watch `user.name` and `user.email`. ASK ASH!! ask about local: false
+    return this.cache.getOrSet(Keys.config.oneWith('user.email', {local: false}), () => {
+      return this.git().getCommitter();
     });
   }
 
--- a/lib/models/repository-states/present.js
+++ b/lib/models/repository-states/present.js
@@ -648,19 +648,26 @@ export default class Present extends State {
     return bundle.branch.aheadBehind.behind;
   }
 
-  // config
-
   getConfig(option, {local} = {local: false}) {
     return this.cache.getOrSet(Keys.config.oneWith(option, {local}), () => {
       return this.git().getConfig(option, {local});
     });
   }
 
-  getCommitter() {
-    // todo (tt, 4/2018): figure out the best way to just watch `user.name` and `user.email`. ASK ASH!! ask about local: false
-    return this.cache.getOrSet(Keys.config.oneWith('user.email', {local: false}), () => {
-      return this.git().getCommitter();
-    });
+  async getCommitter(options = {}) {
+    const output = await this.getConfig(['--get-regexp', '^user.*'], options);
+    const committer = {name: null, email: null};
+    // todo (tt, 4/2018): do we need null byte terminated output here for Windows?
+    if (output) {
+      output.trim().split('\n').forEach(line => {
+        if (line.includes('user.email')) {
+          committer.email = line.slice(11);
+        } else if (line.includes('user.name')) {
+          committer.name = line.slice(10);
+        }
+      });
+    }
+    return committer;
   }
 
   // Direct blob access
--- a/lib/views/co-author-dialog.js
+++ b/lib/views/co-author-dialog.js
@@ -24,6 +24,7 @@ export default class CoAuthorDialog extends React.Component {
     this.state = {
       name: '',
       email: '',
+      submitDisabled: true,
     };
   }
 
@@ -53,7 +54,7 @@ export default class CoAuthorDialog extends React.Component {
         <label className="github-CoAuthorDialog-row">
           <span className="github-CoAuthorDialog-label">Email:</span>
           <input
-            type="text"
+            type="email"
             placeholder="foo@bar.com"
             ref={e => (this.emailInput = e)}
             className="input-text github-CoAuthorDialog-email"
@@ -64,7 +65,7 @@ export default class CoAuthorDialog extends React.Component {
         </label>
         <footer className="github-CoAuthorDialog-row has-buttons">
           <button className="btn github-CancelButton" tabIndex="3" onClick={this.cancel}>Cancel</button>
-          <button className="btn btn-primary" tabIndex="4" onClick={this.confirm}>Add Co-Author</button>
+          <button className="btn btn-primary" disabled={this.state.submitDisabled} tabIndex="4" onClick={this.confirm}>Add Co-Author</button>
         </footer>
       </div>
     );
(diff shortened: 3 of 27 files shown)


simurai:
Currently looks like this:

screen shot 2018-04-03 at 5 05 22 pm

The `NewCoAuthor` dialog is "attached" to the `coAuthorEditor` so it feels like they still belong together.

lib/views/co-author-dialog.js
```diff
@@ -83,11 +84,24 @@ export default class CoAuthorDialog extends React.Component {
...
     this.setState({email: e.target.value});
+    this.validate();
+  }
+
+  @autobind
+  validate() {
+    // email validation with regex has a LOT of corner cases, dawg.
+    // https://stackoverflow.com/questions/48055431/can-it-cause-harm-to-validate-email-addresses-with-a-regex
+    // to avoid bugs for users with nonstandard email addresses,
+    // just check to make sure email address contains `@` and move on with our lives.
```
smashwilson: 💯 💯 💯

annthurium:
@kuychaco @simurai I think we need the buttons for the following reasons:
- a minimal UI is not always the clearest UI.  as a user, I'd be confused about how to submit in the absence of a button.
- having buttons allows us to give a visual indicator of validation state, by disabling the button until input is validated.

simurai:
>  I wonder... could we get away with having no buttons?

As a minimal UI lover, I wouldn't mind :grin:, but @annthurium has some valid points about making it less confusing. Since this dialog is temporary, you only see it once in a while.

I made another change so that the buttons fill the entire space (no odd gap on the left) and wrap when the panel is too narrow.

co-author

kuychaco:
To make it more clear what the toggle icon does, I propose we copy the behavior of GitHub desktop:

toggleicon

Which is to say, rather than use our tooltips (which show up after quite a delay and are a bit "heavy"), show the text next to the icon. And also change the text to say add/remove when appropriate.

kuychaco:
Oh, but then it's not consistent with the icons on the right for hard-wrapping and expanding the commit message editor. And I don't think changing those icons to have this behavior would work well because of how the icons are positioned and the fact that they are side-by-side. Nevermind then, maybe we should just stick with our tooltip but have the text change between "Add" and "Remove". We can also make the delay time for showing shorter

lib/models/repository-states/present.js
```diff
@@ -648,12 +648,21 @@ export default class Present extends State {
...
 
+  // config
+
   getConfig(option, {local} = {local: false}) {
     return this.cache.getOrSet(Keys.config.oneWith(option, {local}), () => {
       return this.git().getConfig(option, {local});
     });
   }
 
+  getCommitter() {
```
annthurium: @smashwilson

lib/views/commit-view.js
```diff
@@ -196,8 +219,8 @@ export default class CommitView extends React.Component {
     /* eslint-disable max-len */
     const svgPath = 'M9.875 2.125H12v1.75H9.875V6h-1.75V3.875H6v-1.75h2.125V0h1.75v2.125zM6 6.5a.5.5 0 0 1-.5.5h-5a.5.5 0 0 1-.5-.5V6c0-1.316 2-2 2-2s.114-.204 0-.5c-.42-.31-.472-.795-.5-2C1.587.293 2.434 0 3 0s1.413.293 1.5 1.5c-.028 1.205-.08 1.69-.5 2-.114.295 0 .5 0 .5s2 .684 2 2v.5z';
     return (
-      <svg className="github-CommitView-coAuthorToggleIcon" viewBox="0 0 12 7" xmlns="http://www.w3.org/2000/svg">
-        <title>Toggle co-authors</title>
+      <svg className={cx('github-CommitView-coAuthorToggleIcon', {focused: this.state.showCoAuthorInput})} viewBox="0 0 12 7" xmlns="http://www.w3.org/2000/svg">
+        <title>Toggle Add or remove co-authors</title>
```
smashwilson: 🚨 🎨  Double capital here: `Toggle Add [...]`

The wording is also a bit awkward here. How about something like "Toggle co-author entry"? @simurai might have ideas too :smile:

lib/views/commit-view.js
```diff
@@ -299,6 +358,13 @@ export default class CommitView extends React.Component {
   toggleCoAuthorInput() {
     this.setState({
       showCoAuthorInput: !this.state.showCoAuthorInput,
+    }, () => {
+      if (this.state.showCoAuthorInput) {
+        this.refCoAuthorSelect.focus();
```
smashwilson: :sparkles: Nice. I didn't even know I wanted this to happen until I saw you two making it happen 😆

lib/views/commit-view.js
annthurium: I meant to change this from "Toggle co author" to "Add/remove co authors" but I fat fingered it. Oops! Will fix. And yeah, I agree the wording could be better.

lib/controllers/commit-controller.js
```diff
@@ -246,6 +245,10 @@ export default class CommitController extends React.Component {
   hasFocus() {
     return this.domNode && this.domNode.contains(document.activeElement);
   }
+
+  hasFocusEditor() {
+    return this.refCommitView.hasFocusEditor();
+  }
```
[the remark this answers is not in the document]
kuychaco: Yup, agreed. I'd love to clean it up. We're relying on tabindexes for the CommitView and we're manually managing focus for the StagingView. If we convert the StagingView lists to use `<select>` tags (it's just a bunch of divs right now), we could potentially manage them with tabindex using traditional focus conventions. Additionally with select tags, I think the somewhat complicated selection logic we've implemented can be replaced with the out-of-the-box selection behavior that comes with native DOM elements. 

In any case, perhaps we can make improvements when we get around to refactoring the StagingView component from Etch to React?

lib/views/commit-view.js
```diff
@@ -115,6 +122,11 @@ export default class CommitView extends React.Component {
 
     const showAbortMergeButton = this.props.isMerging || null;
 
+    const modKey = process.platform === 'darwin' ? 'Cmd' : 'Ctrl';
+
+    // TODO: [ku 4/9/2018] fix tooltip target method (we're using `document.querySelector` because
+    // this.refCoAuthorToggle is temporarily null when the component is unmounted/remounted)
```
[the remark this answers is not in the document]
kuychaco: It's a bit more complicated. That would result in no new tooltip once the DOM node was swapped out. I poked around a bit trying to figure out what's going on, but it was super late at night and my memory is fuzzy on it. I _think_ what's happening is that when we re-render the co-author icon DOM node is being replaced with a new DOM node. During this swap, `this.refCoAuthorToggle` is temporarily set to `null`. While this is the case, the Tooltip component tries to add the tooltip to the `null` target. So, rather than relying on the ref that gets nulled out, we use `document.querySelector('.github-CommitView-coAuthorToggle')` to get ahold of the target to add the tooltip to. 

Changing the `componentWillReceiveProps` to a `componentDidUpdate` seems to fix the issue and allow us to rely on the ref as opposed to querying the DOM. @smashwilson do you see any problems with this?

lib/controllers/commit-controller.js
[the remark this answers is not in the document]
smashwilson: > In any case, perhaps we can make improvements when we get around to refactoring the StagingView component from Etch to React?

:+1: That sounds like a good time to get to it :smile:

Certainly it wouldn't make sense until we do that, I'd say.

lib/views/commit-view.js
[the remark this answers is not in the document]
smashwilson: > Changing the componentWillReceiveProps to a componentDidUpdate in the Tooltip component seems to fix the issue and allow us to rely on the ref as opposed to querying the DOM.

:sparkles: Perfect - especially because [`componentWillReceiveProps()` is being deprecated](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops).

PR #1379, opened 2018-04-06
Add naming convention

simurai:
### Description of the Change

This adds a CSS naming convention ([rendered](https://github.com/atom/github/blob/sm-naming-convention/docs/naming-convention.md)) to `docs/`. We already have been using it more or less, but still good to have it written down somewhere.

### Alternate Designs

- Should it stay CSS only or mixed with JS or other conventions?
- Is it too strict and hard to remember? It could also be simplified more. On the other hand, it's just a guide and fine if not followed always 100%.
- There is no definition where a CSS component "starts" and "ends". It could be related to components in JS, like `github-<ReactComponentName>-<something>`.
- Could we potentially use other ways to style components? Like using CSS in JS? Might not be easy to make the Less theme variables work. And harder for themes, users to override, than just use simple classes.

### Benefits

👉 [Benefits](https://github.com/atom/github/blob/sm-naming-convention/docs/naming-convention.md#benefits).

### Possible Drawbacks

Yet another thing to remember.

### Applicable Issues

This is a straight up copy from https://github.com/atom/design-decisions/pull/5

### DIFF
--- a/docs/naming-convention.md
+++ b/docs/naming-convention.md
@@ -1,0 +1,195 @@
+# CSS Naming Convention
+
+
+This is Atom's naming convention for creating UI in Atom and Atom packages. It's close to [BEM/SUIT](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md), but slightly customized for Atom's use case.
+
+
+## Example
+
+Below the commit box as a possible example:
+
+```html
+<div class='gitub-CommitBox'>
+  <div class='gitub-CommitBox-editor'></div>
+  <footer class='gitub-CommitBox-footer'>
+    <button class='gitub-CommitBox-button'>Commit to sm-branch</button>
+    <div class='gitub-CommitBox-counter is-warning'>50</div>
+  </footer>
+</div>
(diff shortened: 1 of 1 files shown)


simurai:
/cc @annthurium that ❤️ flexbox

This PR has low priority, just added it so I don't forget after reading the chat.

annthurium:
@simurai this is awesome -- thanks for documenting!!

my opinion: css-in-js is great for, like, Facebook where they have 1000s of developers doing ui development on the same codebase, which makes specificity unwieldy.  Since we are not at that scale in terms of team size or codebase size, css in js would be overkill.

The convention you specified here is close enough to BEM to be easy to remember.

simurai:
OK, let's keep it simple. 👍 

> The convention you specified here is close enough to BEM to be easy to remember.

Right, it's BEM with a modified syntax. Pascal/camelCase, instead of underscores. Very close to [SUIT](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md).

PR #1390, opened 2018-04-13
React 16.3.1

smashwilson:
Upgrade us to React 16.3.1 🚀 🌔 

I'm intending to merge this when I've dealt with all test failures and newly introduced deprecations, but I _won't_ be tackling the use of any new features introduced by React 16 just yet, to keep this from sprawling.

- [x] Update `<Portal>` components to use the official `ReactDOM.createPortal()`, and the corresponding tests to use Enzyme
  - [x] PaneItem
  - [x] Decoration
  - [x] DockItem
  - [x] Panel
  - [x] Tooltip
- [x] Actually test it all out live

Supersedes #1174.


smashwilson:
21 failures! Not bad. This should be faster than the Enzyme upgrade was (he says hopefully).

smashwilson:
@kuychaco by the way, since it looks like I'll be revisiting Tooltip as part of the upgrade anyway, want me to tackle this?

https://github.com/atom/github/blob/5f0bf0e0c40679a80c625bd34d71d5ce778029e3/lib/views/commit-view.js#L133

kuychaco:
@smashwilson sure go for it! When I took a stab at it, I hit a snag with the push/pull tooltips FYI. Thanks for taking this on!

smashwilson:
:tada: :tada: :tada:

@atom/github-package: I've fixed all of the failing tests from our suite and I've addressed the problems I noticed in dev mode, but it's quite possible that there are other regressions that I didn't catch. (Yes, this probably means we need a better test suite :eyes:)

PR #1413, opened 2018-04-23
Update recent commits RFC

simurai:
### Description of the Change

This moves the "refs" and "un-pulled commits" to the alternatives section. We might reconsider adding them in a log/history view.

### Benefits

Makes sure the RFC is accurate, but none of the longer term vision is lost.

### Possible Drawbacks

None

### Applicable Issues

Part of #1318

### DIFF
--- a/docs/rfcs/001-recent-commits.md
+++ b/docs/rfcs/001-recent-commits.md
@@ -109,4 +109,3 @@ The modal dialog on "hard reset" is disruptive considering that the lost changes
 4. Context menu with actions.
 5. Balloon with action buttons and additional information.
 6. Show which commits have been pushed.
-7. Show information about other refs.
--- a/docs/rfcs/001-recent-commits.md
+++ b/docs/rfcs/001-recent-commits.md
@@ -29,11 +29,11 @@ If the active repository has no commits yet, display a short panel with a backgr
 
 Otherwise, display a **recent commits** section containing a sequence of horizontal bars for ten **relevant** commits with the most recently created commit on top. The commits that are considered **relevant** include:
 
-* Commits reachable by the remote tracking branch that is the current upstream of `HEAD`.
+* Commits reachable by the remote tracking branch that is the current upstream of `HEAD`. If more than three of these commits are not reachable by `HEAD`, they will be hidden behind an expandable accordion divider.
 * Commits reachable by `HEAD` that are not reachable by any local ref in the git repository.
 * The single commit at the tip of the branch that was branched from.
 
-The most recent three commits are visible by default and the user can scroll to see up to the most recent ten commits.
+The most recent three commits are visible by default and the user can scroll to see up to the most recent ten commits. The user can also drag a handle to resize the recent commits section and show more of the available ten.
 
 ### Commit metadata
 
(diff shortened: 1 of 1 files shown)


docs/rfcs/001-recent-commits.md
```diff
@@ -42,37 +42,16 @@ Each **recent commit** within the recent commits section summarizes that commit'
...
-* A greyed-out state if the commit is reachable from the remote tracking branch but _not_ from HEAD (meaning, if it has been fetched but not pulled).
 
 ### Undo
 
 On the most recent commit, display an "undo" button. Clicking "undo" performs a `git reset` and re-populates the commit message editor with the existing message.
 
-### Refs
-
-Annotate visible commits that correspond to refs in the git repository (branches and tags). If the commit list has been truncated down to ten commits from the full set of relevant commits, display a message below the last commit indicating that additional commits are present but hidden.
-
```
smashwilson: You made a convincing argument that we shouldn't annotate refs at all. We don't have a lot of room to work with and it wouldn't add _that_ much information... refs are likely best left for the log view.

docs/rfcs/001-recent-commits.md
```diff
@@ -29,11 +29,11 @@ If the active repository has no commits yet, display a short panel with a backgr
 
 Otherwise, display a **recent commits** section containing a sequence of horizontal bars for ten **relevant** commits with the most recently created commit on top. The commits that are considered **relevant** include:
 
-* Commits reachable by the remote tracking branch that is the current upstream of `HEAD`. If more than three of these commits are not reachable by `HEAD`, they will be hidden behind an expandable accordion divider.
+* Commits reachable by the remote tracking branch that is the current upstream of `HEAD`.
 * Commits reachable by `HEAD` that are not reachable by any local ref in the git repository.
 * The single commit at the tip of the branch that was branched from.
 
-The most recent three commits are visible by default and the user can scroll to see up to the most recent ten commits. The user can also drag a handle to resize the recent commits section and show more of the available ten.
+The most recent three commits are visible by default and the user can scroll to see up to the most recent ten commits.
```
smashwilson: I think we do still want to make this resizable. Although I will say I find myself not missing it as much as I thought I would.

docs/rfcs/001-recent-commits.md
```diff
@@ -42,37 +42,16 @@ Each **recent commit** within the recent commits section summarizes that commit'
 * GitHub avatar for both the committer and (if applicable) author. If either do not exist, show a placeholder.
 * The commit message (first line of the commit body) elided if it would be too wide.
 * A relative timestamp indicating how long ago the commit was created.
-* A greyed-out state if the commit is reachable from the remote tracking branch but _not_ from HEAD (meaning, if it has been fetched but not pulled).
```
smashwilson: I'm a little conflicted about this one. Do we still want to do this? I feel like the fetched commits would almost always overwhelm the panel. Maybe we could have a `3 fetched` line instead... ?

docs/rfcs/001-recent-commits.md
```diff
@@ -42,37 +42,16 @@ Each **recent commit** within the recent commits section summarizes that commit'
...
-* A "Revert" option. Choosing this performs a `git revert` on the chosen commit.
-* A "Hard reset" option. Choosing this performs a `git reset --hard` which moves `HEAD` and the working copy to the chosen commit. When chosen, display a modal explaining that this action will discard commits and unstaged working directory context. Extra security: If there are unstaged working directory contents, artificially perform a dangling commit, disabling GPG if configured, before enacting the reset. This will record the dangling commit in the reflog for `HEAD` but not the branch itself.
-* A "Mixed reset" option. Choosing this performs a `git reset` on the chosen commit.
-* A "Soft reset" option. Choosing this performs a `git reset --soft` which moves `HEAD` to the chosen commit and populates the staged changes list with all of the cumulative changes from all commits between the chosen one and the previous `HEAD`.
```
smashwilson: We do want to add these as well... I think we're waiting to have a way to undo them first.

docs/rfcs/001-recent-commits.md
```diff
@@ -42,37 +42,16 @@ Each **recent commit** within the recent commits section summarizes that commit'
...
-
-On click, select the commit and reveal a balloon containing:
-
-* Additional user information consistently with the GitHub integration's user mention item.
-* The full commit message and body.
-* The absolute timestamp of the commit.
-* Navigation button ("open" to a git show-ish pane item)
-* Action buttons ("amend" on the most recent commit, "revert", and "reset" with "hard", "mixed", and "soft" suboptions)
-
-![commit-popout](https://user-images.githubusercontent.com/17565/36570682-11545cae-17e8-11e8-80a8-ffcf7328e214.JPG)
```
smashwilson: Definitely still want this. I always want to click on the commits :smile: Figuring out when will be part of our roadmapping this week, I think.

docs/rfcs/001-recent-commits.md
simurai: I think it's handy to get a hint what the commits are about before you pull the changes. Like when you still have to finish something and can't pull right now, you already know what will come. Since these are on the remote, they could even be click-able and open a browser window with a diff.

That said, I agree that if you're more than 2-3 commits behind, it buries your local changes too much. So I'm ok to punt on it and leave it for a possible log/history.

simurai:
> but just haven't scheduled in our roadmap yet

Sorry.. I think I misread the [we'll hold off until future quarters](https://github.com/atom/github/pull/1318#issuecomment-377064520) as "we might get back to this in a few months". :sweat_smile: But the next quarter already starts in a week.

Ok, how about I'll revert everything and _only_ move the "refs" and "fetched but not pulled commits" to alternatives. We can always move the other things later, if we decide not to do them.

PR #1414, opened 2018-04-23
Rework PaneItem management

### LINKED ISSUE #1405: Unified PaneItem management
We have two competing styles of `PaneItem` management:

1. The [`<PaneItem>` component](https://github.com/atom/github/blob/master/lib/views/pane-item.js) uses its single child component as its "item" within a React portal. It's currently used to display `FilePatchController` trees.
2. The components within the [`lib/atom-items` directory](https://github.com/atom/github/tree/master/lib/atom-items) are independent components that share some common internal structure and render themselves to separate React roots.

I'm not entirely fond of either style, to be honest. The `<PaneItem>` container element has two sources of truth for when its item should be rendered - the contents of the Atom workspace, and the structure of the React tree - and getting the two out of sync causes unmounted component chaos. But, the atom-items components duplicate a lot of code and split the React tree, which makes testing more difficult.

I'd like to find a third alternative that addresses these and migrate all of our existing pane items to it. We only have three major items right now, so it shouldn't be too onerous once we decide on a direction to pursue.

smashwilson:
Another bit of ambient cleanup work. This is me trying to get our `PaneItem` and opener management consistent and React-y. If we like it, I should be able to port over all of our existing `PaneItems` pretty quickly, as we don't have that many.

Here's a spoiler of what I'm building toward:

```js
render() {
  return (
    <PaneItem
      workspace={this.props.workspace}
      uriPattern='atom-github://something/root/{filePath...}?q={search}'>
      {({itemHolder, params}) => (
        <ItemComponent
          ref={itemHolder.setter}
          filePath={params.filePath}
          search={params.search}
          another='prop'
        />
      )}
    </PaneItem>
  )
}
```

Fixes #1405.

#### Pane items to convert

- [x] FilePatchController
- [x] IssueishPaneItem
- [x] GitTimingsView


smashwilson:
@kuychaco @annthurium: I'm curious if you two like the API I've come up with here, and if you think it's a net benefit from the way that we used to manage pane items.

For reference, here's the new style:

* [PaneItem](https://github.com/atom/github/blob/341a2c2d025fd8a031c67ab5ae31eea4ac382b37/lib/atom/pane-item.js) is the major bit;
* [URIPattern](https://github.com/atom/github/blob/341a2c2d025fd8a031c67ab5ae31eea4ac382b37/lib/atom/uri-pattern.js) is a support class;
* [IssueishPaneItem](https://github.com/atom/github/blob/341a2c2d025fd8a031c67ab5ae31eea4ac382b37/lib/atom-items/issueish-pane-item.js) is an example of an item component, with [this section from RootController](https://github.com/atom/github/blob/341a2c2d025fd8a031c67ab5ae31eea4ac382b37/lib/controllers/root-controller.js#L331-L341) as an example of the actual `<PaneItem>` tree.

The old style had a few different variants:

* [FilePatchController](https://github.com/atom/github/blob/master/lib/controllers/file-patch-controller.js) used a [PaneItem](https://github.com/atom/github/blob/b54ca06cf2234c872c14c9870a95e0badb476a12/lib/controllers/root-controller.js#L307-L345) and some state within [GithubPackage](https://github.com/atom/github/blob/master/lib/github-package.js#L87).
* [IssueishPaneItem](https://github.com/atom/github/blob/master/lib/atom-items/issueish-pane-item.js#L12-L74) and [GitTimingsView](https://github.com/atom/github/blob/master/lib/views/git-timings-view.js#L353-L367) used variants on creating a detached DOM node and rendering independent React trees within their own static methods.

smashwilson:
😆 Sorry about that, @BinaryMuse. I'm guessing I was overzealously clicking on suggested reviewers?

lib/atom/uri-pattern.js
```diff
@@ -0,0 +1,366 @@
+import url from 'url';
+
+/**
+ * Match and capture parts of a URI, like a specialized dialect of regular expression. This is used by PaneItem to
```
annthurium: yay documentation!!

annthurium:
looks good to me!

lib/atom/pane-item.js
```diff
@@ -0,0 +1,155 @@
...
+ * render() {
+ *   return (
+ *     <PaneItem workspace={this.props.workspace} uriPattern='atom-github://host/{id}'>
+ *       {({itemHolder, params}) => (
+ *         <ItemComponent ref={itemHolder.setter} id={params.id} />
+ *       )}
+ *     </PaneItem>
+ *   );
+ * }
+ */
```
kuychaco: 👍👍

smashwilson:
Cool, thanks for the sanity check :smile:

Issue #270, opened 2016-10-05
How Enterprise?


BinaryMuse:
At minimum we'll need to wait till the GraphQL API is available there, but we're also constantly adding fields to GitHub.com's codebase to account for our API use case — so how can we reliably support Enterprise sooner rather than later?

chickenkiller:
For information, GitHub Enterprise 2.10.0 has just been released, with [support for GraphQL API](https://enterprise.github.com/releases/2.10.0/notes).

Usage of GraphQL API through GitHub Enterprise is documented [here](https://developer.github.com/enterprise/2.10/v4/guides/intro-to-graphql/).

BinaryMuse:
While this release of Enterprise supports GraphQL, the GitHub integration part of this package is still in progress and we're still adding fields to GitHub.com's GraphQL API as we need them, and these new fields will take some amount of time to make it to an Enterprise release.

In an ideal world, we could do some sort of feature detection and use different "versions" of the integration depending on which versions of the GraphQL API are available, but this is something we haven't yet talked about so I don't have any immediate plans to share. It is on my mind, however!

quasipedia:
Since 5 months have passed I thought to ask... Is there any update to share on the status of this? :)

ungarida:
This is a kind of a deal breaker between for using  VisualStudio over Atom. Is there any realistic plan to do this?

tsal:
It's been over a year since GHE started supporting graphql in the API - an update would be nice.

jjnich:
💀 waiting for github enterprise support 💀

sidx64:
This is a much needed feature. Can someone please get this going?

quasipedia:
So, your bot says that at GitHub, you treat your issues as your backlog.... but then you mark each comment from your users who bring your attention to this as "spam"?  I don't get it, my product owner would die to have spontaneous input from our users that would inform him about what users deem important for them...

Is there any any other way we should use to let you know what is important for us?

lee-dohm:
@quasipedia You can use emoji reactions on the original post to upvote it if you want to let us know that this is something that is important to you. The reason why we minimize these comments as off-topic or spam is because they do nothing to move the issue toward completion. When an issue bubbles to the top and we go to look for people giving us helpful information such as debug logs, steps to reproduce the problem, or other helpful information, having to scroll through bunches of "is this done yet?" makes our job take longer. If you would like to give direct input to the team on your thoughts about prioritization, you can email us at atom@github.com.

[...]

PR #1419, opened 2018-04-25
RFC for Interactive Log View

smashwilson:
Starting to write down my thoughts around this. Highly incomplete.

:zap: :art: [rendered](https://github.com/atom/github/blob/aw/interactive-log-view-rfc/docs/rfcs/003-interactive-log-view.md) :art: :zap:

### DIFF
--- a/docs/rfcs/003-interactive-log-view.md
+++ b/docs/rfcs/003-interactive-log-view.md
@@ -1,0 +1,63 @@
+# Interactive Log View
+
+## Status
+
+Proposed
+
+## Summary
+
+Open a pane item displaying a navigable, live 3D model of your git commit graph.
+
+## Motivation
+
+Visualizing the commit graph is an important tool for learning git. A live view of the commit graph can inobtrusively help git newcomers form the correct mental model for commit graph operations (branching, merging, rebasing).
+
+We do not currently have a way to initiate merge, rebase, or cherry-pick operations from the editor. A log view would be the most logical place to do so, because it's where you have the most context about the existing relationship of the refs you'd like to merge or rebase.
+
+A log view would grant us a way to answer the question of where you are relative to nearby refs and tags.
+
+Why 3D? Flat visualizations of git history quickly fall victim to the "guitar hero problem":
+
+[![Git guitar hero](https://user-images.githubusercontent.com/17565/39223169-6f5d6b1e-480e-11e8-809b-60d1dc8261fb.png)](https://twitter.com/henryhoffman/status/694184106440200192)
+
+I'm hopeful that the extra "space" afforded by rendering in WebGL and level-of-detail fading can give us a view that remains functional even in the face of dramatic tangles.
+
+## Explanation
+
+The log view is a pane item launched by:
+
+* The `github:log` command in the command palette.
+* An "Open Log" menu item in the context menu on the recent commit history.
+
+When launched, the log view pane uses WebGL to render a 3D model of your commit graph.
+
+* Each commit is rendered as a sphere, possibly with adjacent text (see below).
+* A line joins each commit to its parent.
+* Refs are rendered as rectangular tags attached to the relevant commits.
+* When a remote is present, commits that are not reachable from any remote ref are rendered _[TODO]_. Commits that have been fetched but are not reachable from any local ref are rendered _[TODO]_.
+
+### Level of detail
+
+Commit information quickly becomes visually overwhelming, even in relatively small repositories. To remain useful as repositories grow in size and complexity, we can employ varying _level of detail_ for commits.
+
+The _focused commit_ is the one currently at the center of the field of view. It is annotated with author avatars, a timestamp, and a commit message header. Zero to many _selected commits_ have been chosen by a hotkey to participate in a pending action, so they too have the highest level of detail.
+
+Commits adjacent to the focused one on the same branch (reachable by only a single entry in `refs/heads`, up to a maximum) are "nearby" and rendered with the next highest level of detail. Each has a timestamp and a commit message header in text next to the commit node.
+
+### Layout algorithm
+
+_[TODO]_
+
+### Searching and filtering
+
+### Navigation
+
+### Merge, rebase, cherry-pick
+
+## Drawbacks
+
+## Rationale and alternatives
+
+## Unresolved questions
+
+## Implementation phases


docs/rfcs/003-interactive-log-view.md
```diff
@@ -0,0 +1,63 @@
...
+I'm hopeful that the extra "space" afforded by rendering in WebGL and level-of-detail fading can give us a view that remains functional even in the face of dramatic tangles.
+
+## Explanation
+
+The log view is a pane item launched by:
+
+* The `github:log` command in the command palette.
+* An "Open Log" menu item in the context menu on the recent commit history.
+
+When launched, the log view pane uses WebGL to render a 3D model of your commit graph.
```
simurai: @smashwilson Do you have an example of a 3D log visualization or is it something that only exists in your imagination. 💭

smashwilson: Totally and completely in my head 😆 

If I get a chance, I might start playing around with a prototype for this to see if it's viable.

docs/rfcs/003-interactive-log-view.md
```diff
@@ -0,0 +1,63 @@
...
+
+Proposed
+
+## Summary
+
+Open a pane item displaying a navigable, live 3D model of your git commit graph.
+
+## Motivation
+
+Visualizing the commit graph is an important tool for learning git. A live view of the commit graph can inobtrusively help git newcomers form the correct mental model for commit graph operations (branching, merging, rebasing).
```
Arcanemagus: "inobtrusively" -> "unobtrusively"

Arcanemagus:
This sounds like a really interesting idea for seeing things immediately around where you are right now, but how will it handle large repositories?

hundreds of commits? thousands? _millions_?

smashwilson:
Scaling up and still being useful is the big challenge in my mind, yeah - to the point that I don't think any existing tool, including `git log`, really does it well.

I was hopeful that these would help:

* LOD scaling - commits close to the focus one show a lot of information; commits far from the focus one only show the shape of the branch structure, and maybe tags.
* Filtering and searching - press `/` and filter the view by commit messages, ref names, etc.
* Some way to focus and inspect the path among a set of refs (what commits are in the symmetric difference? what's their merge base?)

But it's a big problem which will likely have a long tail of trial and error. I was hoping to track down a few example repositories (torvalds/linux, etc) that would push the limits and see how quickly performance and understandability degrade.

simurai:
> how will it handle large repositories?

A while back I was playing around with a log view, well it's not really a typical log view, but more a Git/GitHub navigation or time machine. The inspiration comes from how you navigate through time when setting a date, like this date picker on iOS:

image

Instead of showing a chronological, never-ending list of commits, it would group them in parallel by versions > PRs > commits. This would let you navigate in small or big steps. It also gives you a better understand what commits "belong" together, instead of having the "guitar hero" lines, it would "show me all PRs from this version" or "show me all commits from this PR".

screen shot 2018-05-03 at 9 01 51 am

If you commit straight to master, then the single commit would show up in the 2nd column (since there is no PR). If a repo has no versions and no PRs, then I guess it would just show the long list of commits.

The grey highlight is where your current HEAD is. The faded PRs/commits above show what you haven't pulled yet. With auto-fetch, it would let you know if your team members made new PRs/commits and you know what will happen if you pull. You can "pull" by simply double clicking on a faded commit/PR. This lets you "move forward in time" in a much more controlled way than just pulling all the changes at once. You can also go back in time by double clicking on any version, PR, or commit below your "HEAD line". Single-click could show a preview with a diff and more infos.

Not sure if moving back in time should do a reset hard? Maybe to make this possible at any time, it could create a temporary worktree. Then you can move back in time without having to commit and push first. You could see how code used to be, run tests, copy delete code or so and then just a quickly fast-forward again and your changes are still there.

Clicking above `MASTER` would show PRs that haven't been merged into master yet:

screen shot 2018-05-03 at 11 04 15 am

Sorted by recent changes. So it gives you a hint what might come next. Also handy to quickly checkout and test/contribute to a PR.

Here a more compact version that might even fit at the top of the tree view. You could drill down and either the versions, PRs or commits column would expand:

Showing PRs | Showing commits of a PR
--- | ---
image | screen shot 2018-05-03 at 9 02 44 am

Be able to navigate between files AND history in the same place might be handy. Less switching between panels and you get to see files getting added/removed as you time travel. Each column is scrollable, but the history "section" could also be resizable or collapsable to just a single item (if you really want to save space).

Haven't really thought about stuff like merge, rebase, cherry-pick, squash, rename etc. Maybe by drag&drop? 

---

Anyways, since this goes into a different direction, it probably should be its own RFC. Also, the two approaches don't have to be mutually exclusive. We'll just have to fight over priority. :stuck_out_tongue_winking_eye:

docs/rfcs/003-interactive-log-view.md
HotelCalifornia: On a semi-related note, I met some guys who made a git branch viewer for the hololens (it would display the branch graph fixed in space). So I guess that's one realization of a 3D log visualisation

calebmeyer:
Is there a reason to have this here instead of in the git package? The reason I ask is it applies equally (except for PR features) to git projects hosted elsewhere (or not hosted at all), and I'd really like to see this usable for github enterprise (we're still [not there yet](https://github.com/atom/github/issues/270) with this package).

docs/rfcs/003-interactive-log-view.md
```diff
@@ -0,0 +1,63 @@
...
+
+* The `github:log` command in the command palette.
+* An "Open Log" menu item in the context menu on the recent commit history.
+
+When launched, the log view pane uses WebGL to render a 3D model of your commit graph.
+
+* Each commit is rendered as a sphere, possibly with adjacent text (see below).
+* A line joins each commit to its parent.
+* Refs are rendered as rectangular tags attached to the relevant commits.
+* When a remote is present, commits that are not reachable from any remote ref are rendered _[TODO]_. Commits that have been fetched but are not reachable from any local ref are rendered _[TODO]_.
```
calebmeyer: There's basically 2 pieces of information you want to convey (preferably visually and at a glance) here:
- Local, remote, or both
- which remote(s)

My initial thought (assuming this view has a strong sense of up and down) is that local only commits are rendered as a pyramid (pointing up, which mildly indicates push). Remote only commits are rendered as an inverted pyramid (pointing down, which mildly indicates pull). Any commit that is both local and remote will be a sphere.

My idea for indicating which remote/remotes a commit is part of is to render an icon for each remote a commit is part of. We could use a house icon for origin, and a wavy (diagonal upwards) icon for upstream. Maybe basic geometric shapes (square, circle, triangle, star) for any other remotes. 

If you give each remote its own color, then you could mix the colors to determine the color of the sphere/pyramid. Maybe local only is gray, origin is green, upstream is blue, etc. Then upstream and origin would be cyan/teal (mix of blue and green), etc.

calebmeyer: This scheme works well with the levels of detail. Highest level of detail lets you hover the remote icons to see the names for the remotes. Medium level of detail lets you see the icons, but maybe smaller, or maybe only the first remote. Low level of detail has no icons, but retains the shape and color, so you can visually see if there's a "far away" commit that hasn't been pushed yet.

docs/rfcs/003-interactive-log-view.md
```diff
@@ -0,0 +1,63 @@
...
+
+The log view is a pane item launched by:
+
+* The `github:log` command in the command palette.
+* An "Open Log" menu item in the context menu on the recent commit history.
+
+When launched, the log view pane uses WebGL to render a 3D model of your commit graph.
+
+* Each commit is rendered as a sphere, possibly with adjacent text (see below).
+* A line joins each commit to its parent.
```
calebmeyer: parent(s)*
merge commits have two parents.

Arcanemagus: Merge commits actually have as many parents as needed. Two is just the most common.

PR #1420, opened 2018-04-25
More recent commits RFC edits

simurai:
### Description of the Change

More edits. This time, no moving around, mostly adding mockups for:

- [x] Highlight un-pushed commits
- [x] Ballon

### Applicable Issues

Continuation of #1318

### DIFF
--- a/docs/rfcs/001-recent-commits.md
+++ b/docs/rfcs/001-recent-commits.md
@@ -70,7 +70,7 @@ On click, select the commit and reveal a balloon containing:
 * Navigation button ("open" to a git show-ish pane item)
 * Action buttons ("amend" on the most recent commit, "revert", and "reset" with "hard", "mixed", and "soft" suboptions)
 
-![commit-popout](https://user-images.githubusercontent.com/17565/36570682-11545cae-17e8-11e8-80a8-ffcf7328e214.JPG)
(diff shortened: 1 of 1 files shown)


simurai:
Here a mockup for the "ballon":

ballon

> Navigation button ("open" to a git show-ish pane item)

Is 👆 similar to the current `FilePatchView`? Showing a pane with a diff, but potentially including multiple files? Like on [.com](https://github.com/atom/github/commit/aa8818dbf1e0c0e3effe1c3264b72d93cc1d2b35). Should that be part of a different RFC? Maybe once there is a log view that could also use it.

If so, I'm not so sure how useful the ballon is. The action buttons at the bottom could be in the context menu. It's maybe even quicker to click because you don't have to move with the mouse that far. The only benefit of the ballon would be to see the entire commit message and body. Is that enough?

Here another proposal:

- **Short term**: Clicking on a recent commit opens the commit on .com (with diff and all the infos). Un-pushed commits can't be clicked, so they will have different styling as a hint.
- **Long term**: Once we have a git show-ish pane, clicking on a recent commit opens the git show-ish pane directly (without ballon).

smashwilson:
> Is 👆 similar to the current FilePatchView? Showing a pane with a diff, but potentially including multiple files? Like on .com. Should that be part of a different RFC? Maybe once there is a log view that could also use it.

👍 Pretty much exactly where we were going with this, I think.

And yeah, I agree with your assessment. We can move the buttons as-is to the pane item once it exists.

PR #1421, opened 2018-04-25
Make ListSelections immutable

smashwilson:
To lay some groundwork for converting `StagingView` to React, I'm converting our `ListSelection` models to be immutable.

Rather than updating the selection in-place:

```js
const selection = new ListSelection({items: ['a', 'b', 'c'});
selection.selectFirstItem();
```

Mutation methods return a new ListSelection with the modification made:

```js
const selection0 = new ListSelection({items: ['a', 'b', 'c']);
const selection1 = selection0.selectFirstItem();
```

This should make manipulating selection models in `setState()` calls safer and more ergonomic, because we can use its [updater method variant](https://reactjs.org/docs/react-component.html#setstate):

> `prevState` is a reference to the previous state. It should not be directly mutated. Instead, changes should be represented by building a new object based on the input from `prevState` and `props`.

```js
actionMethod = () => {
  this.setState(prevState => ({
    theSelection: prevState.theSelection.selectFirstItem(), 
  }));
}
```

- [x] Make `ListSelection` immutable
- [x] Make `CompositeListSelection` immutable
- [x] Update the `StagingView` to use its `CompositeListSelection` immutably


smashwilson:
By the way, I did this for `FilePatchSelection` a long time ago when I converted the `FilePatchController` to React. So right now _one_ of our selection models is immutable and the others are mutable, which bugs me. :smile:

smashwilson:
:shipit: to prepare for the final phase of the React port :metal:

Issue #381, opened 2016-12-18
Convert Etch components to React components


smashwilson:
Incrementally complete the conversion from etch to React throughout the codebase. This will be easiest to do if we go top-down, gradually pushing the `EtchWrapper` boundary further into our component hierarchy.

- [x] controllers/status-bar-tile-controller.js
  - [x] views/branch-view.js
  - [x] views/branch-menu-view.js
  - [x] views/push-pull-view.js
  - [x] views/push-pull-menu-view.js
  - [x] views/changed-files-count-view.js
- [x] controllers/file-patch-controller.js
  - [x] views/file-patch-view.js
    - [x] views/hunk-view.js (`HunkView` + `LineView`)
- [x] controllers/git-tab-controller.js
  - [x] views/git-tab-view.js
    - [x] views/staging-view.js
      - [x] views/file-patch-list-item-view.js
      - [x] views/merge-conflict-list-item-view.js
    - [x] controllers/commit-view-controller.js
      - [x] views/commit-view.js

smashwilson:
Finished! We are now 100% free-range, organic React :tada:

PR #1424, opened 2018-04-27
Port the StagingView to React

smashwilson:
The final part of #381.

- [ ] Tests green
- [ ] Try it out live


smashwilson:
Our tests are green and everything seemed to be well-behaved while I was trying it live, but there is a fair amount of risk of this regressing _something_. I'm merging now so we have a decent chunk of time to find regressions before the next ship :smile:

PR #1428, opened 2018-05-01
Upgrade Relay to 1.6.0

smashwilson:
Upgrade Relay and associated dependencies to the latest version (1.6.0). Bring our GraphQL schema up to date and regenerate all generated query files.

Let's see if this breaks horribly 😅


smashwilson:
🙌

Issue #1368, opened 2018-03-30
Pressing "Escape" when asked to confirm a Force Push executes the Force Push


mattmattmatt:
### Prerequisites

* Have you tried to reproduce the problem in [Safe Mode][safe-mode]? **Yes.**
* Have you followed all applicable steps in the [debugging guide][debugging-guide]? **Yes.**
* Have you checked the FAQs on the [message board][message-board] for common solutions? **Yes.**
* Have you checked that your issue [isn't already filed][already-filed]? **Yes.**

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

### Description

Pressing "Escape" when asked to confirm a Force Push executes the Force Push instead of canceling the Push.

### Steps to Reproduce

1. Cmd/Ctrl-click the "Push" button in the interface or use command palette to "Force Push"
1. Dialog window pops up, asking to "Cancel Push" or "Force Push"
1. Press "Escape" key on keyboard

**Expected behavior:**

Dialog window closes without Force Pushing (as if "Cancel Push" was clicked).

**Actual behavior:**

Force Push is being executed (as if "Force Push" was clicked).

**Reproduces how often:**

All of the times.

### Versions

macOS Sierra, v10.12.6

```shell
Atom    : 1.25.0
Electron: 1.7.11
Chrome  : 58.0.3029.110
Node    : 7.9.0

apm  1.19.0
npm  3.10.10
node 6.9.5 x64
atom 1.25.0
python 2.7.12
git 2.14.3
```

rsese:
Thanks @mattmattmatt :+1:  I can reproduce on macOS 10.12.6 with Atom 1.26.0-beta0 - clicking `Cancel Push` works ok.

PR #1452, opened 2018-05-08
Fix so dialogs are canceled when pressing escape/closing them

Ben3eeE:

### Requirements

### Description of the Change

The label for canceling a dialog needs to be "Cancel" or "No" so that when canceling the dialog either by pressing escape or by closing the dialog cancels the dialog. Otherwise these actions default to the first item. This can be a bad default choice for the force push dialog which would force push when you close the dialog.

image

### Alternate Designs

There is a cancelId option [according to the electron documentation](https://electronjs.org/docs/api/dialog) but this is not supported on Windows.

### Benefits

* Don't accidentally force push or merge with conflict markers when the user intended to cancel the dialog
* Moves the cancel button to where it should be

### Possible Drawbacks

* Someone relies on pressing escape to execute a force push? :sweat_smile: 

### Applicable Issues

Fixes https://github.com/atom/github/issues/1368

### DIFF
--- a/lib/controllers/root-controller.js
+++ b/lib/controllers/root-controller.js
@@ -648,7 +648,7 @@ export default class RootController extends React.Component {
       detailedMessage: `for the following files:\n${conflictedFiles}\n` +
         'Would you like to apply the changes with merge conflict markers, ' +
         'or open the text with merge conflict markers in a new file?',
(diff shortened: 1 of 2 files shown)


smashwilson:
> Someone relies on pressing escape to execute a force push? 😅

Oh no my workflow!

Haha, thanks @Ben3eeE :rocket:

Issue #1463, opened 2018-05-15
Slowness when typing in commit message editor due to frequent `loadUsersFromLocalRepo` calls


maxbrunsfeld:
### Prerequisites

* Have you tried to reproduce the problem in [Safe Mode][safe-mode]?
* Have you followed all applicable steps in the [debugging guide][debugging-guide]?
* Have you checked the FAQs on the [message board][message-board] for common solutions?
* Have you checked that your issue [isn't already filed][already-filed]?

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

### Description

When typing a commit message, Atom starts to feel sluggish. I recorded a profile and it looks like `loadUsersFromLocalRepo` is getting called each time I type a key.

### Steps to Reproduce

1. Open the commit message editor
2. Start recording a profile
3. Start typing

**Expected behavior:**

Low CPU usage.

**Actual behavior:**

High CPU usage.

**Reproduces how often:**

100% for me.

### Versions

```
Atom    : 1.29.0-dev-c950a3e89
Electron: 2.0.0
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

### Additional Information

Zoomed-out profile (showing multiple keystrokes):

![screen shot 2018-05-15 at 11 26 03 am](https://user-images.githubusercontent.com/326587/40076224-37ee31f6-5833-11e8-8c32-48b3dc292e5b.png)

Zoomed-in profile (showing details):

![screen shot 2018-05-15 at 11 26 11 am](https://user-images.githubusercontent.com/326587/40076239-425005f2-5833-11e8-9be1-688ed38e2724.png)

annthurium:
should be fixed by https://github.com/atom/github/pull/1464 but please let me know if any additional issues come up.

maxbrunsfeld:
⚡️

PR #1464, opened 2018-05-15
stop calling `repository.didUpdate` when commit message changes.

annthurium:
We are performing a bunch of unnecessary operations when the commit message changes.  It's causing some slowness.  

See https://github.com/atom/github/issues/1463 for details.

As far as I can tell, updating the repository when the commit message changes may have been necessary at some point in the past (like for the old way amending used to work, perhaps?) But it doesn't seem to be useful now.  Nevertheless, it's important to test carefully here to make sure removing this won't cause any surprise regressions.

**Test Plan**

manually tested the following flows:
- make a commit from Git pane, with and without co authors
- undo a commit, with and without co authors
- make a commit with expanded editor
- toggle back and forth between expanded editor and commit editor in Git pane
- make a commit from the command line.  Verify that commit shows up in recent commit history.

unit tests
- added a test that spies on the function we don't want to be called.  Negated assertions to prove that it's actually testing the thing we care about.

### DIFF
--- a/lib/models/repository-states/present.js
+++ b/lib/models/repository-states/present.js
@@ -45,11 +45,7 @@ export default class Present extends State {
   }
 
   setCommitMessage(message) {
-    const oldMessage = this.commitMessage;
     this.commitMessage = message;
-    if (oldMessage !== message) {
-      this.didUpdate();
-    }
   }
 
   getCommitMessage() {
--- a/test/controllers/commit-controller.test.js
+++ b/test/controllers/commit-controller.test.js
@@ -87,6 +87,16 @@ describe('CommitController', function() {
       assert.strictEqual(wrapper.find('CommitView').prop('message'), 'some message');
     });
 
+    it('repository.didUpdate is not called when commit message changes', function() {
+      repository.setCommitMessage('some message');
+      const wrapper = shallow(app, {disableLifecycleMethods: true}).instance();
+      sinon.spy(wrapper.props.repository.state, 'didUpdate');
+      assert.strictEqual(wrapper.getCommitMessage(), 'some message');
+      wrapper.handleMessageChange('new message');
+      assert.strictEqual(wrapper.getCommitMessage(), 'new message');
+      assert.isFalse(wrapper.props.repository.state.didUpdate.called);
+    });
+
     describe('when a merge message is defined', function() {
       it('is set to the merge message when merging', function() {
         app = React.cloneElement(app, {isMerging: true, mergeMessage: 'merge conflict!'});
--- a/test/controllers/commit-controller.test.js
+++ b/test/controllers/commit-controller.test.js
@@ -87,7 +87,7 @@ describe('CommitController', function() {
       assert.strictEqual(wrapper.find('CommitView').prop('message'), 'some message');
     });
 
-    it('repository.didUpdate is not called when commit message changes', function() {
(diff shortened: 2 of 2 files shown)


test/controllers/commit-controller.test.js
```diff
@@ -87,6 +87,16 @@ describe('CommitController', function() {
       assert.strictEqual(wrapper.find('CommitView').prop('message'), 'some message');
     });
 
+    it('repository.didUpdate is not called when commit message changes', function() {
```
maxbrunsfeld: This is very minor, but it looks like all of the surrounding tests' descriptions are phrased so that they sound right after the word "it". I'd maybe tweak the description to say something like

`it("does not cause the repository to update")`.

lib/models/repository-states/present.js
```diff
@@ -45,11 +45,7 @@ export default class Present extends State {
   }
 
   setCommitMessage(message) {
-    const oldMessage = this.commitMessage;
     this.commitMessage = message;
-    if (oldMessage !== message) {
-      this.didUpdate();
-    }
```
smashwilson: Oh, interesting. Changing the message triggered an update, and therefore at least a partial re-render of the React tree, so that UI elements that change state based on the message's length and presence have a chance to be updated. Specifically I'm thinking of:

* The enablement of the "Commit" button; it should be disabled if the message is empty; and
* The "remaining characters" counter that changes as you type the subject line;

But, the way the commit message is handled has been shuffled a lot with the introduction of the `AtomTextEditor` wrapper component and various refactorings, so I'm not 100% sure that these things _won't_ be updated by something else.

Can you confirm that those didn't break? If they still work, then we definitely should take this out because it's not needed. If they don't, an alternate approach would be to fuss with either the `UserStore` or the `GitTabController` to fine-tune when `loadUsersFromLocalRepo` is invoked. We should only really need to re-call it when something has changed that would change the available co-authors, like a fetch or a commit.

smashwilson: Looking a little more closely at @maxbrunsfeld's flame graph, one thing that stands out to me is that we're calling `setState()` and running an entire render cycle every time `loadUsersFromLocalRepo` finishes, even though the result is the same. I'm guessing that's the `setState` call in here:

https://github.com/atom/github/blob/df5ef8c48f0fe833772a6212200a5490c09bcdfa/lib/controllers/git-tab-controller.js#L69-L74

Maybe we should also prevent `UserStore.onDidUpdate()` from firing when the user list hasn't actually changed... ?

annthurium: @smashwilson yes, my original solution was for the `UserStore` to not invoke `loadUsersFromLocalRepo` when not needed.

However, I noticed we were also performing a bunch of other unnecessary operations on the repository when the commit message was updated (, and figured it would be a better performance improvement to fix the problem here instead of in the user store.

I did some additional testing and confirmed that the commit button is still disabled if the message is empty, and the remaining characters counter changes as you type in the commit box.  Which makes sense, because the `CommitView` component has its own logic for dealing with commit message updates, that forces a re render when the message has changed.

smashwilson: > Which makes sense, because the CommitView component has its own logic for dealing with commit message updates, that forces a re render when the message has changed.

:sparkles: This was the piece that I wasn't sure was in place or not. Excellent :cake:

PR #1476, opened 2018-05-18
Populate co-authors from mentionable users from the GitHub API

smashwilson:
If the current repository has one or more remotes hosted on a GitHub instance _and_ you've authenticated to that instance on the GitHub tab, fetch the "mentionable" users from the GraphQL API rather than the local git history. This will give us GitHub usernames which we can use for autocompletion, and, importantly, it will omit users who have been blocked by the current user.

Along the way, I'm building out some infrastructure we can use to test components that query GraphQL.

- [x] Acquire fetch query functions for GraphQL endpoints that have already been authenticated within `RelayNetworkLayerManager`.
- [x] When the current repository has at least one GitHub remote with an authentication token:
  - [x] Retrieve mentionable users from that repository via GraphQL instead of pulling them from git history
  - [x] Iterate over multiple pages
  - [x] Handle users without a public email account
- [x] Use an `Author` model now that we have enough data to warrant one
- [x] Restructure `UserStore` storage to pre-sort authors
- [x] Use a `ModelObserver` within the `UserStore` to watch the `Repository` and reload authors when it broadcasts a change
- [x] Pass the `UserStore` through the component tree as a prop. Extract authors with an `<ObserveModel>` component closer to the co-author field.
- [x] Observe the login model and ensure that co-authors are fetched if a token becomes available later
- [x] Clear authors entirely when the Repository changes or when a GitHub remote is added or removed.
- [x] Throttle GraphQL requests so we don't query over and over again when the local Repository changes
- [x] Render and autocomplete GitHub handles when they are available
- [x] Verify OAuth scopes and treat insufficient scopes as "logged out"
- [x] Remove users from the list with `shift-delete`

### DIFF
--- a/lib/models/github-login-model.js
+++ b/lib/models/github-login-model.js
@@ -46,7 +46,7 @@ export default class GithubLoginModel {
       return UNAUTHENTICATED;
     }
 
-    if (/^https?:/.test(account)) {
+    if (/^https?:\/\//.test(account)) {
       // Avoid storing tokens in memory longer than necessary. Let's cache token scope checks by storing a set of
       // checksums instead.
       const hash = crypto.createHash('md5');
@@ -54,17 +54,24 @@ export default class GithubLoginModel {
       const fingerprint = hash.digest('base64');
 
       if (!this.checked.has(fingerprint)) {
-        const scopes = new Set(await this.getScopes(account, password));
+        try {
+          const scopes = new Set(await this.getScopes(account, password));
 
-        for (const scope of this.constructor.REQUIRED_SCOPES) {
-          if (!scopes.has(scope)) {
-            // Token doesn't have enough OAuth scopes, need to reauthenticate
-            return INSUFFICIENT;
+          for (const scope of this.constructor.REQUIRED_SCOPES) {
+            if (!scopes.has(scope)) {
+              // Token doesn't have enough OAuth scopes, need to reauthenticate
+              return INSUFFICIENT;
+            }
           }
-        }
 
-        // We're good
-        this.checked.add(fingerprint);
+          // We're good
+          this.checked.add(fingerprint);
+        } catch (e) {
+          // Bad credential most likely
+          // eslint-disable-next-line no-console
+          console.error(`Unable to validate token scopes against ${account}`, e);
+          return UNAUTHENTICATED;
+        }
       }
     }
 
--- a/lib/models/github-login-model.js
+++ b/lib/models/github-login-model.js
@@ -1,10 +1,15 @@
+import crypto from 'crypto';
 import {Emitter} from 'event-kit';
 
-import {UNAUTHENTICATED, createStrategy} from '../shared/keytar-strategy';
+import {UNAUTHENTICATED, INSUFFICIENT, createStrategy} from '../shared/keytar-strategy';
 
 let instance = null;
 
 export default class GithubLoginModel {
+  // Be sure that we're requesting at least this many scopes on the token we grant through github.atom.io or we'll
+  // give everyone a really frustrating experience ;-)
+  static REQUIRED_SCOPES = ['repo', 'user:email']
+
   static get() {
     if (!instance) {
       instance = new GithubLoginModel();
@@ -16,6 +21,7 @@ export default class GithubLoginModel {
     this._Strategy = Strategy;
     this._strategy = null;
     this.emitter = new Emitter();
+    this.checked = new Set();
   }
 
   async getStrategy() {
@@ -34,11 +40,34 @@ export default class GithubLoginModel {
 
   async getToken(account) {
     const strategy = await this.getStrategy();
-    let password = await strategy.getPassword('atom-github', account);
+    const password = await strategy.getPassword('atom-github', account);
     if (!password) {
       // User is not logged in
-      password = UNAUTHENTICATED;
+      return UNAUTHENTICATED;
     }
+
+    if (/^https?:/.test(account)) {
+      // Avoid storing tokens in memory longer than necessary. Let's cache token scope checks by storing a set of
+      // checksums instead.
+      const hash = crypto.createHash('md5');
+      hash.update(password);
+      const fingerprint = hash.digest('base64');
+
+      if (!this.checked.has(fingerprint)) {
+        const scopes = new Set(await this.getScopes(account, password));
+
+        for (const scope of this.constructor.REQUIRED_SCOPES) {
+          if (!scopes.has(scope)) {
+            // Token doesn't have enough OAuth scopes, need to reauthenticate
+            return INSUFFICIENT;
+          }
+        }
+
+        // We're good
+        this.checked.add(fingerprint);
+      }
+    }
+
     return password;
   }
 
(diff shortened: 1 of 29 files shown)


smashwilson:
I've hit a hiccup now that I'm actually querying GraphQL:

> Your token has not been granted the required scopes to execute this query. The 'email' field requires one of the following scopes: ['user:email', 'read:user'], but your token has only been granted the: ['read:org', 'repo'] scopes. Please modify your token's scopes at: https://github.com/settings/tokens.

We'll need to request the `user:email` scope to be able to get public email addresses from mentionable users, which we need to correlate mentionable users to git authors.

@BinaryMuse, if you get a moment - can you link me to the setup for the OAuth app? How would I go about adding a scope to it, and what does the flow look like for users who are already logged in?

BinaryMuse:
@smashwilson The OAuth app [is setup here](https://github.com/github/atom.io/blob/8aa43ffd33664ba3e990d2210ec69651eaee3e7a/config/initializers/omniauth.rb#L7-L12); the scopes are inline. The flow is similar for already logged-in users if I'm not mistaken: atom.io will bounce them to the dotcom OAuth screen stating that the app needs new permissions. They'll be bounced back to the same atom.io confirmation screen as before.

smashwilson:
:wave: Greetings @dmleong, @benbalter!

This pull request addresses feedback that @dmleong had when we presented the co-author feature at Demo Day - because we were populating the co-author dropdown from local git history, it could include blocked users. After this change, the dropdown is populated by querying the GraphQL API for "mentionable" users within the current repository's GitHub remotes instead, which should exclude blocked users properly.

We still fall back to using local git history if:

* The current repository has no remotes hosted on .com
* The user has not authenticated to GitHub in the GitHub tab

If a user opens Atom on a repository that doesn't satisfy these conditions and later it does (by adding remotes at the command line, for example) it will clear and re-populate accordingly.

I wanted to check in with you to make sure that this was sufficient, and if there was anything else we could do to improve in this area :smile:

dmleong:
Hi @smashwilson thanks for tagging us! Is there any way to remove any co-author suggestions from the list? I could see this being a problem if someone unexpected pops up and you would like to dismiss them from your list.

smashwilson:
> Is there any way to remove any co-author suggestions from the list?

Ah, there is not! That should be easy enough. What I'm thinking now:

* Highlight the entry and shift-delete to add an email address to a list in your Atom config
* The `UserStore` omits entries with email addresses matching those in the config list

The nice thing here is that this list will be respected across all of your Atom projects, even those that aren't backed by .com.

An alternative would be to actually block users on .com when doing this. Would that be better? It feels a bit too aggressive for a keybinding you can trigger by mistake...

lib/models/github-login-model.js
```diff
@@ -54,6 +90,23 @@ export default class GithubLoginModel {
     this.didUpdate();
   }
 
+  async getScopes(host, token) {
+    if (atom.inSpecMode()) {
+      throw new Error('Attempt to check token scopes in specs');
+    }
+
+    const response = await fetch(host, {
+      method: 'HEAD',
```
annthurium: TIL -- I didn't realize `HEAD` is a http method.  (tho I guess there are all kindsa other weird ones like `PURGE`)

lib/models/user-store.js
```diff
@@ -1,79 +1,231 @@
...
-    delete users[NO_REPLY_GITHUB_EMAIL];
-
-    return Object.keys(users)
-      .map(email => ({email, name: this.users[email]}))
-      .sort((a, b) => {
-        if (a.name < b.name) { return -1; }
-        if (a.name > b.name) { return 1; }
-        return 0;
-      });
+    return this.users;
```
annthurium: question...if users haven't created a GitHub access token, do we fall back on local git history?  Or do we not populate the list of users at all?

lib/models/github-login-model.js
```diff
@@ -1,10 +1,15 @@
...
 
-import {UNAUTHENTICATED, createStrategy} from '../shared/keytar-strategy';
+import {UNAUTHENTICATED, INSUFFICIENT, createStrategy} from '../shared/keytar-strategy';
 
 let instance = null;
 
 export default class GithubLoginModel {
+  // Be sure that we're requesting at least this many scopes on the token we grant through github.atom.io or we'll
+  // give everyone a really frustrating experience ;-)
+  static REQUIRED_SCOPES = ['repo', 'read:org', 'user:email']
```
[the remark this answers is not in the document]
smashwilson: > unified Atom login experience

I would love to have this so much. We'd need to find a way to make "GitHub identity" a part of the core API somehow, so that packages could just consume it. Maybe have a separate package that exists purely to handle the UI for authentication...

I'm not sure how much effort it would be. We'd need to scope it out first in an Atom RFC and get buy-in from the core team. Maybe we could wait to see how @shana can improve our authentication story first, and then generalize from there.

smashwilson:
@dmleong Hmm am I doing that wrong?

When I was investigating the GraphQL API, it looked like mentionable users with email address privacy enabled were returned as empty strings:

```
query {
  user(login: "simurai") {
    name
    email
  }
}
```

```
{
  "data": {
    "user": {
      "name": "simurai",
      "email": ""
    }
  }
}
```

Re-reading the [commit email address docs](https://help.github.com/articles/about-commit-email-addresses/), it looks like I _should_ be prepending the user ID, to correlate across username changes... although I can't find a way to pull the right user ID from the GraphQL API yet. But ID-based anonymized email addresses still leak the user login, right... ?

> Note: If you created your GitHub account after July 18, 2017, your GitHub-provided no-reply email address is a seven-digit ID number and your username in the form of ID+username@users.noreply.github.com.

I suppose I'm not sure what you mean by "private account" and the best way to prevent leaking them 😅

smashwilson:
> Is the intention of using email addresses to find others who aren't in the mentionable users GraphQL API or is it to verify that they are mentionable users?

We're using email addresses to:

* Construct [co-author trailers](https://help.github.com/articles/creating-a-commit-with-multiple-authors/) on your commits.
* Generate `https://avatars.githubusercontent.com/u/e?email=` URLs to [display user avatars](https://github.com/atom/github/blob/1ba6e5391ab0570c45078dd79e95ec8346875f31/lib/views/recent-commits-view.js#L45-L62).

If we query the GraphQL API, the co-author dropdown will show _no_ users who are not returned by the GraphQL query - we use the mentionable user list as the source of truth. We will _only_ include users from your git history if we don't have a way to get to GraphQL (and we clear them out if that changes, by logging in for example).

> If the users are already returned via GraphQL as mentionable, I think we are okay to proceed without email verification.

Okay, good 👍 Especially because I don't see a way to distinguish between verified and non-verified addresses in the API, hah. I thought that an account's primary email had to be verified regardless?

> Regarding a fallback using email addresses, I would prefer we only use public, verified email addresses to look up the user.

The only looking up we do is for avatar images, which uses whatever email is present in the git data. In the case of commits containing co-author trailers built with this UI, that's done with emails returned from GraphQL's user object, which are either public and verified or `login@users.noreply.github.com`.

We don't stop users from manually adding private email addresses here:

screen shot 2018-05-29 at 3 17 06 pm

... or by setting git's `user.name` and `user.email`, although at least the push will be rejected then 🤔 

If we _do_ have private email addresses in your git history, we'll leak them to someone sniffing your network traffic in the form of HTTP requests to `https://avatars.githubusercontent.com/u/e?email=...`. I'm afraid I don't know if we can do anything about that because we don't have a good way to identify the address as private.

I think you could also expose a co-author's private email address within git history by mistake if you enter it yourself. Does [the "block pushes" setting](https://blog.github.com/2017-04-11-private-emails-now-more-private/) apply to co-author trailers, too... ? Otherwise I don't know how to detect that, either. The help docs around co-authors imply that this is the committer's responsibility when composing the message:

> If a person chooses to keep their email address private, you should use their GitHub-provided no-reply email to protect their privacy. Otherwise, the co-author's email will be available to the public in the commit message. If you want to keep your email private, you can choose to use a GitHub-provided no-reply email for Git operations and ask other co-authors to list your no-reply email in commit trailers.

dmleong:
@smashwilson thank you for the thorough explanation! 

Let's go ahead and proceed with what you had. It looks like since it's git data, there's not a whole lot of protection we can use (since git is inherently built this way) and we can use commit signing as a form of security on the end user's side.

smashwilson:
@dmleong 🙇Thanks for your input and for bearing with my long-windedness 😄 Co-author exclusion is in; authors can be excluded by pressing `shift-delete` while an author is highlighted in the select list, or by adding email addresses to the package configuration.

Issue #1490, opened 2018-05-30
Uncaught TypeError: Cannot read property 'rememberFocus' of null


sguthals:
[Enter steps to reproduce:]

1. Made Changes
2. Staged Commits
3. Added @smashwilson as a co-author (wasn't auto-filled)
4. Clicked "Add Co-Author"
5. Error Appears
6. Co-author still listed there
7. Committed changes
8. Co-author still populating the co-author space
9. Didn't actually commit - but that might be due to the auth (https://github.com/github/atom.io/pull/1249)

**Atom**: 1.28.0-beta1 x64
**Electron**: 2.0.1
**OS**: Mac OS X 10.13.4
**Thrown From**: [github](https://github.com/atom/github) package 0.15.2

### Stack Trace

Uncaught TypeError: Cannot read property 'rememberFocus' of null

```
At /Users/sarahguthals/.atom/packages/github/lib/controllers/git-tab-controller.js:315

TypeError: Cannot read property 'rememberFocus' of null
    at GitTabController.rememberLastFocus (/packages/github/lib/controllers/git-tab-controller.js:315:35)
    at AutosizeInput.focus (/packages/github/node_modules/react-input-autosize/lib/AutosizeInput.js:173:15)
    at Select.focus (/packages/github/node_modules/react-select/lib/Select.js:278:15)
    at refCoAuthorSelect.map.c (/packages/github/lib/views/commit-view.js:352:41)
    at RefHolder.map (/packages/github/lib/models/ref-holder.js:70:14)
    at CommitView.setState (/packages/github/lib/views/commit-view.js:352:30)
    at Rf (/packages/github/node_modules/react-dom/cjs/react-dom.production.min.js:143:60)
    at Sf (/packages/github/node_modules/react-dom/cjs/react-dom.production.min.js:144:304)
    at Xh (/packages/github/node_modules/react-dom/cjs/react-dom.production.min.js:219:152)
    at Rh (/packages/github/node_modules/react-dom/cjs/react-dom.production.min.js:211:414)
    at Uh (/packages/github/node_modules/react-dom/cjs/react-dom.production.min.js:210:278)
    at Sh (/packages/github/node_modules/react-dom/cjs/react-dom.production.min.js:210:113)
    at interactiveUpdates (/packages/github/node_modules/react-dom/cjs/react-dom.production.min.js:225:398)
    at Jd (/packages/github/node_modules/react-dom/cjs/react-dom.production.min.js:89:113)
```

### Commands

```
[...]

Issue #1403, opened 2018-04-20
Ref soup


smashwilson:
We use a lot of refs in our JSX, which feels un-React-y.

There are a number of reasons for this:

1. The package used to use [Etch](https://github.com/atom/etch) exclusively. Etch makes it much easier to operate imperatively on `this.element`. As different components were ported to React, it was often more convenient to use refs to defer re-architecting the way that some things were written.
2. Before we adopted Enzyme, refs were the most convenient way to access children within test cases. Some tests are still written in the "descending through a ref tree" style to access things they need to.
3. We provide a number of components to declaratively interact with the Atom API (`Decoration`, `Command`, `PaneItem`). The Atom API accepts DOM elements as "items," so often we need to use a ref to pass one.
4. We have a (fairly clunky) mechanism to manage preserving focus as tabs as hidden and shown. Because focus is an imperative DOM API, we need refs to test for and operate on element focus.

Refs that exist from categories (1) and (2) should be relatively easy to eliminate. I suspect, but have not confirmed, that we could tidy up quite a few of them.

I suspect that we can do better on (3) and (4) as well with alternate designs... but I'm not sure what those might look like. Brainstorming is welcome :smile:

smashwilson:
For (4), what I'd like to build is something like this:

```jsx
<SomeParentComponent>
  <PreserveFocus>
     <div>
       <!-- more layers of composition -->
        <Focusable key="something">
           <button>Hooray</button>
         </Focusable>
      </div>
  </PreserveFocus>
</SomeParentComponent>
```

PR #1492, opened 2018-05-30
Band-aids for focus management code

smashwilson:
Until we can tackle #1403 and improve our approach to remembering focus within the git and GitHub tabs, let's add quick guards against trying to call methods on empty refs.


smashwilson:
@sguthals would you mind verifying that this fixes #1490 for you?

1. cd into your atom/github clone
2. `git fetch && git checkout aw/ref-holder`
3. `atom-beta -d` to open Atom

PR #1503, opened 2018-06-01
RFC for Issueish List View

smashwilson:
:stars: _This is based heavily on some excellent internal design ideation by @simurai._ :stars:

I've tried to identify a subset of his vision we could reasonably ship within the next month. My goals here are to rework the GitHub panel to explore an alternate navigation flow (accordion list :point_right: detail panes) and to set the stage just enough to be able to start on pull request reviews. This will likely not be the final form of "the first pane you see," because we'll probably have at least one level of navigation above it - but we can explore that in follow-on work.

**[:art: rendered :art:](https://github.com/atom/github/blob/aw/rfc-pr-list/docs/rfcs/004-issueish-list.md)**

/cc @donokuda 
/cc @sguthals 
/cc @shana

### DIFF
--- a/docs/rfcs/004-issueish-list.md
+++ b/docs/rfcs/004-issueish-list.md
@@ -61,7 +61,7 @@ For a pull request, the issueish pane shows:
 * Reaction emoji and counts.
 * Details and links for all status checks.
 
-_TBD: Issueish pane item design_
+![pane](https://user-images.githubusercontent.com/378023/41007657-cfc8d3c0-6961-11e8-9bf7-7d5b7353199c.png)
 
 ## Drawbacks
 
--- a/docs/rfcs/004-issueish-list.md
+++ b/docs/rfcs/004-issueish-list.md
@@ -42,7 +42,7 @@ Each list item renders a tile containing a compact set of information about that
 
 ![list item](https://user-images.githubusercontent.com/378023/41136622-1102db54-6b12-11e8-8b9b-49ecc45ac98f.png)
 
-Clicking on a list item opens an issueish pane item for the chosen issueish in the same pane container as the parent component (by default, the right dock). If the issuish pane item is already open, it is activated instead.
+Clicking on a list item opens an issueish pane item for the chosen issueish. If the issueish pane item is already open, it is activated instead.
 
 ### Issueish Pane Item: Pull Request
 
--- a/docs/rfcs/004-issueish-list.md
+++ b/docs/rfcs/004-issueish-list.md
@@ -61,7 +61,7 @@ For a pull request, the issueish pane shows:
 * Reaction emoji and counts.
 * Details and links for all status checks.
 
-![pane](https://user-images.githubusercontent.com/378023/41007657-cfc8d3c0-6961-11e8-9bf7-7d5b7353199c.png)
+![pane](https://user-images.githubusercontent.com/378023/41007709-0f9ba9a0-6962-11e8-8b2f-bf6aee8cf8fc.png)
 
 ## Drawbacks
 
--- a/docs/rfcs/004-issueish-list.md
+++ b/docs/rfcs/004-issueish-list.md
@@ -50,18 +50,24 @@ For a pull request, the issueish pane shows:
 
 * Author avatar
 * Title
+* Link to .com. -> [atom/github#1503](https://github.com/atom/github/pull/1503)
+* PR status badge. -> `Open`.
+* Branches -> `master` < `aw/rfc-pr-list`
+* `Commits` with count, links to .com (for now)
+* `Checks` with count, links to .com (for now)
+* `Files changed` with count, links to .com (for now)
+  * Plus a "Checkout" button to fetch (if necessary) and check out the pull request. Only enabled if the current pull request is not the current one.
+* `Conversation` with comment count.
+  * Reaction emoji and counts.
+  * Full description as rendered markdown.
+  * Comments as rendered markdown.
 * Controls for actions:
-  * "Checkout" to fetch (if necessary) and check out the pull request. Only enabled if the current pull request is not the current one.
-  * "Browse" to launch a browser on the corresponding GitHub page.
-  * "Merge" to merge the pull request on GitHub if it is open.
+  * Mergability status -> `Able to merge`, links to the [Merging controls at the bottom](https://github.com/atom/github/pull/1503#partial-pull-merging)
+  * "Merge PR" to merge the pull request on GitHub if it is open.
   * "Close" to close the pull request, unmerged, if it is open.
-  * "Re-open" to re-open a pull request if it is closed.
-* Pull request overview details: commit count, files changed.
-* Full description as rendered markdown.
-* Reaction emoji and counts.
-* Details and links for all status checks.
+  * "Re-open PR" to re-open a pull request if it is closed.
 
-![pane](https://user-images.githubusercontent.com/378023/41007709-0f9ba9a0-6962-11e8-8b2f-bf6aee8cf8fc.png)
+![pane](https://user-images.githubusercontent.com/378023/41033008-0900b4ec-69c0-11e8-9def-68e6f5b40901.png)
 
 ## Drawbacks
 
--- a/docs/rfcs/004-issueish-list.md
+++ b/docs/rfcs/004-issueish-list.md
@@ -22,7 +22,7 @@ As an initial building block toward a pull request review workflow.
 
 Within the GitHub panel, render a vertical stack of two collapsible lists of _issueish_ (pull request or issue) items:
 
-_First list: current pull request_. If the active branch is associated with one or more open pull requests on the GitHub repository, render an item for each.
+_First list: current pull request_. If the active branch is associated with one or more open pull requests on a GitHub repository, render an item for each. "Associated with" means that the pull request's head ref and head repository matches the upstream remote ref for the current branch in the active git repository.
 
 _Second list: all open pull requests_. List all open pull requests on the GitHub repository.
 
--- a/docs/rfcs/004-issueish-list.md
+++ b/docs/rfcs/004-issueish-list.md
@@ -72,7 +72,7 @@ For a pull request, the issueish pane shows:
 
 If no current PR can be found, an "open new pull request" button is shown. If needed it also offers to "Publish" or "Push".
 
-![new pr](https://user-images.githubusercontent.com/378023/41100932-9a3cd7be-6a9d-11e8-85c4-c995b71548c7.png)
+![new pr](https://user-images.githubusercontent.com/378023/41136021-33249004-6b0f-11e8-9cf0-08bf4a9a2767.png)
 
 When the current branch is the default branch, e.g. `master`, a message is shown that suggests to "Create a new branch".
 
--- a/docs/rfcs/004-issueish-list.md
+++ b/docs/rfcs/004-issueish-list.md
@@ -99,8 +99,8 @@ With that said, the choices for the specific lists we show are a bit arbitrary. 
 
 ### Before RFC merge:
 
-- [ ] What else from the existing issueish pane should we keep? Comments, timeline events?
-- [ ] Are there other pull request actions it would be useful to support?
+- [x] What else from the existing issueish pane should we keep? Comments, timeline events?
+- [x] Are there other pull request actions it would be useful to support?
 
 ### Out of scope:
 
--- a/docs/rfcs/004-issueish-list.md
+++ b/docs/rfcs/004-issueish-list.md
@@ -24,7 +24,7 @@ Within the GitHub panel, render a vertical stack of two collapsible lists of _is
 
 _First list: current pull request_. If the active branch is associated with one or more open pull requests on a GitHub repository, render an item for each. "Associated with" means that the pull request's head ref and head repository matches the upstream remote ref for the current branch in the active git repository.
 
-_Second list: all open pull requests_. List all open pull requests on the GitHub repository.
+_Second list: all open pull requests_. List all open pull requests on the GitHub repository, ordered by decreasing creation date.
 
 Each list has a "collapse arrow" in its header. Clicking the collapse arrow toggles the visibility of that list's items, accordion-style.
 
--- a/docs/rfcs/004-issueish-list.md
+++ b/docs/rfcs/004-issueish-list.md
@@ -78,7 +78,8 @@ When the current branch is the default branch, e.g. `master`, a message is shown
 
 ## Drawbacks
 
-None!
+* "All pull requests" could easily be overwhelming on moderate to high traffic repositories. Stay tuned for more refinements on this front.
+* Opening a pane item for each pull request click is heavyweight from a navigational standpoint. We may explore showing a popup as an intermediate state.
 
 ## Rationale and alternatives
 
--- a/docs/rfcs/004-issueish-list.md
+++ b/docs/rfcs/004-issueish-list.md
@@ -106,6 +106,7 @@ With that said, the choices for the specific lists we show are a bit arbitrary. 
 
 - [ ] How can we allow a user to customize the lists?
 - [ ] How can we notify a user about updated activity on a visible PR?
+- [ ] Where should you be able to merge, close, or re-open pull requests?
 
 ## Implementation phases
 
--- a/docs/rfcs/004-issueish-list.md
+++ b/docs/rfcs/004-issueish-list.md
@@ -1,0 +1,98 @@
+# Issueish List
+
+## Status
+
+Proposed
+
+## Summary
+
+Display a list of all open pull requests in the current repository in the GitHub tab.
+
+## Motivation
+
+To provide a navigational element that makes sense even if you aren't on an active feature branch.
+
+To give users a way to see an overview of what's going on in the repository.
+
+As an initial building block toward a pull request review workflow.
+
+## Explanation
+
+### Accordion Lists
+
+Within the GitHub panel, render a vertical stack of two collapsible lists of _issueish_ (pull request or issue) items:
+
+_First list: current pull request_. If the active branch is associated with one or more open pull requests on the GitHub repository, render an item for each.
+
+_Second list: all open pull requests_. List all open pull requests on the GitHub repository.
+
+Each list has a "collapse arrow" in its header. Clicking the collapse arrow toggles the visibility of that list's items, accordion-style.
+
+If either list exceeds 20 items, truncate the list and render a "More" link after its final item. Clicking "more" opens the corresponding search on GitHub.
+
+Each list item renders a tile containing a compact set of information about that pull request:
+
+* Mini author avatar
+* Title, truncated if necessary
+* Terse relative timestamp (1d, 2h, 30m)
+* Pull request state: open, closed, merged
+* Status check summary
+
+_TBD: Tile design_
+
+Clicking on a list item opens an issueish pane item for the chosen issueish in the same pane container as the parent component (by default, the right dock). If the issuish pane item is already open, it is activated instead.
+
+### Issueish Pane Item: Pull Request
+
+For a pull request, the issueish pane shows:
+
+* Author avatar
+* Title
+* Controls for actions:
+  * "Checkout" to fetch (if necessary) and check out the pull request. Only enabled if the current pull request is not the current one.
+  * "Browse" to launch a browser on the corresponding GitHub page.
+  * "Merge" to merge the pull request on GitHub if it is open.
+  * "Close" to close the pull request, unmerged, if it is open.
+  * "Re-open" to re-open a pull request if it is closed.
+* Pull request overview details: commit count, files changed.
+* Full description as rendered markdown.
+* Reaction emoji and counts.
+* Details and links for all status checks.
+
+_TBD: Issueish pane item design_
+
+## Drawbacks
+
+This does not offer a mechanism to create _new_ pull requests, which we have now. We also lose timeline events on the pull request.
+
+## Rationale and alternatives
+
+Our current GitHub panel focuses on showing you stuff about _the pull request that's associated with your current branch._ The problem is, it's difficult to unambiguously determine that in the general case.
+
+The first thing you see today when you open the GitHub panel on the `master` branch of an active repository is not very helpful:
+
+![wat](https://user-images.githubusercontent.com/17565/40857603-99b92304-65a9-11e8-986e-0f14290bda8a.png)
+
+This is a list of _all pull requests on GitHub that have a head ref called "master", from any head repository_. You can then "pin" any of them to see that pull request's details. This isn't useful on master and it's unclear to users what this is supposed to accomplish. Pinning was intended to be an infrequent edge case when we couldn't find the right PR for a given ref, not the first interaction you have with the package.
+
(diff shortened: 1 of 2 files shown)


docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,98 @@
...
+
+If either list exceeds 20 items, truncate the list and render a "More" link after its final item. Clicking "more" opens the corresponding search on GitHub.
+
+Each list item renders a tile containing a compact set of information about that pull request:
+
+* Mini author avatar
+* Title, truncated if necessary
+* Terse relative timestamp (1d, 2h, 30m)
+* Pull request state: open, closed, merged
+* Status check summary
```
smashwilson: ☝️ For this I was thinking a "mini donut view" might be cool. It'd be pretty small and it'd still let you tell the difference between "everything failed," "some stuff was green," and "all good to go".

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,98 @@
...
+
+_TBD: Tile design_
+
+Clicking on a list item opens an issueish pane item for the chosen issueish in the same pane container as the parent component (by default, the right dock). If the issuish pane item is already open, it is activated instead.
```
smashwilson: Like @simurai and I were talking about earlier, there are a few directions we could go here. To list a few:

1. Open a "detail view" in a new PaneItem.
   * ...in the same dock as the PR list (how I have it written here)
   * ...in the workspace center (how the PR pane items work today)
   * ...in the workspace center as a pending item
2. Implement a balloon/tooltip view to host the "details."
3. Navigate to a "detail view" in-place within the existing PaneItem. Provide some kind of breadcrumb on the detail view to navigate back to the "list view".

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,98 @@
...
+For a pull request, the issueish pane shows:
+
+* Author avatar
+* Title
+* Controls for actions:
+  * "Checkout" to fetch (if necessary) and check out the pull request. Only enabled if the current pull request is not the current one.
+  * "Browse" to launch a browser on the corresponding GitHub page.
+  * "Merge" to merge the pull request on GitHub if it is open.
+  * "Close" to close the pull request, unmerged, if it is open.
+  * "Re-open" to re-open a pull request if it is closed.
```
smashwilson: This is a cool opportunity to finally support some of these. I think these actions are what will make those issueish items _useful_.

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,98 @@
...
+* Pull request overview details: commit count, files changed.
```
smashwilson: Future launching-off points to navigable multi-file patch items. Stay tuned for more RFCs :notes: 📻

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,98 @@
...
+* Full description as rendered markdown.
+* Reaction emoji and counts.
+* Details and links for all status checks.
+
+_TBD: Issueish pane item design_
+
+## Drawbacks
+
+This does not offer a mechanism to create _new_ pull requests, which we have now. We also lose timeline events on the pull request.
```
smashwilson: This would actually suck to lose, I use "publish pull request" all the time. @simurai, @donokuda, any thoughts on how we could work this in - preferably without ballooning our scope?

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,98 @@
...
+Showing a PR list instead provides a uniform, more easily understood entry point to the package's GitHub functionality, and paves the way to other pull-request-focused activities in the future. "All open PRs" seems like a reasonable starting point, and "current PRs" preserves the ability to take advantage of your local editor context.
+
+With that said, the choices for the specific lists we show are a bit arbitrary. We'll need to research and iterate on them quite a bit to find what's most useful for the most people, but for now we need to start with something.
+
+## Unresolved questions
+
+### Before RFC merge:
+
+- [ ] How can we still offer "push/publish+new pull request"?
+- [ ] What else from the existing issueish pane should we keep? Comments, timeline events?
```
smashwilson: The main reason I'd consider taking these out is because they don't really add anything to just... looking at the PR on .com.

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,98 @@
...
+Clicking on a list item opens an issueish pane item for the chosen issueish in the same pane container as the parent component (by default, the right dock). If the issuish pane item is already open, it is activated instead.
+
+### Issueish Pane Item: Pull Request
+
+For a pull request, the issueish pane shows:
+
+* Author avatar
+* Title
+* Controls for actions:
+  * "Checkout" to fetch (if necessary) and check out the pull request. Only enabled if the current pull request is not the current one.
```
smashwilson: "Checkout" for consistent git terminology. Even though it technically fetches first too maybe.

A few edge cases:

* What if your working copy is dirty? Do we use worktrees? Disable the button? Supporting worktrees properly would be rad, but also pretty complicated.
* What if you already have the pull request checked out? What if it isn't up to date? Do we automatically pull for you? Maybe only if it works with `--ff-only`?

docs/rfcs/004-issueish-list.md
simurai: A donut could also only be used for mixed states (fail and success). For "everything failed/succeeded" we could still use the x and checkmark. Then you don't have to look too closely if the donut contains any red/green parts.

image

Or maybe a donut with a x or checkmark.

docs/rfcs/004-issueish-list.md
simurai: This is super tough to decide.

1. Something that bothers me with the current FilePatch pane. After committing having to manually close the empty pane. I know that I could just keep it open and it might be re-used, but somehow I always feel the urge to "clean up" and only show what's really needed. Are you also suffering from this disease? The big advantage would be that you're free to move the pane around and maybe even have multiple PR panes open. Even side by side if needed.
2. A balloon/tooltip is great for quick "in and out", but you can't keep it open to use as a reference because it auto-closes once the focus moves to the buffer.
3. This is somewhere in between, no clean-up necessary and you can keep it open as long as needed. Downside: It doesn't let you quickly jump between different detail views and you always have to click "back" first. Apple's HIG [discourages](https://developer.apple.com/macos/human-interface-guidelines/windows-and-views/drawers/) of using this approach. Plus you're limited to the narrow width of a dock item.

Any favorite? Hmm.... :thinking: I don't yet. Maybe best to first mock them up to compare or even do a static prototype.

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,98 @@
...
+  * "Browse" to launch a browser on the corresponding GitHub page.
+  * "Merge" to merge the pull request on GitHub if it is open.
+  * "Close" to close the pull request, unmerged, if it is open.
+  * "Re-open" to re-open a pull request if it is closed.
+* Pull request overview details: commit count, files changed.
+* Full description as rendered markdown.
```
simurai: Since the full description could be pretty long, we could have max height with the option to "show more". Or have the full description at the "end" so you don't have to scroll down too much to see the status checks.

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,98 @@
...
+
+As an initial building block toward a pull request review workflow.
+
+## Explanation
+
+### Accordion Lists
+
+Within the GitHub panel, render a vertical stack of two collapsible lists of _issueish_ (pull request or issue) items:
+
+_First list: current pull request_. If the active branch is associated with one or more open pull requests on the GitHub repository, render an item for each.
```
simurai: In what case would it show more than one PR? Is if someone forks a repo, then makes a PR that has the same branch name as a PR that is on the upstream repo? Like

- `atom/atom:master` <- `atom/atom:feature`
- `atom/atom:master` <- `username/atom:feature`

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,98 @@
...
+Each list has a "collapse arrow" in its header. Clicking the collapse arrow toggles the visibility of that list's items, accordion-style.
+
+If either list exceeds 20 items, truncate the list and render a "More" link after its final item. Clicking "more" opens the corresponding search on GitHub.
+
+Each list item renders a tile containing a compact set of information about that pull request:
+
+* Mini author avatar
+* Title, truncated if necessary
+* Terse relative timestamp (1d, 2h, 30m)
+* Pull request state: open, closed, merged
```
simurai: If the 2nd list shows only "open" PRs, then we probably don't need to show the state also in each list item. Maybe the title can say "Open PRs" to make it clear. And on the right the repo as a confirmation:

image

docs/rfcs/004-issueish-list.md
smashwilson: Yah, really good point. I'm thinking moving these to the bottom is probably a good idea, so you have to pass your eyes over all the PR state on the way to Doing A Thing.

docs/rfcs/004-issueish-list.md
smashwilson: >  Is if someone forks use the same branch name and make a PR to your origin?

Yeah, that's the case where we show multiple now, like all of the "master" PRs it offers to let you pin.

> Couldn't this be reduced to just one if we would limit it to the current remote repo? Then there can only be one branch with the same name.

Which remote is the "current" one? Right now we just ask you for the "upstream" one.

This gets tricky when you have a network of multiple forks involved. I think you can send multiple pull requests from the same head ref to different base repositories in the same fork network. You might even be able to open multiple pull requests from the same head ref to different base refs within the same repository... ? I haven't tried it yet.

In any case, the by-far most common case is to have zero or one pull requests for a given head ref and head repository, but we've got a long tail of wacky edge cases where you _could_ have several. I figured that because we're already building a view based on stacked lists, we could handle them _and_ make it more consistent just by making the "current" section a list, too.

docs/rfcs/004-issueish-list.md
smashwilson: Is there a way to generalize this... ? Like, if every entry in a given list has the same value for some attribute, we omit it from the item display. Or if that attribute is locked to a single value by the query maybe, like if the search is `repo:atom/github is:pr is:merged`, we could infer that the status is always `merged`.

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,100 @@
...
+Clicking on a list item opens an issueish pane item for the chosen issueish in the same pane container as the parent component (by default, the right dock). If the issuish pane item is already open, it is activated instead.
+
+### Issueish Pane Item: Pull Request
+
+For a pull request, the issueish pane shows:
+
+* Author avatar
+* Title
+* Controls for actions:
+  * "Checkout" to fetch (if necessary) and check out the pull request. Only enabled if the current pull request is not the current one.
```
Arcanemagus: > Only enabled if the current pull request is not the current one.

This sentence is recursive. Did you mean something like this?

```md
Only enabled if the current checked out branch is not the branch of the pull request.
```

If so I would go with "visible" myself instead of enabled, but if you don't want the UI "jumping" on users enabled is better.

docs/rfcs/004-issueish-list.md
simurai: > Could we make the description the bit that scrolls?

If we move the merge/close buttons to the bottom, having the description/comments scrollable, would be a good idea. 👍

An alternative to "Show more" for very long descriptions could be a "jump to new comments", in case we wanted to add some sort of read/new notification.

docs/rfcs/004-issueish-list.md
simurai: > Which remote is the "current" one?

The one that it would use to pull/push. I think it's called "remote tracking branch". But I didn't consider that a local branch could have **no** remote tracking branch. Hmm.. maybe the title could say

- "Current PR" if there is just one
- "Related PRs" if there are more

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,106 @@
...
+* Author avatar
+* Title
+* Link to .com. -> [atom/github#1503](https://github.com/atom/github/pull/1503)
+* PR status badge. -> `Open`.
+* Branches -> `master` < `aw/rfc-pr-list`
+* `Commits` with count, links to .com (for now)
+* `Checks` with count, links to .com (for now)
+* `Files changed` with count, links to .com (for now)
+  * Plus a "Checkout" button to fetch (if necessary) and check out the pull request. Only enabled if the current pull request is not the current one.
+* `Conversation` with comment count.
```
simurai: When making the mockup I had a deja-vu 😄 and sure enough, there is a closed PR that included the conversation and comments in a panel (now dock item): https://github.com/atom/github/pull/494

It was ok for small PRs. But for PRs that had a long description and lots of long comments, it just felt too cramped and you scroll forever. Here a gif that tries to simulate it with this PR: https://github.com/atom/github/pull/1370

timeline

We then concluded that it's better to open a center pane for the conversation. So not sure if we should try again having the conversation in a dock item? An alternative might be to **only** have the description, but not the comments. If you want the read the whole timeline with comments, click on `Conversation` and it links to .com.

simurai:
Here a slightly different proposal. I'll make first a comment before editing. Note: The "Reviews" are already included just to show how they could be integrated, but they won't be part of this RFC.

### 1. Preview popover

In order to quickly get a preview with more details about a PR, when clicking on a "Open PRs" list item, a popover shows up:

popover

From the popover, you can click on "Conversation" which opens a center pane with the PR description and timeline that is currently already implemented.

prs

Clicking on the rest of the popover items opens the page on .com for now. In the future they might also open a center pane.

image

### 2. Checkout

Clicking on the "Checkout" button, checks out that branch. The popover closes and the PR list changes to the "Current PR":

prs 2

It's similar to the popover, but it's more permanent and doesn't disappear when focus moves away. Clicking on "< PRs" at the top, switches back to the PR lists again.

docs/rfcs/004-issueish-list.md
smashwilson: Oh, right - I can get that remote from your branch information. Well then :smile:

docs/rfcs/004-issueish-list.md
simurai: I updated the RFC to keep the existing center pane when clicking on "Conversation". I'm not sure how valuable it is, but since it's already built, we can keep it for now?

Other things from the current GitHub panel:

- Labels? Might still be ok to keep them?
- The detailed CI status?
  image
  It takes up quite some space and I'm not sure how valuable it is to see which one is failing. In the current mockup there is the "Checks", but I think that's a new feature that not every repo uses yet. Then the "Able to merge" would include CI, but also if a branch has conflicts or there if still a review is needed. If any of these fail, you would have to click on it and see on .com what it is and take further actions.

docs/rfcs/004-issueish-list.md
smashwilson: > If so I would go with "visible" myself instead of enabled, but if you don't want the UI "jumping" on users enabled is better.

Yeah, my general preference is to avoid hiding temporarily inapplicable controls - if you don't see a button, you have no idea you _could_ check out a PR, but if the button is there and disabled, you know that you can check out some PRs but not this one. (You can even add an explanation why on the tooltip).

Not a hard-and-fast rule, though. For example, we should only show "Reopen" on a closed PR, because we'll always be able to show some kind of "PR action button cluster" in the same place and that's enough of a cue for that functionality.

Also thanks for catching my tangled wording there :+1: 🙇

docs/rfcs/004-issueish-list.md
smashwilson: Hahaha! Oops.

> We then concluded that it's better to open a center pane for the conversation.

👍

docs/rfcs/004-issueish-list.md
smashwilson: > I updated the RFC to keep the existing center pane when clicking on "Conversation". I'm not sure how valuable it is, but since it's already built, we can keep it for now?

:+1: I like it for:

* Seeing the full conversation without Massive Endless Cramped Scrolling
* Detailed checks and build links
* Being able to drag and drop around as its own thing, so you can open more than one at once - like to keep an eye on one PR to see if it's green while you're working on another one

> It takes up quite some space and I'm not sure how valuable it is to see which one is failing.

One of my most-used features, personally 😁 It's a great way to keep an eye on builds while continuing to work, and to jump right to the problem when there is one.

Maybe we could come up with a more compact representation (eventually, if not here)?

simurai:
> How would you feel about adding an "open" button to the popup balloon that opens the "full pane item" view of that PR, with the full conversation and the other stuff that's visible in the popup (commit count, status checks, etc)?

Yeah, I was thinking the same thing. But not sure.. :smile: I was more thinking we can keep the current conversation center pane, since it's already built, but not invest more time on it by adding the popover content. Because I wonder how many people actually use it vs. just open the PR page on .com that is fully featured.

What about this:

1. Phase: Make list items open as new dock items (like it currently is proposed).
2. Phase: Consider adding a popover and maybe adding the popover content also to the center pane.

With a dock item, you could split it to the left and kinda simulate like it would be part of the center pane.

split 2

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,112 @@
...
+
+If either list exceeds 20 items, truncate the list and render a "More" link after its final item. Clicking "more" opens the corresponding search on GitHub.
+
+![list](https://user-images.githubusercontent.com/378023/41136538-ad5c461c-6b11-11e8-9e1d-e4a674f628cd.png)
+
+Each list item renders a tile containing a compact set of information about that pull request:
+
+* Mini author avatar
+* Title, truncated if necessary
+* PR number (`#1503`)
```
simurai: Apparently in some teams, they use the PR number/ID to refer to a PR. Therefore it should be helpful seeing it in the list so you can search by number/ID and not just title.

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,112 @@
...
+* Author avatar
+* Title
+* Branches -> `master` < `aw/rfc-pr-list`
+* "Checkout" button to fetch (if necessary) and check out the pull request. Only enabled if the current pull request is not the current one.
+* `Commits` with count, links to .com (for now), optional with avatars
+* `Checks` with count, links to .com (for now)
+  * CI status, each item links to the detail page
+* `Files changed` with count, links to .com (for now), optional with "+-" bar
+* Mergability status -> `Able to merge`, links to the [Merging controls at the bottom](https://github.com/atom/github/pull/1503#partial-pull-merging)
+  * "Merge PR" to merge the pull request on GitHub if it is open.
```
simurai: In our design review, it was suggested to not show the close and merge button. Just because there isn't enough information and context to make a good decision. It might still be handy for people that mostly merge their own PRs, but it could also lead to "accidents".

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,112 @@
...
+To give users a way to see an overview of what's going on in the repository.
+
+As an initial building block toward a pull request review workflow.
+
+## Explanation
+
+### Accordion Lists
+
+Within the GitHub panel, render a vertical stack of two collapsible lists of _issueish_ (pull request or issue) items:
+
```
annthurium: reading this, it wasn't immediately clear to me that the number was "time."  I saw "3M" and thought "3 million"?  So can we clarify that label somehow?  Also is this the date that the pull request was opened, or the date of the first commit in that pull request?

docs/rfcs/004-issueish-list.md
```diff
@@ -0,0 +1,112 @@
...
+- [ ] Are there other pull request actions it would be useful to support?
+
+### Out of scope:
+
+- [ ] How can we allow a user to customize the lists?
+- [ ] How can we notify a user about updated activity on a visible PR?
+
+## Implementation phases
+
+1. Accordion list infrastructure: search model, collapsible list component.
```
annthurium: are you going to make a generic "list" component that could be re-used elsewhere in the UI?  It seems like we have a lot of "one off" components, and it might make sense to think about what could be made generic.  (Ideally, I'd love a React UI components library for Atom that provides generic modals and lists and forms and other stuff, but one thing at a time.)

docs/rfcs/004-issueish-list.md
Arcanemagus: I thought the same thing, and was going to comment on it before realizing it is using `M` to signify months since `m` would be taken by minutes.

annthurium: yeah, ideally we could use "minutes", "weeks", "etc" instead of abbreviations.  Moment.js has a "humanize time since" function that provides this in an i18n-friendly manner.  https://momentjs.com/docs/#/durations/humanize/

docs/rfcs/004-issueish-list.md
smashwilson: > Are you going to make a generic "list" component that could be re-used elsewhere in the UI?

Yah, I was mentally planning on having an `AccordionList` component.

> It seems like we have a lot of "one off" components, and it might make sense to think about what could be made generic.

This is true, we don't as of yet have many reusable building block-style components. Does anything come to mind for abstractions we could introduce to consolidate existing ones? That would be a cool way to improve the codebase 🤔 

> Ideally, I'd love a React UI components library for Atom that provides generic modals and lists and forms and other stuff, but one thing at a time.

I've had Secret Plans :tm: to extract our [Atom API integration components](https://github.com/atom/github/tree/master/lib/atom) to their own npm package for a while. We could build on that :smile:

simurai: 👍 I would also love if Atom would provide UI components beyond the basic ones from the [styleguide](https://github.com/atom/styleguide). Then not every package author has to create their own from scratch.

> having an `AccordionList` component

A while (very long time) ago, I made an accordion component: https://github.com/atom/atom-ui/pull/14

accordion

It uses the  `<details>` and `<summary>` HTML elements to provide the collapsing functionality. Not sure if we wanted to use that. Or just initially?

simurai:
Taking a step back, can you imagine using this proposed PR list? I can think of these scenarios where a PR list would be handy:

- See the CI status of a PR even when on a different branch.
- Switch/checkout a branch based on the PR title or PR number. For example if someone says: "I pushed it to [link]" in Slack. Since you can see the PR number in the URL, you can open Atom directly, find the PR and checkout the branch.

And it's meant "As an initial building block toward a pull request review workflow", but I'm not quite sure if I would use the PR list as an entry point when doing a PR review. In most cases, I probably first end up on the PR page on .com and just then decide to review it. The reason is that when getting a notification about a new PR, I either save it to my personal TODO list or just keep the browser tab open. Then whenever I have time, I'll come back to do the review. Other people might have different places where they keep these reminders, like email inbox, project board, but they all have in common that you will click on a link and end up on the PR page on .com. Which brings me to:

It might be worth to make the "I'm looking at a line on .com's diff, I want that same line being open in Atom" as quick and easy as possible? I'll make a separate RFC and we can discuss it there.

ps. I don't mean to say we shouldn't start implementing this RFC, it would be a big improvement over the current version. Just that it would still be a very long way to "I never have to leave Atom" and in the meantime there is a chance to greatly improve the .com <-> Atom workflow.

docs/rfcs/004-issueish-list.md
annthurium: yeah, if we care about Atom's "hackability" providing a generic UI toolkit that package authors could leverage seems like a REALLY high value thing.  (And then we can also bake in accessibility / i18n).  But one thing at a time, since creating a UI toolkit is a large undertaking.

PR #1523, opened 2018-06-12
Pull Request list

smashwilson:
Implement the first part of [RFC-002](https://github.com/atom/github/blob/master/docs/rfcs/002-issueish-list.md): two lists of pull requests in the GitHub tab, one showing the "current" pull request (if any), and the other showing all pull requests on the chosen remote.

##### Before + After

*GitHub tab: before* | *GitHub tab: after*
--- | ---
before | screen shot 2018-06-21 at 1 13 47 pm

*GitHub tab: after, on default ref*

after-default-ref

*GitHub tab: after, on unpublished ref*

after-publish-create

The "current pull request" list displays any pull requests that are associated with the ref that `HEAD` pushes to that are also targeted at the chosen remote.

Special handling also exists for other cases:

* When `HEAD` is a detached ref (like a tag);
* When no repository exists for the push remote of the `HEAD` ref;
* When the GitHub repository has no default ref (before the first push);
* A `HEAD` ref that's not on the default branch, but that hasn't moved from it;
* When a push is in progress;
* When an upstream ref exists, but the local ref has unpushed commits;
* When the upstream ref exists and is up to date, but has no associated pull request.

##### Component hierarchy

* **GithubTabController** pane item methods; fetch async data from repository model; top-level tab actions
  * **RemoteContainer** fetch the appropriate token from the login model; query GraphQL for data about the chosen remote; render login, loading, or error views when appropriate
    * **RemoteController** createPR action method
      * **IssueishSearchesController** manage set of Search models
        * **IssueishSearchContainer** perform GraphQL query for a single search. handle loading and error views.
          * **IssueishListController** translate GraphQL result objects into Issueish models
            * **IssueishListView** render Accordion widget with issueish results
        * **CurrentPullRequestContainer** perform GraphQL query for current PR. handle loading and error views.
          * **IssueishListController** translate GraphQL result objects into Issueish models
            * **IssueishListView** render Accordion widget with issueish results

##### Countdown to :rocket:

- [x] Show the "create a pull request" tile on the "Current pull request" accordion when empty
- [x] Fetch and display the real build status
- [x] Render the real issueish relative age
- [x] Open an IssueishPaneItem on result click
- [x] Render "More..." link when list is truncated
- [x] Rename `RemotePrController`
- [x] Allow IssueishPaneItems to be pending
- [x] Handle QueryRenderer errors within an IssueishListContainer
- [x] Refresh issueish queries
  - [x] On remote git operations
- [x] Investigate backing the "current" PR with an [`associatedPullRequest` from the upstream ref](https://developer.github.com/v4/object/ref/#connections).
- [x] :art: Center the loading spinner in the Github tab again.
- [x] Clean up and consistency
  - [x] Consistently use `tests/factories/pull-request-result` factory
  - [x] Delete now-unused components
  - [x] Remove unused CSS
- [x] Add differentiation icons to each message in CreatePullRequestTile
- [x] Add newlines between sentences in CreatePullRequestTile messages
- [x] Investigate de-emphasizing bolded messages with a different kind of highlighting method that doesn't look as clickable
- [x] Re-word the onDefaultRef message


lib/views/accordion.js
```diff
@@ -0,0 +1,85 @@
...
+    autobind(this, 'toggle');
+
+    this.state = {
+      expanded: true,
+    };
+  }
+
+  render() {
+    return (
+      <details className="github-Accordion" open={this.state.expanded}>
```
simurai: This seems to not work for the first click?

1. Reload Atom -> `open` is present
2. Click on the header -> `open` stays present
3. Click again on the header -> `open` gets removed

open

lib/views/issueish-list-view.js
```diff
@@ -0,0 +1,55 @@
...
+      </Accordion>
+    );
+  }
+
+  renderIssueish(issueish) {
+    return (
+      <Fragment>
+        <img
+          className="github-IssueishList-item--avatar"
+          src={issueish.getAuthorAvatarURL() + "&s=32"}
```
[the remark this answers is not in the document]
smashwilson: Ah right cool. I've added a `size` parameter to the `Issueish.getAuthorAvatarURL()` method and used `URL` manipulation methods to add the query string parameter... just in case.

lib/views/accordion.js
smashwilson: Ahhh good eye. Looks like I was missing a `.preventDefault()` call.

smashwilson:
> Refresh issueish queries periodically

An open question: how should we handle this? We want to be able to see newly created pull requests, and we want to update the build statuses in "real time".

The easy answer is to include an explicit refresh button somewhere. I tend to click those reflexively. Would refreshing on a timer be enough?

Another idea would be to refresh any time the remote is interacted with (by a fetch or a push). I feel like that link would be unclear to users though.

styles/issuesish-list-view.less
```diff
@@ -0,0 +1,65 @@
+@import "variables";
+
+// TODO: Rename
```
smashwilson: 🙈 Oh whoops, did I get the naming scheme wrong... ?

smashwilson:
The pull request publishing tile is in 🏁 It handles all of the edge cases that the PrInfoController did:

screen shot 2018-06-14 at 2 13 05 pm

screen shot 2018-06-14 at 2 12 52 pm

styles/issuesish-list-view.less
simurai: Just a little. :wink: But no worries. It's more like a guide.

Typically, the double dash `--` is used for _variations_ of a component. Like `Button--big`, `Button--small`. They are both buttons, but maybe look a bit different. For `.github-Accordion-list--item` I think I'll change it to `.github-Accordion-listItem` or similar.

simurai:
> An open question: how should we handle this? We want to be able to see newly created pull requests, and we want to update the build statuses in "real time".

Does the GitHub API offer something like a "push notification"? So we don't constantly have to ask if something changed, but GitHub tells us? Like keep some sort of websocket open, like chats do. Or is that overkill? For example, it would be nice after creating a new PR on .com and switching to Atom if the new PR shows up as "Current pull request" and I don't have to manually refresh or wait a few minutes until the timer resets.

> Another idea would be to refresh any time the remote is interacted with (by a fetch or a push). I feel like that link would be unclear to users though.

I think that's a good idea. Because in those situations the chance is much higher that you want to check the PR list. Maybe not as the only method to refresh, but more complementary/in addition. Also it could refresh on focusing the Atom window (with a certain max to limit for those that switch windows very frequently). Like refresh after coming back from lunch or so.

styles/donut-chart.less
```diff
@@ -0,0 +1,13 @@
+@import 'variables';
+
+.donut-ring-pending {
+  stroke: @gh-background-color-yellow;
+}
+
+.donut-ring-succeeded {
```
simurai: Are the donut charts already supposed work? Currently they all have a single `<circle>` with the class `donut-ring-success`. So it wouldn't match ☝️ above class and it's invisible.

Anyways, I [resized](https://github.com/atom/github/pull/1523/files#diff-6fa7b7b04ab5913e8c481ea8db68f092R43) the chart to fit the smaller space.

smashwilson:
> Does the GitHub API offer something like a "push notification"? So we don't constantly have to ask if something changed, but GitHub tells us?

Not now, but eventually. The GraphQL spec has a streaming extension and it's on the integration team's radar, if not their immediate roadmap. I lurk in their Slack channels so hopefully I'll know when it's on the way and can volunteer as an early adopter 😇 

> I think that's a good idea. Because in those situations the chance is much higher that you want to check the PR list. Maybe not as the only method to refresh, but more complementary/in addition. Also it could refresh on focusing the Atom window (with a certain max to limit for those that switch windows very frequently). Like refresh after coming back from lunch or so.

Hmm, okay! I'll give those a shot for this PR and we can see how it feels, maybe.

styles/issuesish-list-view.less
smashwilson: Ohh right. Got it 👍

lib/views/accordion.js
smashwilson: Oh, sure. I do like keeping markup semantic when I get the chance and not just using `<div>` every time. But if it's more convenient we can make the elements whatever's easiest.

simurai:
> Open an IssueishPaneItem on result click

It currently opens a center pane and shows the conversation (PR body, commits, comments). Is that intended?



I thought the plan was to "Open a detail view in a new PaneItem. ...in the same dock as the PR list", the [1.1 option](https://github.com/atom/github/pull/1503#discussion_r192490234).

But having it tried out in practice, I like the current implementation. Going straight to the center and not have and extra dock item floating around. 👍 We only loose the labels that are currently not shown in the center pane, but I think that's fine and we might can add them later again.

Then for the future when we might add PR reviews, we can rethink our design and see how reviews fit in when shown in a center pane.

smashwilson:
Filed atom/atom#17535 based on the work to get the pending item toggling to non-pending properly ✏️

smashwilson:
> It currently opens a center pane and shows the conversation (PR body, commits, comments). Is that intended?

Yah, I tweaked it in the RFC because you'd brought up the earlier point about "scrolling forever" being awkward in a dock item. But now I see that you'd already changed the IssueishPaneItem design to be a dock-item. Oops 😆 

Also, I'm scoping this specific PR to include the work for the pull request list, but I'm not planning to touch the IssueishPaneItem at all here. I'll revisit it to (at least) add the "checkout PR" button somewhere next, most likely, then tackle a larger-scope redesign based on your work there.

simurai:
> I'll revisit it to (at least) add the "checkout PR" button somewhere next, most likely, then tackle a larger-scope redesign based on your work there.

Sounds good. 👍 I'll try to redesign it for "by default in the center", but also "adapting the UI when dragged into a dock".

simurai:
Ok, I made a `sm-pr-pane` branch to play around a bit:

pr-detail

Would still need a lot of restyling, but I think it's a good approach. Open in the center, let people drag into a dock if they want to keep it on the side while coding.

lib/containers/current-pull-request-container.js
```diff
@@ -0,0 +1,167 @@
...
+
+import {autobind} from '../helpers';
+import {RemotePropType, BranchSetPropType, OperationStateObserverPropType} from '../prop-types';
+import IssueishListController, {BareIssueishListController} from '../controllers/issueish-list-controller';
+import CreatePullRequestTile from '../views/create-pull-request-tile';
+import RelayNetworkLayerManager from '../relay-network-layer-manager';
+
+export default class CurrentPullRequestContainer extends React.Component {
+  static propTypes = {
+    repository: PropTypes.shape({
```
annthurium: yay for being specific about prop types!

annthurium:
@smashwilson it would be great if you could add some before/after screenshots to this pull request.

lib/models/issueish.js
```diff
@@ -0,0 +1,65 @@
...
+
+  getGitHubURL() {
+    return this.url.toString();
+  }
+
+  getAuthorLogin() {
+    return this.authorLogin;
+  }
+
+  getAuthorAvatarURL(size = 32) {
```
annthurium: I wonder if we are getting the author avatar URL in a bunch of places in .our codebase.  if so, might not hurt to DRY it up. (Not something that needs to be addressed as part of this PR, just thinking out loud about future cleanup.)

lib/views/accordion.js
```diff
@@ -0,0 +1,104 @@
+import React, {Fragment} from 'react';
+import PropTypes from 'prop-types';
+
+import {autobind} from '../helpers';
+
+export default class Accordion extends React.Component {
+  static propTypes = {
+    leftTitle: PropTypes.string.isRequired,
+    rightTitle: PropTypes.string,
+    results: PropTypes.arrayOf(PropTypes.any).isRequired,
```
annthurium: do we not know the shape of the results?

smashwilson: Accordion is meant to be a reusable widget; in general, `results` could be anything that the `children` render-prop can interpret. Of course, in practice, these are always `Issueish` instances, because the only time render an Accordion is when we have a list of those. It felt wrong to constrain it here, though.

annthurium: 👍

smashwilson:
Okay, I'll merge when I have ✅ reviews from @annthurium, @sguthals, and @simurai.

test/containers/current-pull-request-container.test.js
```diff
@@ -0,0 +1,207 @@
...
+    resolve();
+    await promise;
+
+    assert.isFalse(wrapper.update().find('BareIssueishListController').prop('isLoading'));
+  });
+
+  it('passes an empty result list and an error prop to the controller when errored', async function() {
+    const {reject, promise} = useEmptyResult();
+
+    const e = new Error('oh no');
```
annthurium: image

test/controllers/issueish-list-controller.test.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import {shallow} from 'enzyme';
+
+import {createPullRequestResult} from '../fixtures/factories/pull-request-result';
+import Issueish from '../../lib/models/issueish';
+import {BareIssueishListController} from '../../lib/controllers/issueish-list-controller';
+
+describe('IssueishListController', function() {
```
annthurium: This component has an error state too, yes?  Would that be worth testing?

test/models/search.test.js
```diff
@@ -0,0 +1,34 @@
+import Remote, {nullRemote} from '../../lib/models/remote';
+
+import Search from '../../lib/models/search';
+
+describe('Search', function() {
+  const origin = new Remote('origin', 'git@github.com:atom/github.git');
+
+  it('generates a dotcom URL', function() {
```
annthurium: would it be good to test the error thrown for non-dotcom repos here also?

test/views/issueish-list-view.test.js
```diff
@@ -0,0 +1,145 @@
...
+      const wrapper = mount(buildApp({isLoading: false, total: 1, issueishes: [allGreen]}));
+      assert.isTrue(wrapper.find('span.github-IssueishList-item--status').hasClass('icon-check'));
+    });
+
+    it('renders an x if all status checks have failed', function() {
+      const wrapper = mount(buildApp({isLoading: false, total: 1, issueishes: [allRed]}));
+      assert.isTrue(wrapper.find('span.github-IssueishList-item--status').hasClass('icon-x'));
+    });
+
+    it('renders a donut chart if status checks are mixed', function() {
```
annthurium: image

Please let me know if you get tired/annoyed of the bitmoji thing, I can stop.

simurai:
> it would be great if you could add some before/after screenshots to this pull request.

👍 I was wondering why the status icons aren't centered in @smashwilson's screenshots... mystery solved:

icons

lib/views/accordion.js
smashwilson: ... so this is still true here, but I see I got lazy in the PropTypes for `IssueishListView`, where we _do_ know that they're an array of `Issueishes`. I just pushed a commit that tightens up that constraint there. 😅

test/controllers/issueish-list-controller.test.js
smashwilson: Done 👍

test/views/issueish-list-view.test.js
smashwilson: > Please let me know if you get tired/annoyed of the bitmoji thing, I can stop.

I'll be sure to let you know when it stops making me laugh every time 😆

sguthals:
👋 
Thank you @smashwilson and @simurai for this awesome progress on the pr review experience in Atom 🎉 

A couple things that I think we could improve before 🚢 
## Calling attention to a UI change
When you are on master, `Current pull request` has the following messaging:
screen shot 2018-06-21 at 3 00 58 pm
And when you create a new branch or switch branches that are in sync with master you see the following:
screen shot 2018-06-21 at 3 00 50 pm
These are extremely similar in look, and nothing happens to call attention to the fact that they have changed. Could we brainstorm a solution to drawing attention to this updated information for the user so that they know what to do next?

## (potentially)Fixing messaging
The messaging in the first image above is:
` You are currently on your repository's default branch. Create a new branch to share your work with a pull request`
However, I see three possible scenarios for a user when they are leaving the master branch:
1. They are going to a new branch and will potentially open a pr for that branch (what is currently in the messaging)
2. They are going to an existing branch that is not a part of a pr
3. They are going to an existing branch that is a part of a pr
I'm trying to figure out where we want to go eventually with our messaging here. Eventually we want things like what we have in the [rfc](https://github.com/atom/github/blob/master/docs/rfcs/002-issueish-list.md):
> "Checkout" button to fetch (if necessary) and check out the pull request. Only enabled if the current pull request is not the current one.
It would be great to get some mock ups where this entire feature would start. As in, what could we imagine is the entry point to the entire pull request experience? 
I would like to continue this conversation in a different Issue, for the purposes of this pr, considering we don't have checkout yet, should we add something to the effect of ` You are currently on your repository's default branch. Checkout or create a new branch to share your work with a pull request`?

## Confirmation of scope of this pr
Finally, just to confirm, this pr scope does not include the following from the rfc yet:
- [ ] Branches -> master < aw/rfc-pr-list
- [ ] "Checkout" button to fetch (if necessary) and check out the pull request. Only enabled if the current pull request is not the current one.
- [ ] Commits with count, links to .com (for now), optional with avatars
- [ ] Checks with count, links to .com (for now)
- [ ] CI status, each item links to the detail page
- [ ] Files changed with count, links to .com (for now), optional with "+-" bar
- [ ] Mergability status -> Able to merge, links to the Merging controls at the bottom
- [ ] "Merge PR" to merge the pull request on GitHub if it is open.
- [ ] "Close" to close the pull request, unmerged, if it is open.
- [ ] "Re-open PR" to re-open a pull request if it is closed.
- [ ] Conversation with comment count, opens the current PR timeline in a center pane.
- [ ] Reaction emoji and counts.
- [ ] Description (PR body) as rendered markdown.

That is totally fine, I just wanted to make sure I wasn't missing that :D

smashwilson:
> These are extremely similar in look, and nothing happens to call attention to the fact that they have changed

👍 Yeah, good point. Maybe we could differentiate them with some iconography on the left... ? For reference, you can see all of the special cases for that tile here:

https://github.com/atom/github/blob/a0c17a759245f6bdc325241cae3b51d38df6accb/lib/views/create-pull-request-tile.js#L24-L102

By the way: something that I really wanted to do was, rather than just bolding some text, offer a control that lets you _do_ the thing that we're prompting for. The problem is that we can't currently enter the branch creation flow because it's done with a popup. I'm hoping that we can revisit that after #1370 is accepted.

> (potentially) Fixing messaging

For a bit of background, I took the messaging verbatim from the old GitHub tab code, which I'd introduced kind of ad-hoc in #1376. Happy to workshop better text with anyone who has ideas for improvements :smile: We do have less space in the tile version, so it could likely stand to be made more terse, if nothing else.

> However, I see three possible scenarios for a user when they are leaving the master branch:

🤔 True, true. Is there some simple way we can say "you have to move off of master for this to be useful"... ? The compare view from master to master would always be empty, which is why we have that particular scenario special-cased to begin with. I'm not sure of an alternative that covers all of the possible next actions without being really wordy.

> I'm trying to figure out where we want to go eventually with our messaging here. Eventually we want things like what we have in the rfc:
>
> > "Checkout" button to fetch (if necessary) and check out the pull request. Only enabled if the current pull request is not the current one.

That's destined for the issueish pane item at the moment, though. The "current pull request" is populated based on `HEAD` so checkout would never be a viable action, right?

> Finally, just to confirm, this pr scope does not include the following from the rfc yet:

👍 Yup yup. This is intended to include everything from the [Accordion lists](https://github.com/atom/github/blob/master/docs/rfcs/002-issueish-list.md#accordion-lists) and [New PR](https://github.com/atom/github/blob/master/docs/rfcs/002-issueish-list.md#new-pr) sections of the RFC. "Issueish Pane Item" is still yet to be done. The pane item I'm opening on click here is entirely pre-existing code.

sguthals:
> Maybe we could differentiate them with some iconography on the left... ?

Yeah! That's a great start! @simurai what do you think?

> By the way: something that I really wanted to do was, rather than just bolding some text, offer a control that lets you do the thing that we're prompting for. The problem is that we can't currently enter the branch creation flow because it's done with a popup. I'm hoping that we can revisit that after #1370 is accepted.

✨ ✨ ✨ 

> I'm not sure of an alternative that covers all of the possible next actions without being really wordy.

Let's brainstorm just a bit but not let it totally block us today

daviwil:
This UI looks so friggin' good ✨

simurai:
The [messages](https://github.com/atom/github/blob/a0c17a759245f6bdc325241cae3b51d38df6accb/lib/views/create-pull-request-tile.js#L24-L102) are now split into two parts:

1. **Why** there is no PR.
2. Possible **next step**.

The "next step" has an icon as a hint.

image

## All messages

Screenshot | if
--- | ---
isOnDefaultRef | `isOnDefaultRef`
isSameAsDefaultRef | `isSameAsDefaultRef`
hasNoDefaultRef | `hasNoDefaultRef`
isDetachedHead | `isDetachedHead`
isRepositoryNotFound | `isRepositoryNotFound`
differentRemote | `differentRemote`
default | `default`

## Alternatives

I tried _italic_ but it doesn't stick out as much:

image

Since both parts have `bold` words, it might look less like they are links to click?

/cc @sguthals @annthurium @smashwilson

smashwilson:
pshooooo

Issue #812, opened 2017-05-16
Navigation between merge conflicts


smashwilson:
From smashwilson/merge-conflicts#267:

> Quite possible I missed something visually while using this tool. If that is the case, just let me know.
>
> I had two conflicts in the same file. I clicked on the "use ours" button.
> The update to the file was made in atom's buffer.
> I was completely clueless how to move to the next conflict quickly. I did not see a button the footer UI and scrolling wasn't really an option as there were a few thousand lines between the different conflicts.
> I ended up needing google to get me to atom.io/packages/merge-conflicts so I could figure out the key combo.
>
> Felt like there should have been a UI button on the footer interface.
>
> merge-conflicts v1.4.4.

It'd be cool to have a (discoverable) way to navigate among conflicts within a file or across files.

smashwilson:
A related suggestion from @Haacked that's a good idea:

> When you click on a conflicted file, scroll the first conflict into the view.

ungb:
Also on this, I noticed that when clicking on the file, it doesn't take you to the merge conflict for the first one.  See gif:



![merge](https://cloud.githubusercontent.com/assets/5624870/26425959/957cbf38-408b-11e7-9110-9abf8ff5ecd4.gif)

nadim:
Is it possible for an official response on this?

BinaryMuse:
@nadim Well @smashwilson works for GitHub on Atom, so since he opened the issue I think it's fairly official already. :) This is something we'll definitely fix when we can.

vol24pl:
Well, for me "prev/next conflict" might be the most important feature in the conflicts handling.

https://github.atom.io/videos/github-feature-resolve-conflict.mp4

By looking at this movie from official website I thought that clicking on conflict takes you to the first conflict in the file (not the beginning of the file). Conveniently in the movie beggining of the file was also in the visual center of the conflicted lines :)

kanzelm3:
@heisian that's a simple workaround, but from a UX standpoint, I think this would still be a great feature to add to the native conflict resolution tool in Atom. The merge conflict resolution tool on github.com has this functionality and it allows you to jump through conflicts and resolve them without having to remember what characters to search for. It's not super critical, but it's a nice to have!

Rihoj:
@smashwilson I see that the Feature Sprint v0.20.0 was closed out, but this was not. Was this never continued?

daninfuchs:
[...]

PR #1525, opened 2018-06-13
RFC for Open in Atom

simurai:
A quick way to switch from github.com to Atom.

open-in-atom

### DIFF
--- a/docs/rfcs/003-open-in-atom.md
+++ b/docs/rfcs/003-open-in-atom.md
@@ -1,0 +1,137 @@
+# Open in Atom
+
+## Status
+
+Proposed
+
+## Summary
+
+A quick way to switch from github.com to Atom.
+
+## Motivation
+
+There are many ways how a user can discover new pull requests. There are also many ways where a user might save a pull request for later. But something that they all have in common: A URL.
+
+![image](https://user-images.githubusercontent.com/378023/41331997-d8782114-6f16-11e8-9709-9299b250292e.png)
+
+At the time you decide to start working on something, chances are high you are on github.com or can click a link to github.com. Therefore let's make it easy switch from github.com and continue in Atom.
+
+### User Story:
+
+In the morning Jody arrives at the office and checks their email. There are lots of notifications from GitHub. Jody briefly scans through them and stars the ones that look interesting or important. Then Jody keeps working on a new feature they started yesterday. After lunch, Jody decides to go back to their starred emails. A notification that got their interest is about a review request from a co-worker. Jody is eager to take a look and clicks on the link in the email. The browser opens and the PR page is shown. Jody reads through the conversation to understand what the PR is about, then switches to the "Files changed" to look at the code changes. Jody is thinking: "Hmm.. I wonder how it would feel to change the order?".
+
+This is a moment where an "Open in Atom" feature would come in handy. Effortlessly switch from looking at a line on github.com to have the same line open in Atom, ready to be edited and tested locally.
+
+Instead, Jody does the following:
+
+1. Scrolls to the top of the page
+2. Clicks on "Conversation"
+3. Scrolls to the bottom of the page
+4. Clicks on "command line instructions"
+5. Copies the `git checkout` part
+6. Opens Terminal
+7. CDs to the repo's directory
+8. Pastes the `git checkout` command
+9. Types `atom .`
+10. While Atom is opening, goes back to the diff to copy the line.
+11. In Atom, does a "Find in project" with the copied line
+12. Clicks on the found result
+13. Finally is ready to edit the line
+
+How can we improve this "start on github.com, continue in Atom" workflow? How can the dozen of steps in the above example be reduced to just 1-2 clicks?
+
+## Explanation
+
+In various places on github.com a user has the option to "Open in Atom". It's the same idea as the "Open in Desktop" that already exists.
+
+> Note: Out of simplicity, GitHub Desktop or other editors are not shown in the mockups. If multiple options are available, a dropdown could be used to make a choice.
+
+![image](https://user-images.githubusercontent.com/378023/41336470-4fe89d04-6f27-11e8-9d0d-9f5b4682db9b.png)
+
+Once a user authenticates the GitHub package in Atom with their GitHub login, an "Open in Atom" option will be available on github.com in the following places:
+
+### Repo
+
+1. Clones the repo to Atom's "Project Home" (config) directory
+2. Opens a new Atom window with the repo as its project
+
+![image](https://user-images.githubusercontent.com/378023/41266162-c59358b6-6e30-11e8-9cf9-3c420fb1c2db.png)
+
+### Pull Request
+
+1. Clones the repo (if needed)
+2. Checks out the branch of the PR
+3. Opens a new Atom window
+    - Or focuses the window if the repo already exists as an opened project
+4. Opens the GitHub panel showing the PR details
+
+![image](https://user-images.githubusercontent.com/378023/41266455-e0abfd96-6e31-11e8-9f0e-124a3adc4058.png)
+
+
+### Merge conflict
+
+1. Clones the repo (if needed)
+2. Checks out the branch
+3. Opens a new Atom window
+    - Or focuses the window if the repo already exists as an opened project
+4. Opens the Git panel (showing all conflicting files)
+5. Opens the (first) conflicting file
+6. Scrolls to the (first) conflict
+
+![image](https://user-images.githubusercontent.com/378023/41271078-6c8e6a6c-6e49-11e8-96ee-5b0b5e55ef3d.png)
+
+### File
+
+1. Clones the repo (if needed)
+2. Checks out the branch
+3. Opens a new Atom window
+    - Or focuses the window if the repo already exists as an opened project
+4. Opens the file
+
+![image](https://user-images.githubusercontent.com/378023/41266930-30a24092-6e34-11e8-9c6e-36af97bc2bb3.png)
+
+### Line
+
+1. Clones the repo (if needed)
+2. Checks out the branch
+3. Opens a new Atom window
+    - Or focuses the window if the repo already exists as an opened project
+4. Opens the file
+5. Scrolls to the line
+6. Shows a blinking cursor at the beginning of the line 
+
+![open-in-atom](https://user-images.githubusercontent.com/378023/41338002-ccfe8944-6f2b-11e8-95d7-7e823e6d8643.png)
+
+
+## Drawbacks
+
+None that come to mind.
+
+## Rationale and alternatives
+
+By reducing the friction for moving from github.com to Atom, we can improve the following:
+
+- Higher quality of code reviews.
+    - Sometimes a reviewer might have the urge to suggest something, but they are not sure without testing it first. But since testing locally takes so many steps, they might just leave a comment like "Looks good to me" and move on. 👉 The easier it is to test changes, the better the reviews will become.
+- More contributions.
+    - If a reviewer tests something locally and makes edits, they might commit the changes and push it back to the PR. 👉 Reviewers that can easily edit and test a PR more likely become collaborators.
+- Better integration with other applications.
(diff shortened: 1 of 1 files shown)


docs/rfcs/003-open-in-atom.md
```diff
@@ -0,0 +1,137 @@
...
+> Note: Out of simplicity, GitHub Desktop or other editors are not shown in the mockups. If multiple options are available, a dropdown could be used to make a choice.
+
+![image](https://user-images.githubusercontent.com/378023/41336470-4fe89d04-6f27-11e8-9d0d-9f5b4682db9b.png)
+
+Once a user authenticates the GitHub package in Atom with their GitHub login, an "Open in Atom" option will be available on github.com in the following places:
+
+### Repo
+
+1. Clones the repo to Atom's "Project Home" (config) directory
+2. Opens a new Atom window with the repo as its project
```
smashwilson: Would we want to offer a config option for new window / add to most recent window?

Personally, I use `atom -a` pretty heavily from the command-line, and organize my project roots into a single Atom window per monitor. I'm sure there's some variance in usage here though.

docs/rfcs/003-open-in-atom.md
```diff
@@ -0,0 +1,137 @@
...
+
+![image](https://user-images.githubusercontent.com/378023/41266162-c59358b6-6e30-11e8-9cf9-3c420fb1c2db.png)
+
+### Pull Request
+
+1. Clones the repo (if needed)
+2. Checks out the branch of the PR
```
smashwilson: Note that this'll take a special refspec to fetch PR refs (in the common case of checking out a PR from a fork). We'll already need to cross that bridge to implement "checkout PR" anyway, though.

docs/rfcs/003-open-in-atom.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Merge conflict
+
+1. Clones the repo (if needed)
+2. Checks out the branch
+3. Opens a new Atom window
+    - Or focuses the window if the repo already exists as an opened project
+4. Opens the Git panel (showing all conflicting files)
+5. Opens the (first) conflicting file
+6. Scrolls to the (first) conflict
```
smashwilson: Related: we should also be doing this when you open a file from the Git tab, but we aren't yet; see #812.

docs/rfcs/003-open-in-atom.md
```diff
@@ -0,0 +1,137 @@
...
+
+![image](https://user-images.githubusercontent.com/378023/41271078-6c8e6a6c-6e49-11e8-96ee-5b0b5e55ef3d.png)
+
+### File
+
+1. Clones the repo (if needed)
+2. Checks out the branch
```
smashwilson: After each of these "checks out the branch" steps, we'll also need to do a fetch if the commit can't be found locally.

Also, we'll need to make sure we choose the branch with a fetch upstream remote and ref matching the one on .com - they don't _have_ to be named the same thing. There are some trailing edge cases to consider like "you already have multiple branches in your local repo that all fetch from the same place".

docs/rfcs/003-open-in-atom.md
```diff
@@ -0,0 +1,137 @@
...
+3. Opens a new Atom window
+    - Or focuses the window if the repo already exists as an opened project
+4. Opens the file
+
+![image](https://user-images.githubusercontent.com/378023/41266930-30a24092-6e34-11e8-9c6e-36af97bc2bb3.png)
+
+### Line
+
+1. Clones the repo (if needed)
```
smashwilson: Similarly, for the "clone the repo (if needed)" steps - do you have any insight into how we can determine if the repo already exists or not? I'd rather not be too prescriptive about the way users organize their filesystems if we can avoid it. Maybe we can store a mapping of repository URL or nwo :point_right: project path in a file under `ATOM_HOME` and use that to find them?

docs/rfcs/003-open-in-atom.md
```diff
@@ -0,0 +1,137 @@
...
+## Rationale and alternatives
+
+By reducing the friction for moving from github.com to Atom, we can improve the following:
+
+- Higher quality of code reviews.
+    - Sometimes a reviewer might have the urge to suggest something, but they are not sure without testing it first. But since testing locally takes so many steps, they might just leave a comment like "Looks good to me" and move on. 👉 The easier it is to test changes, the better the reviews will become.
+- More contributions.
+    - If a reviewer tests something locally and makes edits, they might commit the changes and push it back to the PR. 👉 Reviewers that can easily edit and test a PR more likely become collaborators.
+- Better integration with other applications.
+    - Be able to open files, lines, merge conflicts etc. in Atom isn't just limited to github.com, it's also possible for other applications. For example if there is a merge conflict in GitHub Desktop, instead of manually switching and trying to find the conflict, GitHub Desktop could send you to Atom in a more seamless way.
```
smashwilson: 💯

docs/rfcs/003-open-in-atom.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Alternatives
+
+An alternative is to let users discover PRs/issues or get notified about changes **within** Atom. This would certainly be nice to have, but it will be a very long road until a "never have to leave Atom" is reached. Especially for our tiny team. We also don't have to pick one or the other and both approaches can complement each other.
+
+An alternative to have an Atom integration directly build into github.com would be to create a **browser extension**. A benefit is that it can be extended to other web sites/apps. For example an "Open in Atom" link inside Gmail for GitHub notifications. Downside: Users have to first install the extension.
```
smashwilson: Yah, I much prefer this to writing browser extensions, I think - much more discoverable.

docs/rfcs/003-open-in-atom.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Unresolved questions
+
+- [ ] Is github.com interested in adding an "Open in Atom" option? And possibly other editors? There is already a [Clone in Xcode](https://blog.github.com/2017-06-05-clone-in-xcode/) option.
+- [ ] Are there any security risks that need to be investigated beforehand.
+- [ ] What could the API (URI) look like? E.g. `atom://github/open?pull=1234`
```
smashwilson: Hmm, let's see. How about:

```
atom://github/open/smashwilson%2Fremote-repo?pull=1234&line=456
```

And from an Enterprise instance, we'd encode the full remote URL:

```
atom://github/open/https%3A%2F%2Fgithub.enterprise.com%2Fsmashwilson%2Fremote-repo?ref=branchname&sha=abc123&file=some%2Fpath%2Fsource.js&line=456
```

Hard to say until we start implementing it though, I'm sure there's more info I won't realize we need until we're neck-deep in code 😆

docs/rfcs/003-open-in-atom.md
```diff
@@ -0,0 +1,137 @@
...
+12. Clicks on the found result
+13. Finally is ready to edit the line
+
+How can we improve this "start on github.com, continue in Atom" workflow? How can the dozen of steps in the above example be reduced to just 1-2 clicks?
+
+## Explanation
+
+In various places on github.com a user has the option to "Open in Atom". It's the same idea as the "Open in Desktop" that already exists.
+
+> Note: Out of simplicity, GitHub Desktop or other editors are not shown in the mockups. If multiple options are available, a dropdown could be used to make a choice.
```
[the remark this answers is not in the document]
simurai: Based on https://github.com/github/atom-log/issues/425#issuecomment-261125392, we should be able to add more clients. But that might only be for cloning a repo and not for all the other places.

docs/rfcs/003-open-in-atom.md
```diff
@@ -0,0 +1,137 @@
...
+
+![image](https://user-images.githubusercontent.com/378023/41336470-4fe89d04-6f27-11e8-9d0d-9f5b4682db9b.png)
+
+Once a user authenticates the GitHub package in Atom with their GitHub login, an "Open in Atom" option will be available on github.com in the following places:
```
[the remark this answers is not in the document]
simurai: Heh.. good question. 🤔

docs/rfcs/003-open-in-atom.md
simurai: Would be nice. I think if you clone in Atom, it adds a new project to the current window.

Personally, I use the same Atom window, but try to switch between projects with packages like https://atom.io/packages/project-switcher2. So it replaces the current project in the current window. 😄

docs/rfcs/003-open-in-atom.md
simurai: > Maybe we can store a mapping of repository URL or nwo 👉 project path in a file under `ATOM_HOME` and use that to find them?

Is `ATOM_HOME` the "Project Home" config?

image

> I'd rather not be too prescriptive about the way users organize their filesystems

Yeah, it would become pretty annoying if you would have to tell Atom where to find the repo each time. I guess Desktop gets around that by having an index of all projects/repos so it knows their local path.

smashwilson: > Is `ATOM_HOME` the "Project Home" config?

Nah, it's `~/.atom` by default - the location where Atom looks for its config, user init scripts, package installs, things like that. We already have a `github.cson` file there that we could use for this.

smashwilson: > I guess Desktop gets around that by having an index of all projects/repos so it knows their local path.

🤔 I wonder if it's stored somewhere we could read and write?

simurai: > I wonder if it's stored somewhere we could read and write?

Probably not from another application?

Ohh.. now I just remembered.. Atom has this "Reopen Project":

image

Maybe we could check that list? Could be like:

1. Check "Reopen Project" list if the repo exists.
2. Check "Project Home" directories if the repo exists. Maybe even nested?
3. If the repo is not found, show a dialog where you can choose the local path.
    - But you can also choose an existing directory that contains the repo. Atom then _should_ add that directory to the "Reopen Project" list so it will find it automatically the next time.

Not sure where "Reopen Project" gets stored, but might be neat to have it as `.atom/projects.cson` so that you can make a backup or also edit it by hand. I think https://atom.io/packages/project-manager does something like that.

Scope creep! :stuck_out_tongue_closed_eyes:

docs/rfcs/003-open-in-atom.md
simurai: > we'd encode the full remote URL:

Or in all cases? Would this be possible: Keep it exactly the same as .com URLs, except replace `https://github.com` with 👉 `atom://github/open`

- .com URL: `https://github.com/atom/github/pull/1525/files#diff-7e96b28649369913c5e87f137813ade5R131`
- Atom: `atom://github/open/atom/github/pull/1525/files#diff-7e96b28649369913c5e87f137813ade5R131`

This would make it easier to turn a .com URL into and Atom URI. By hand or by other apps, browser extensions etc.

Also, as a first step, Atom could offer an "Open from GitHub URL" like in [Visual Studio](https://github.com/github/VisualStudio/pull/1763). Then later consider adding "Open in Atom" buttons to .com for an even better UX.

docs/rfcs/003-open-in-atom.md
jcansdale: This functionality could be easily implemented as a bookmarklet.

image

For example the following pseudo-link would send any URL to the `git-client` protocol (which is used by Visual Studio):
```
javascript:window.location="git-client://clone?repo="+document.URL
```

This might be useful for prototyping or for users who want this functionality close at hand. Just a crazy idea. 😉

docs/rfcs/003-open-in-atom.md
jcansdale: Or even simpler:
```
atom://github?open=https://github.com/atom/github/pull/1525/files#diff-7e96b28649369913c5e87f137813ade5R13
```
You'd need to escape the `#`, but apart from that I think it would work.

simurai:
Just FYI: I gave this another try:

1. Added the following bookmarklet `javascript:window.location="atom://open-from-github?url="+document.URL`
2. Installed this Atom package https://github.com/simurai/open-from-github

Then it lets me switch between projects.. super fast! 🐎 

open-in-atom

But I think that's as far as my skills (and time) let me go.. :sweat_smile: Potentially clone the repo, check out the right branch, open the file, move the cursor to the right line.. would still need to be added. Plus probably tons of edge cases.

PR #1532, opened 2018-06-15
Render markdown flavored emojis in recent commits view.

annthurium:
Before:
screen shot 2018-06-15 at 3 27 12 pm

After:
screen shot 2018-06-15 at 3 26 38 pm

### Approach:
I used the package [node-emoji](https://github.com/omnidan/node-emoji) because it was updated relatively recently, and small (27.9 kb, or 12.4 kb zipped according to bundephobia).

###Test Plan:
- [x] manually test that commit titles and text contains emoji
- [x] manually test that commits that contain colons but no emoji display normally
- [x] add unit tests for commit titles and commit text that contain emoji

🍕 🦄 ❤️

### DIFF
--- a/lib/views/recent-commits-view.js
+++ b/lib/views/recent-commits-view.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import moment from 'moment';
 import cx from 'classnames';
-import { emojify } from 'node-emoji';
+import {emojify} from 'node-emoji';
 
 import Timeago from './timeago';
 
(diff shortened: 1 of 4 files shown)


daviwil:
Nice!  I know that the Atom Release Publisher service uses the [emojis](https://www.npmjs.com/package/emojis) library to render emojis in commit messages, but I don't know if the results you'd get would be better or worse.

jasonrudolph:
🙌 :shipit: 🎉

Issue #1530, opened 2018-06-14
Push/pull only works when pressing the text itself, not the status bar tile


50Wliu:
### Prerequisites

* Have you tried to reproduce the problem in [Safe Mode][safe-mode]?
* Have you followed all applicable steps in the [debugging guide][debugging-guide]?
* Have you checked the FAQs on the [message board][message-board] for common solutions?
* Have you checked that your issue [isn't already filed][already-filed]?

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

### Description

The push/pull action only goes through when you are clicking on the actual text, not the tile itself, despite the entire tile appearing to be clickable.

![github-push-pull-button-area](https://user-images.githubusercontent.com/2766036/41429143-326671b2-6fda-11e8-955a-4f8e10a816b9.gif)

### Steps to Reproduce

1. Attempt to click the push/pull tile, but *not* the text itself
2. See the tile react to being clicked, but nothing actually happens
3. Try again by clicking on the text area
4. Notice the push/pull action actually occurs

**Expected behavior:**

Both 2) and 4) perform the push/pull action

**Actual behavior:**

Only clicking on the text itself triggers the action

**Reproduces how often:**

100%

### Versions

You can get this information from copy and pasting the output of `atom --version` and `apm --version` from the command line. Also, please include the OS and what version of the OS you're running.

```
Atom    : 1.29.0-dev-0422e0d61
Electron: 2.0.1
Chrome  : 61.0.3163.100
Node    : 8.9.3
```
Windows 10 x64.

annthurium:
Thanks @50Wliu for reporting!  We should definitely fix this.  I can add this to my list of short term priorities.

annthurium:
hmm, actually this issue belongs in https://github.com/atom/status-bar.  Lemmesee if I can find a way to move it.

annthurium:
I don't think you can actually move issues...recreated the issue over there.

https://github.com/atom/status-bar/issues/224

Issue #1531, opened 2018-06-14
Push/pull tooltip shows up much quicker than other status bar tooltips


50Wliu:
### Prerequisites

* Have you tried to reproduce the problem in [Safe Mode][safe-mode]?
* Have you followed all applicable steps in the [debugging guide][debugging-guide]?
* Have you checked the FAQs on the [message board][message-board] for common solutions?
* Have you checked that your issue [isn't already filed][already-filed]?

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

### Description

The tooltip for the push/pull status bar tile shows up much quicker (almost instantly) than any other tile in the status bar.

![github-push-pull-button-tooltip](https://user-images.githubusercontent.com/2766036/41429407-09e7a5f2-6fdb-11e8-8a8f-7bad42da2731.gif)

### Steps to Reproduce

1. Hover over the push/pull status bar tile

**Expected behavior:** Tooltip to appear after a delay

**Actual behavior:** Tooltip appears almost instantly

**Reproduces how often:** 100%

### Versions

```
Atom    : 1.29.0-dev-0422e0d61
Electron: 2.0.1
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

annthurium:
thanks for reporting! moved to `status-bar` repo: https://github.com/atom/status-bar/issues/225

annthurium:
@50Wliu ahh, my mistake, it's not!  Will reopen and close that other issue.  Sorry about that.

annthurium:
addressed in https://github.com/atom/github/pull/1534

PR #1534, opened 2018-06-18
Make entire component for push/pull tile in status bar clickable.

annthurium:
fixes https://github.com/atom/github/issues/1530 and https://github.com/atom/github/issues/1531

I don't know how to unit test this.  If we had Selenium tests, or something that actually fired up the UI, that might be possible. 

Test plan:
- test clicking on the non-text area of the status bar in dev mode
- test clicking on neighboring tiles to make sure they fire the expected actions.

### DIFF
--- a/lib/views/push-pull-view.js
+++ b/lib/views/push-pull-view.js
@@ -195,18 +195,14 @@ export default class PushPullView extends React.Component {
       tileState = tileStates.detached;
     }
 
-    const Tag = tileState.onClick ? 'a' : 'span';
-
     return (
       <div
+        onClick={tileState.onClick}
         ref={this.refTileNode.setter}
         className={cx('github-PushPull', 'inline-block', {'github-branch-detached': isDetached})}>
         {tileState && (
           <Fragment>
-            <Tag
-              onClick={tileState.onClick}
-              className="push-pull-target"
-              key="push-pull-target">
+            <span>
               {tileState.secondaryText && (
                 <span className="secondary">
                   <span className={getIconClass(tileState.secondaryIcon)} />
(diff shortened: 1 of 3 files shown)


styles/push-pull-view.less
```diff
@@ -2,6 +2,11 @@
 
 // Used in the status-bar
 
+// click target should take up the width of the container
+.push-pull-target {
```
simurai: This works for the height, but for the One UI themes, the `width` doesn't go all the way to the edge because the parent [github-PushPull](https://github.com/atom/github/blob/4519a1217313591aae9b672ee3451bebe6c7ed1a/lib/views/push-pull-view.js#L201-L203) element has some padding:

image

Hmm.. would it be possible to move the click target to that parent `github-PushPull` element? And remove the `<a class="push-pull-target">`?

If not or if there is a lot of refactoring needed, we could also do this:

1. Merge this PR as is. Fixes height.
2. Add some "negative margin hack" to the One UI themes.

Like:

```less
.push-pull-target {
  display: block;
  margin: 0 -0.75em;
  padding: 0 0.75em;
}
```

annthurium: thanks for the feedback! I can take a stab at removing the `pushPull` element.

lib/views/push-pull-view.js
```diff
@@ -215,14 +211,14 @@ export default class PushPullView extends React.Component {
...
-            </Tag>
+            </span>
             <Tooltip
               key="tooltip"
               manager={this.props.tooltipManager}
               target={this.refTileNode}
               title={`<div style="text-align: left; line-height: 1.2em;">${tileState.tooltip}</div>`}
-              showDelay={200}
-              hideDelay={100}
+              showDelay={atom.tooltips.hoverDefaults.delay.show}
```
annthurium: it makes me nervous using dot notation to access deeply nested object properties, without existence checks.  However I think this is an atom api that's used in a lot of places, so maybe it's safe to assume this is ok.

annthurium: right right, I meant api as in interface/contract, that `atom.tooltips.whatever` would continue to be available.  Not api as in REST api. 😄

PR #1547, opened 2018-06-22
Move git and github tab items to a PaneItem

### LINKED ISSUE #1543: Eliminate DockItem
The `DockItem` component does nothing that `PaneItem` can't do; an Atom dock item is just a pane item that implements a method called `getDefaultLocation()`. Let's refactor to only use `PaneItem`, then get rid of `DockItem`.

smashwilson:
Let's eliminate some dead code and be a bit more consistent.

Fixes #1543.


smashwilson:
Phew. Did not expect this one to take so long. In any case, it's working properly now. As a side bonus we have an integration test harness, too.

Issue #1538, opened 2018-06-19
Undoing a commit with a co-author throws an exception that abruptly closes the git tab.


annthurium:
### Description

Undoing a commit with a co-author is busted.

### Steps to Reproduce

1. Open Atom and open the Git tab.
2. Make some changes.
3. Stage the changes.
4. Commit the changes (with a co author)
5. Click "undo" on the most recent commit.

**Expected behavior:**

The changes in the last commit would be staged, the last commit would disappear, and the co-author would still be selected.

**Actual behavior:**

The Git tab abruptly disappears, with the following stack trace:
```
TypeError: selected.matches is not a function
    at authors.filter.selectedAuthors.find.selected (/Users/annthurium/github/atom/node_modules/github/lib/views/commit-view.js:487:94)
    at Array.find (native)
    at authors.filter (/Users/annthurium/github/atom/node_modules/github/lib/views/commit-view.js:487:68)
    at Array.filter (<anonymous>)
    at matchAuthors (/Users/annthurium/github/atom/node_modules/github/lib/views/commit-view.js:486:36)
    at Select.filterOptions (/Users/annthurium/github/atom/node_modules/react-select/lib/Select.js:1112:12)
    at Select.render (/Users/annthurium/github/atom/node_modules/react-select/lib/Select.js:1262:46)
    at Gg (/Users/annthurium/github/atom/node_modules/react-dom/cjs/react-dom.production.min.js:166:226)
    at Kg (/Users/annthurium/github/atom/node_modules/react-dom/cjs/react-dom.production.min.js:179:75)
    at th (/Users/annthurium/github/atom/node_modules/react-dom/cjs/react-dom.production.min.js:201:22)
    at uh (/Users/annthurium/github/atom/node_modules/react-dom/cjs/react-dom.production.min.js:202:234)
    at Rh (/Users/annthurium/github/atom/node_modules/react-dom/cjs/react-dom.production.min.js:211:393)
    at Uh (/Users/annthurium/github/atom/node_modules/react-dom/cjs/react-dom.production.min.js:210:278)
    at Sh (/Users/annthurium/github/atom/node_modules/react-dom/cjs/react-dom.production.min.js:210:113)
    at wh (/Users/annthurium/github/atom/node_modules/react-dom/cjs/react-dom.production.min.js:208:281)
    at kg (/Users/annthurium/github/atom/node_modules/react-dom/cjs/react-dom.production.min.js:206:333)
[...]

PR #1558, opened 2018-06-28
Fix uncaught exception when undoing coauthored commit

annthurium:
Fixes https://github.com/atom/github/issues/1538

When undoing a co-authored commit, we were passing a plain object instead of an `Author` model.  Since this object doesn't have a `matches` method, it blew up during rendering and shut down the whole tab.

bitmoji

Test plan:
- [x] manually test undoing a co authored commit
- [x] unit test to ensure that `Author` model is passed as a prop.

### DIFF
--- a/test/controllers/git-tab-controller.test.js
+++ b/test/controllers/git-tab-controller.test.js
@@ -713,13 +713,15 @@ describe('GitTabController', function() {
         const repository = await buildRepository(workdirPath);
         sinon.spy(repository, 'undoLastCommit');
         fs.writeFileSync(path.join(workdirPath, 'new-file.txt'), 'foo\nbar\nbaz\n');
+        const coAuthorName = 'Janelle Monae';
+        const coAuthorEmail = 'janellemonae@github.com';
 
         await repository.stageFiles(['new-file.txt']);
         const commitSubject = 'Commit some stuff';
         const commitMessage = dedent`
           ${commitSubject}
 
-          Co-authored-by: Foo Bar <foo@bar.com>
+          Co-authored-by: ${coAuthorName} <${coAuthorEmail}>
         `;
         await repository.commit(commitMessage);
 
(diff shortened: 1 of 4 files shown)


test/controllers/git-tab-controller.test.js
```diff
@@ -713,13 +713,15 @@ describe('GitTabController', function() {
         const repository = await buildRepository(workdirPath);
         sinon.spy(repository, 'undoLastCommit');
         fs.writeFileSync(path.join(workdirPath, 'new-file.txt'), 'foo\nbar\nbaz\n');
+        const coAuthorName = 'Janelle Monae';
```
annthurium: one thing I learned is that `fubar` is a military term for "effed up beyond all recognition."  Which uhhh, isn't an image I really want to call to mind, so I'm going to try to use more creative examples as much as I can.

annthurium:
@smashwilson: do you think there might be any other places hanging around where we're not using an `Author` model and should be?

test/controllers/git-tab-controller.test.js
smashwilson: Hah, sure, that's fair.

I've been tending to use `aaa / bbb / ccc` style placeholders, or suffixing with a one-up number (`user0`, `user1`). If I can I'll also try to name things after the role I'm intending them to play in the test (`Pull Request With An Empty Description`).

... But I get lazy sometimes too :wink:

test/controllers/git-tab-controller.test.js
```diff
@@ -745,10 +747,9 @@ describe('GitTabController', function() {
         commitMessages = wrapper.find('.github-RecentCommit-message').map(node => node.text());
         assert.deepEqual(commitMessages, ['Initial commit']);
 
+        const expectedCoAuthor = new Author(coAuthorName, coAuthorEmail);
         assert.strictEqual(wrapper.find('CommitView').prop('message'), commitSubject);
-        assert.deepEqual(wrapper.find('CommitView').prop('selectedCoAuthors'), [
-          {name: 'Foo Bar', email: 'foo@bar.com'},
-        ]);
+        assert.deepEqual(wrapper.find('CommitView').prop('selectedCoAuthors'), [expectedCoAuthor]);
```
smashwilson: It's really helpful that `deepEqual` works on object instances 😅

annthurium: YUP

annthurium:
> Hmm, taking a look at the coverage data, one notable thing is that it looks like we don't have any tests covering the new co-author form at all. So opening that, cancelling, and confirming a new co-author would be things it would be good to test.

I'm confused - it looks like the form states are tested in `co-author-form.test.js`.  Is the coverage report not picking this up?

annthurium:
> I'm confused - it looks like the form states are tested in co-author-form.test.js. Is the coverage report not picking this up? Or am I reading the report incorrectly?

nevermind - I see now that the report is saying we don't have any coverage of the co-author functionality in `CommitView`.  Added some tests.

PR #1574, opened 2018-07-10
Create a passwordless temporary keychain on macOS

### LINKED ISSUE #1385: Keytar tests fail when using security binary on CircleCI
I've seen this more often recently. It might have to do with CircleCI's build environment somehow?

smashwilson:
This _might_ fix #1385.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,11 @@ jobs:
     steps:
     - checkout
     - run:
+        name: create and set the default keychain
+        command: |
+          security create-keychain -p "" temporary
+          security default-keychain -s temporary
+    - run:
         name: download build-package.sh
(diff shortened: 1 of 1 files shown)


smashwilson:
Okay, I ran ten consecutive rebuilds of this and got a green build every time. I think I actually got it?

.circleci/config.yml
```diff
@@ -18,6 +18,11 @@ jobs:
 
     steps:
     - checkout
+    - run:
+        name: create and set the default keychain
+        command: |
+          security create-keychain -p "" temporary
+          security default-keychain -s temporary
```
smashwilson: @daviwil I bet this would fix our problems on VSTS too 🤔

daviwil: Sweet! I'll give it a shot today

PR #1579, opened 2018-07-12
Support arbitrary channel names when launching Atom Helper

daviwil:
### Description of the Change

This change modifies the `getAtomAppName` function to use a looser regex for identifying the channel name of the current Atom version.  This is necessary on macOS because we will now be calling the helper app `Atom Dev Helper.app` during CI builds against Atom master.

### Alternate Designs

Could have added `dev` to the existing channel name regex, but this seemed more future-proof.  Happy to go back to the old approach, though.

### Benefits

Less code changes to this function in the future?

### Possible Drawbacks

What happens if a weird Atom build with a different version format comes along?

### Applicable Issues

https://github.com/atom/atom/pull/17680

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -31,7 +31,7 @@ export function getPackageRoot() {
 }
 
 function getAtomAppName() {
-  const match = atom.getVersion().match(/-([a-z]+)(\d+|-)/);
+  const match = atom.getVersion().match(/-([A-Za-z]+)(\d+|-)/);
   if (match) {
     const channel = match[1];
     return `Atom ${channel.charAt(0).toUpperCase() + channel.slice(1)} Helper`;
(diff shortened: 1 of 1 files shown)


daviwil:
CI seems to be failing because of the security binary config issue

lib/helpers.js
```diff
@@ -31,9 +31,10 @@ export function getPackageRoot() {
 }
 
 function getAtomAppName() {
-  const match = atom.getVersion().match(/beta|nightly/);
+  const match = atom.getVersion().match(/-([a-z]+)(\d+|-)/);
```
smashwilson: No chance we'll use uppercase letters in our channel names?  :smile:

daviwil: I had it there at first but took it out because I figured it wouldn't happen.  I can add it back!

PR #1593, opened 2018-07-19
Restyle hovercards

simurai:
### Description of the Change

This restyles the "hovercards".

Before | After
--- | ---
screen shot 2018-07-19 at 10 01 56 pm | screen shot 2018-07-19 at 10 02 41 pm
screen shot 2018-07-19 at 10 02 57 pm | screen shot 2018-07-19 at 10 03 23 pm

### Alternate Designs

We could go closer to .com.

### Benefits

- Easier to read
- Long PR titles don't get cut off

Also, there is now a `github-Popover` class that can be re-used for other tooltips.

### Possible Drawbacks

Might not work with all themes the same.

### Applicable Issues

None, this is a direct order from @donokuda

### DIFF
--- a/lib/controllers/status-bar-tile-controller.js
+++ b/lib/controllers/status-bar-tile-controller.js
@@ -131,7 +131,7 @@ export default class StatusBarTileController extends React.Component {
           manager={this.props.tooltips}
           target={this.refBranchViewRoot}
           trigger="click"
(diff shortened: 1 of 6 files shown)


daviwil:
Looks amazing!  I love the subtle drop shadow

PR #1599, opened 2018-07-20
add accessbility linter

annthurium:
This pull requests adds the eslint plugin `jsx-a11y` so we can lint against common accessbility anti patterns.

Running this plugin with the recommended configuration, there were 84 errors.  I think it will be best to fix these incrementally to avoid a sprawling PR with a lot of merge conflicts. Two rules were easily fixable and seemed in scope for this pull request, though:

[Images must have alt text](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/alt-text.md)
[anchor elements must have an href](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md)

The only visible styling change is that the `RemoteSelectorView` now has buttons instead of links.  It looks like this:
screen shot 2018-07-20 at 10 04 19 am

### DIFF
--- a/lib/views/issueish-list-view.js
+++ b/lib/views/issueish-list-view.js
@@ -115,6 +115,7 @@ export default class IssueishListView extends React.Component {
   }
 
   renderMoreTile() {
+    /* eslint-disable jsx-a11y/anchor-is-valid */
     if (this.props.onMoreClick) {
       return (
         <div className="github-IssueishList-more">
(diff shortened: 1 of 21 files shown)


lib/views/issueish-list-view.js
```diff
@@ -114,6 +115,7 @@ export default class IssueishListView extends React.Component {
   }
 
   renderMoreTile() {
```
annthurium: this is actually a link - we're taking you out of atom and into dotcom.  (The other instances I found are really playing the role of a button, though.)

Issue #1237, opened 2017-10-30
Init should default to current project root


50Wliu:
### Prerequisites

* Have you tried to reproduce the problem in [Safe Mode][safe-mode]?
* Have you followed all applicable steps in the [debugging guide][debugging-guide]?
* Have you checked the FAQs on the [message board][message-board] for common solutions?
* Have you checked that your issue [isn't already filed][already-filed]?

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

### Description

Today I created a new folder and opened Atom in it, intending to start working on a new Git repository.  So I clicked on the 0 files icon and then tried to init a repository.  Instead of populating the location field with the project I had opened, it defaulted to `%USERPROFILE%\github`.

### Steps to Reproduce

1. Open a folder in Atom that is not a Git repository
2. Try to init through the github package

**Expected behavior:** Field should be pre-populated with current project path

**Actual behavior:** Populated with `C:\Users\<username>\github`

**Reproduces how often:** 100%

### Versions

```
Atom    : 1.23.0-dev-042c22f43
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

simurai:
Ran into this too. Here a gif:

![init](https://user-images.githubusercontent.com/378023/34548951-64056e3c-f148-11e7-80a2-d6833688701e.gif)

Having to enter the project name seems unexpected and in my case I just clicked on "Init" which then created a Git repo with all my projects in the parent `github` directory.

Would it be possible to skip the second step? So if you click on the "Create repository" button, it would initialize in the current directory without having to type the path.

50Wliu:
> Would it be possible to skip the second step? So if you click on the "Create repository" button, it would initialize in the current directory without having to type the path.

This seems logical.

smashwilson:
This was fixed in #1627 and we just didn't notice the issue link :smile:

PR #1627, opened 2018-08-03
🐛 Fix default dir for git repo init

Wittiest:
### Description of the Change
The linked issue details the unexpected default directory when using the atom sidebar to initialize a git repository. This PR ensures that the default directory is the working directory, as would be expected.

Before this change, the conditional on line 35 of [init-dialog.js](https://github.com/atom/github/blob/master/lib/views/init-dialog.js) would always set the default directory to the home directory.
### Alternate Designs
An in-depth look at why initDialogPath is null here could possibly provide a more elegant solution; however, this is concise enough to fix the issue without needing consideration for negative fallouts.

### Benefits
Users initializing a git repository with the sidebar will now have the default directory begin being set to the current working directory. This is what most users would expect, rather than it defaulting to the home directory.

### Possible Drawbacks
None

### Applicable Issues
https://github.com/atom/github/issues/1237

### DIFF
--- a/lib/controllers/root-controller.js
+++ b/lib/controllers/root-controller.js
@@ -410,6 +410,10 @@ export default class RootController extends React.Component {
       return null;
     }
 
+    if (!initDialogPath) {
+      initDialogPath = this.props.repository.getWorkingDirectoryPath();
+    }
+
     return new Promise(resolve => {
       this.setState({initDialogActive: true, initDialogPath, initDialogResolve: resolve});
     });


smashwilson:
Ah, excellent. Thanks, this had been bugging me 😄 

> An in-depth look at why initDialogPath is null here could possibly provide a more elegant solution; [..]

Here's the section that's _supposed_ to be populating the path argument:

https://github.com/atom/github/blob/a749da1dcc331bee4d7da730f9d220c2ac682708/lib/views/git-tab-view.js#L206-L217

But this will only work if you have an open TextEditor focused in your workspace center, which is not always the case. We could likely take that from `this.props.repository.getWorkingDirectoryPath()` instead... ?

> [..] however, this is concise enough to fix the issue without needing consideration for negative fallouts.

I agree. :shipit: :wink:

PR #1647, opened 2018-08-14
Update workflow

smashwilson:
Our workflow has drifted from the way we've documented it somewhat since we're in Editor Tools. Let's bring our markdown up to date with reality 💪

### DIFF
--- a/docs/how-we-work.md
+++ b/docs/how-we-work.md
@@ -26,7 +26,7 @@ This includes work like typos in comments or documentation, localized work, or r
 
 ##### Process
 
-1. Isolate work on a feature branch in the `atom/github` repository and open a pull request. Title-only pull requests are fine. If it's _really_ minor, like a one-line diff, committing directly to `master` is also perfectly acceptable.
+1. Isolate work on a feature branch in the `atom/github` repository and open a pull request. Remember to add the pull request to the current sprint board. Title-only pull requests are fine. If it's _really_ minor, like a one-line diff, committing directly to `master` is also perfectly acceptable.
 2. Ensure that our CI remains green across platforms.
 3. Merge your own pull request; no code review necessary.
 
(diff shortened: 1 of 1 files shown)


docs/how-we-work.md
```diff
@@ -114,20 +118,28 @@ When finalizing your review:
...
+At the end of each development sprint:
+
+1. _In your atom/github repository:_ run `apm publish preminor` to create the first prerelease version or `apm publish prerelease` to increment an existing prerelease version. Note the generated version number and ensure that it's correct. If the currently deployed version is `v0.19.2`, the first prerelease should be `v0.20.0-0`; if the existing prerelease is `v0.20.0-0`, the next prerelease should be `v0.20.0-1`.
+2. _In your atom/atom repository:_ create a new branch and edit `package.json` in its root directory. Change the version of the `"github"` entry beneath `packageDependencies` to match the prerelease you just published.
+3. _In your atom/atom repository:_ Run `script/build --install`. This will update Atom's `package-lock.json` files and produce a local development build of Atom with your prerelease version of atom/github bundled.
```
annthurium: should this be `script/build --ci --install`?

docs/how-we-work.md
```diff
@@ -114,20 +118,28 @@ When finalizing your review:
...
+  * :boom: _If the build fails,_ correct any bugs and begin again at (1) with a new prerelease version.
+4. Run `apm uninstall github` and `apm uninstall --dev github` to ensure that you don't have any [locally installed atom/github versions](/CONTRIBUTING.md#living-on-the-edge) that would override the bundled one.
```
annthurium: I forgot to `apm uninstall github` yesterday.  hmm, maybe that's my problem.

docs/how-we-work.md
smashwilson: Not in this case! `--ci` can speed up installation substantially, but it only works it `package-lock.json` is up to date, which it won't be here because we just edited `package.json`.

If anyone _else_ checks out this atom/atom pull request with the lockfile changes included, then they can add `--ci` to get a boost to their build time.

docs/how-we-work.md
smashwilson: That shouldn't cause the problem you're seeing. 🤔 The risk of not doing this is that you end up testing your linked / git-installed version of atom/github instead of the bundled and snapshotted one we're actually trying to verify.

docs/how-we-work.md
annthurium: ohhhh, that makes sense!  Thanks for clarifying.

PR #1651, opened 2018-08-16
send GitHub username with usage metrics

annthurium:
## Rationale

In order to plan a good roadmap for Atom's future, we need to have solid info about how users are interacting with Atom.

In the past we have anonymized all usage metrics.  Keeping the data anonymized impairs our ability to answer questions like:

- How does the Atom GitHub package fit in with GitHub.com workflows?
- How does the Atom GitHub package fit in with workflows in other GitHub tools (such as GitHub Desktop)?
- Does using Atom allow GitHub users to be more productive?

These questions are important. Therefore, we are introducing a change to pass the GitHub username with our usage metrics, if the user has authenticated with the GitHub package.

You do not have to authenticate with GitHub to use Atom, or the GitHub package.  Some functionality does require authenticating, though.

 If users don't want their GitHub username to be included in usability metrics, they can opt-out of sending their usage stats to GitHub.  Or they can choose not to authenticate.

## Approach
Use graphQL to fetch the GitHub username and send it to the `metrics` package.  The `ObserveModel` decorator wraps the login component so that we can re-fetch the data anytime the user performs a new login action.

## Alternate Approaches
We could send the authentication token to the metrics back end and do the username lookup there.  However, anytime you're sending tokens around, that's a security risk. Also, the approach I chose is more

### DIFF
--- a/lib/containers/__generated__/currentPullRequestContainerQuery.graphql.js
+++ b/lib/containers/__generated__/currentPullRequestContainerQuery.graphql.js
@@ -1,6 +1,6 @@
 /**
  * @flow
- * @relayHash 5821e0667d3f593dc75d2c4ac34985ca
+ * @relayHash 40b7bc7d6c17a665a1fe95116c6f6719
  */
 
 /* eslint-disable */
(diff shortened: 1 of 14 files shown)


Aerijo:
> If users don't want their GitHub username to be included in usability metrics, they can opt-out

Personally, I think identifiable metrics need to be _opt-in_ if you want to keep trust in the metrics package. I imagine people won't be happy if they discover it switched to sending non-anonymous data without their explicit acceptance.

annthurium:
@Aerijo Since this is a sensitive topic, we're going to discuss this at a maintainer's meeting sometime in the next couple of weeks. We'll let you know in advance that this in on the agenda.  Looking forward to hearing your perspective there.

Issue #1573, opened 2018-07-10
Discoverability


smashwilson:
I'd like to start a conversation about possible ways that we can improve discoverability of the git and GitHub features in Atom.

This is what we have in place presently:

(1) The welcome package includes buttons to open the Git and GitHub tabs, and a nudge referencing the status bar tile:

<img width="300" alt="screen shot 2018-07-10 at 8 14 54 am" src="https://user-images.githubusercontent.com/17565/42509293-732a2650-8419-11e8-81a9-6d90c71a4e1f.png">

(2) The hints that cycle when no panes are open in the workspace center includes mentions of the keyboard shortcuts:

<img width="300" alt="screen shot 2018-07-10 at 8 17 33 am" src="https://user-images.githubusercontent.com/17565/42509414-c59fa400-8419-11e8-9df3-f0a157385391.png">

(3) The status bar tiles themselves are always visible:

<img width="300" alt="screen shot 2018-07-10 at 8 18 33 am" src="https://user-images.githubusercontent.com/17565/42509453-e3adf0aa-8419-11e8-946a-11eb4f68ed77.png">

(4) The package's initial state. The first time that Atom is opened with the GitHub package installed, if the Welcome package is not shown, the right dock is opened with the Git and GitHub tabs present. Each time a new project is initialized, the Git and GitHub tabs are opened in the right dock but the right dock itself is not expanded.

(5) The "Packages" menu includes a "GitHub" entry, with the keyboard shortcuts listed:

<img width="300" alt="screen shot 2018-07-10 at 8 28 56 am" src="https://user-images.githubusercontent.com/17565/42509916-5e738d94-841b-11e8-915c-7cde411afbef.png">

But our feeling is that these are not sufficient:

[<img width="678" alt="screen shot 2018-07-10 at 8 22 21 am" src="https://user-images.githubusercontent.com/17565/42509624-876f585a-841a-11e8-9921-559e3959becd.png">](https://twitter.com/rauschma/status/1007103720830783489)

The GitHub tab is even more hidden: you have to know about the keyboard shortcut or the command palette entry to get to it.

Soon, thanks to @annthurium, we'll have access to metrics that we can use to validate this hypothesis, and measure the impact of any changes that we make to improve this 😁 Before we even have a baseline, though, I'd :heart: to start collecting ideas and brainstorming about different ideas we could explore.

The real trick is going to be threading a needle between letting users know that the package exists and how to get to it, without being obnoxious and annoying users who know about it but don't choose to use it.

[...]

PR #1658, opened 2018-08-22
add usage tracking for clicking the `ChangedFilesCount` component

annthurium:
### Description of the Change

In https://github.com/atom/github/issues/1573, we discussed discoverability of the Atom package.  @simurai proposed changing the status bar tile to say "Git" instead of "n file(s)".  I think that's a really good idea, but before implementing that change, I'd like to instrument this component and establish a baseline.  That way we'll know if there's a correlation between changing the text and how often it's clicked, even if causation is harder to establish. :-p 

This test also adds unit tests for `ChangedFilesCountView` since there previously were none. 

### Alternate Designs

Based on conversations with @telliott27, I'm using the `addEvent` api instead of the `incrementCounter` api.  In the past, client apps has used counters for feature usage.  However, counters don't retain metadata for each individual event, so `addEvent` gives us a fuller picture of what's going on.

### Benefits

The ability to understand how users are interacting with this component.  More broadly, I want to understand if users are having trouble discovering the github package.

### Possible Drawbacks

None that I can think of. 

### Applicable Issues

https://github.com/atom/github/issues/1573

### DIFF
--- a/lib/views/changed-files-count-view.js
+++ b/lib/views/changed-files-count-view.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Octicon from '../atom/octicon';
+import {addEvent} from '../reporter-proxy';
 
 export default class ChangedFilesCountView extends React.Component {
   static propTypes = {
@@ -15,6 +16,11 @@ export default class ChangedFilesCountView extends React.Component {
     didClick: () => {},
   }
 
+  handleClick() {
+    addEvent('click', {package: 'github', component: 'ChangedFileCountView'});
+    this.props.didClick();
+  }
+
   render() {
     const label =
       (this.props.changedFilesCount === 1)
@@ -24,7 +30,7 @@ export default class ChangedFilesCountView extends React.Component {
       <button
         ref="changedFiles"
         className="github-ChangedFilesCount inline-block"
-        onClick={this.props.didClick}>
+        onClick={this.handleClick.bind(this)}>
         <Octicon icon="diff" />
         {label}
         {this.props.mergeConflictsPresent && <Octicon icon="alert" />}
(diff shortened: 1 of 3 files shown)


lib/views/changed-files-count-view.js
```diff
@@ -24,7 +30,7 @@ export default class ChangedFilesCountView extends React.Component {
       <button
         ref="changedFiles"
         className="github-ChangedFilesCount inline-block"
-        onClick={this.props.didClick}>
+        onClick={this.handleClick.bind(this)}>
```
annthurium: @smashwilson : would it be better to use `autobind` here for consistency's sake even though there's only one method to bind and no constructor in this component?

annthurium: @smashwilson thanks for the extra context!  I was just being lazy. :-p . I'll fix it.

smashwilson:
> This pull request also adds unit tests for ChangedFilesCountView since there previously were none.

Yesss 🥇 

So somewhat awkwardly, a bunch of the tests for the status bar components are smashed up into [`test/controllers/status-bar-tile-controller.test.js`](https://github.com/atom/github/blob/dfcfc645cabfcf59ff5741b7b7fa63f413aa9f4a/test/controllers/status-bar-tile-controller.test.js). There _are_ a few for the change files view:

https://github.com/atom/github/blob/dfcfc645cabfcf59ff5741b7b7fa63f413aa9f4a/test/controllers/status-bar-tile-controller.test.js#L649-L680

But they're pretty sparse and quite old. I'd say delete those in favor of yours ✂️

annthurium:
@smashwilson some of the tests in `status-bar-tile-controller.test.js` are exercising functionality that couldn't be tested by the `ChangedFilesCountView` in isolation, without breaking encapsulation.  For example, testing that clicking the `ChangedFilesCountView` actually toggles the tab, or testing that the status tile icons don't show up if the project lacks a repository.

There is one test in `status-bar-tile-controller` that does feels redundant tho, so I'll remove that one.

PR #1512, opened 2018-06-06
Render file patches with a decorated Editor

smashwilson:
Use decorations within an Atom TextEditor element to implement a file patch item. This should get us improved performance for large diffs, editable diff support, and open the path for multi-file diff views like a commit or pull request pane.

Just to set proper expectations, I'll likely need to pause and resume this a few times before it lands. Don't worry, I won't forget about it :smile:

##### Remaining Work

- [x] Split `FilePatchController` out into a `FilePatchItem`, `FilePatchContainer`, `FilePatchController`, and `FilePatchView`, as per #1436.
- [x] Inherit basic item wiring (`getTitle()`, etc) onto `FilePatchItem`.
- [x] Render an `<AtomTextEditor>` within the `FilePatchView`.
- [ ] Model: render a `FilePatch` to the unadorned text that should be displayed within the editor accompanied by the buffer line ranges that correspond to each hunk and line range
- [ ] View: render each group of controls from the original FilePatchView within an appropriately placed Decoration
  - [ ] CSS tweaks to keep it looking right
- [ ] Controller: port interactions over and wire 'em up to the right controls
  - [ ] Figure out how to preserve selection semantics
- [ ] **Upstream in Atom**
  - [ ] Support custom line numberings in the line number gutter
  - [ ] Support the creation of multiple line number gutters
- [ ] Catch test coverage up to the speculative work. Get us ✅ again.
- [ ] Measure performance difference on large diffs and see where our limit is now
  - [ ] Re-introduce the "large diff" gate

###### Stretch goals

- [ ] Make the editor editable. Alter the `FilePatch` accordingly.
- [ ] Add syntax highlighting by setting the editor's grammar.

Fixes #1502. ...Eventually 🕐

PR #1664, opened 2018-08-23
Skip all tests when engines.atom has been bumped to exclude the current Atom version

smashwilson:
In #1512, I'm introducing a dependency on functionality that was added to Atom in atom/atom#17736, which is currently only available in nightly or dev builds of Atom. To capture this, I want to be able to bump the semver range specified in the `.engines.atom` field of our `package.json` to document the minimum supported Atom version, but our CI matrix rows for stable and beta will fail for _everything_ until that change makes it way to a stable Atom release.

This adds a check for compatibility between the declared Atom version range and the current host Atom range. If the Atom instance running the test suite is not expected to succeed because it's out of that range, we skip all tests. This will result in a "passing" test suite.

### DIFF
--- a/test/runner.js
+++ b/test/runner.js
@@ -79,7 +79,7 @@ module.exports = createRunner({
   htmlTitle: `GitHub Package Tests - pid ${process.pid}`,
   reporter: process.env.MOCHA_REPORTER || 'list',
   overrideTestPaths: [/spec$/, /test/],
-}, mocha => {
+}, (mocha, {terminate}) => {
   // Ensure that we expect to be deployable to this version of Atom.
   const engineRange = require('../package.json').engines.atom;
   const atomEnv = global.buildAtomEnvironment();
@@ -93,9 +93,8 @@ module.exports = createRunner({
       `This version of atom/github is currently incompatible with the ${atomReleaseChannel} ` +
       'Atom release channel.\n',
     );
-    // A regular expression that matches nothing.
-    mocha.grep(/.^/);
-    return;
+
+    terminate(0);
   }
 
   const Enzyme = require('enzyme');
--- a/test/runner.js
+++ b/test/runner.js
@@ -5,6 +5,7 @@ import path from 'path';
 
 import until from 'test-until';
 import NYC from 'nyc';
+import semver from 'semver';
 
 chai.use(chaiAsPromised);
 global.assert = chai.assert;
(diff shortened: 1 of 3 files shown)


test/runner.js
```diff
@@ -79,6 +80,24 @@ module.exports = createRunner({
...
+  const atomReleaseChannel = atomEnv.getReleaseChannel();
+  atomEnv.destroy();
+
+  if (!semver.satisfies(semver.coerce(atomVersion), engineRange)) {
+    process.stderr.write(
+      `Atom version ${atomVersion} does not satisfy the range "${engineRange}" specified in package.json.\n` +
+      `This version of atom/github is currently incompatible with the ${atomReleaseChannel} ` +
+      'Atom release channel.\n',
+    );
+    // A regular expression that matches nothing.
```
annthurium: just for my own education: why do we need a regex that matches nothing?

smashwilson: Well... I'm trying to find a graceful way to "cancel" the current test run and exit with a 0 status from the test runner. Mocha's `.grep()` method lets you filter the tests that execute by matching their titles against a regular expression, so providing it a regexp that will never match is equivalent to saying "don't run anything."

It's kind of awkward though. And I think it causes some misleading output from Mocha, although I tried to mitigate that with the `process.stderr` lines up above.

I tried doing a `process.exit(0)` first, but that caused the main process to think the renderer process had crashed and exit with a 1. Before that I was going to do this in a standalone script called from `script/cibuild`, but AppVeyor and CircleCI are both using the [atom/ci build scripts](https://github.com/atom/ci) which are hard-coded to run `atom --test test/`, so the test runner is the earliest common place to put the version check logic.

Now I'm thinking the Right Thing To Do :tm: is to do a little work upstream in atom-mocha-test-runner to support a clean early exit.

smashwilson:
Here we go:

screen shot 2018-08-24 at 9 02 48 am

screen shot 2018-08-24 at 9 03 41 am

screen shot 2018-08-24 at 9 20 16 am

I didn't end up messing with the status or checks APIs to report "you can't deploy to this channel now" after all, though. It would take a moderate amount of work to put in place, we wouldn't be able to see it where we really need it (on the release branches when we're tagging a version!), and it'd be obsolete if we move into atom/atom eventually anyway.

We'll just need to remember which versions of atom/github rely on newer Atom features, by leaving a comment on the Atom ship-ready issue maybe? In any case it hasn't come up that often.

@annthurium what do you think?

annthurium:
@smashwilson I'm totally okay with the lower-effort version of this - agree that it hasn't arisen often since usually we can programmatically check for new functionality in atom core.

PR #1667, opened 2018-08-24
Handle staged changes with special characters in the filename

annthurium:
### Description of the Change
when a user clicks on a changed file in the `StagingView`, the github package creates a uri to pass to `atom.workspace.open()` so the user can open the file and view its contents.

However, we were not properly encoding all components of this URI.  Which is only a problem if there are special characters in the filename that need to be escaped.  In this case, an invalid URI would be generated, Atom would look for a file that doesn't exist, and show an error state thinking that the entire repository had only been a dream.

This PR addresses this issue by URI encoding the relative path to the file.

### Applicable Issues

https://github.com/atom/github/issues/1610

### DIFF
--- a/test/controllers/file-patch-controller.test.js
+++ b/test/controllers/file-patch-controller.test.js
@@ -96,6 +96,18 @@ describe('FilePatchController', function() {
 
       getFilePatchForPath = sinon.stub(repository, 'getFilePatchForPath');
     });
+    describe('buildURI', function() {
+      it('correctly uri encodes all components', function() {
+        const filePath = '???.txt';
+        const stagingStatus = 'staged';
+        const URI = FilePatchController.buildURI(filePath, workdirPath, stagingStatus);
+        assert.isTrue(URI.includes(encodeURIComponent(filePath)));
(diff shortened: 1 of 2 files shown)


test/controllers/file-patch-controller.test.js
```diff
@@ -96,6 +96,18 @@ describe('FilePatchController', function() {
 
       getFilePatchForPath = sinon.stub(repository, 'getFilePatchForPath');
     });
+    describe('buildURI', function() {
```
annthurium: @smashwilson do you think it would be worth it to also add tests at the react component layer?   IE, clicking on staged changes and verifying that the filepath is correct?  Or is this sufficient?

smashwilson: This should be fine. I'd say that in this case, _good test coverage of the `buildURI()` method with wacky URIs + good test coverage of launching pane items = good test coverage of launching pane items with wacky URIs_. In other words, adding an integration-ish test for this particular loophole wouldn't really catch anything that that wouldn't.

:shipit:

annthurium: bitmoji

thank you!

Issue #1656, opened 2018-08-20
Multiple tabs within an IssueishPaneItem


smashwilson:
To pave the way for pull request review display, let's split the IssueishPaneItem's timeline into a series of tabs, each containing a subview of the pull request:

* Description
* Timeline (the existing view, with non-review comments, push events, and so on)
* Commits
* Changes: the full diff of the pull request, with files navigable as in #1655

We can render them horizontally in the workspace center and accordion-style when docked.

smashwilson:
> I'm curious about the reasoning behind separating "Description" and "Timeline" as opposed to showing them in the same tab

In my original write-up? Totally arbitrary, haha. I think it was mostly to have a nice even four items.

Do we just want to take the conversation / commits / checks / files changed headers from dotcom, and then add in reviews later? It might be jarring to start synced up and then immediately deviate. But I do kind of like that breakdown, and it's be nice to have that instant familiarity with what users would expect to see where.

> Reversing the order is an interesting suggestion. I'm not sure which I prefer at this point. This is definitely something I'd like to get some UXR data on.

Me neither! I kind of default to wanting to start consistent in order as well?

annthurium:
ok so another question...if we go for Conversations, Commits, Checks, Files Changed, are we going to support the Checks api since that's what lives on that tab in dotcom?  Or do we just stuff the CI status checks that already exist in GitHub package on that tab?

simurai:
> it's be nice to have that instant familiarity with what users would expect to see where.

Yeah, the familiarity is a big plus. At the same time it could lead to confusion and frustration. If people see the same navigation as on .com, they might expect the same functionality when clicking the tabs. People will be like "Ohh.. right. I can't do this here" and the more this happens, the less they trust these tabs. If we copy .com, we kinda have to copy 100% of it. Like putting an iframe in the center pane. So by **not** doing the same as on .com, we can better manage expectations and it doesn't feel strange if some parts are missing or are different.

This question about how close should it be to .com comes up from time to time, and we haven't really found a clear answer. Lately I started to lean more towards not trying to make it close to .com. Mainly because of our team size, it seems impossible to ever catch up. But also if we don't copy .com 1:1, it leaves some room to do things that are not possible on .com and only make sense in the context of an editor. I know, much easier said than done. :sweat_smile:

smashwilson:
[...]

PR #1684, opened 2018-09-07
Add tabs for issueish panel items

annthurium:
### Description of the Change

This change moves some of the information from the pull request view into tabs.

### Screenies
screen shot 2018-09-20 at 1 53 39 pm
screen shot 2018-09-20 at 1 53 52 pm

docked:
screen shot 2018-09-20 at 3 06 26 pm

### TODO
- [x] Put a border around the active/selected tab
- [x] get some input from team about tab content / ordering, make decision
- [x] Add octicons to tabs
- [x] Make the tabs render accordion style when the pane is docked.
- [x] Add unit tests
- [x] Investigate whether only the active tab should render first for performance reasons.
- [x] add gifs and screenshots to pr body
- [x] add content to `Commits` tab.
- [x] Clean up issue rendering
- [x] handle pagination for PRs with >100 commits

### Alternate Designs

- We had some back and forth about what information goes on each tab.  @simurai's initial designs called for `description`, `timeline`, `commits`, and `changes` as the tab items.  We had some back and forth about whether it might be better to use consistent language with what's used on dotcom.  See the issue for details.  Ultimately it would be best to validate our assumptions with uxr before we ship.

- I could have rolled my own tabs instead of using a lib.  Although it sounded like fun, I thought it might be a more efficient use of time to not reinvent wheels.  This lib is small and well maintained and uses good a11y practices.

### Benefits

Separating the information into tabs gives us room for pull request reviews in the editor, which is where all this is headed.

### Possible Drawbacks

It's possible that users will find the new UI confusing since we're moving information around.  

### Applicable Issues

https://github.com/atom/github/issues/1656

### Test plan

- [x] manually test rendering an issue to verify that there are no visual regressions.
- [x] unit tests for new `PrCommitView` component to verify that:
  - show / hide more buttons show and hide commit message body
  - show / hide more buttons are only rendered if there is a commit message body
  - renders link to dotcom, author name and avatar, commit headline, humanized time since the commit happened
- [x] unit tests for `IssueishDetailView` to test that it renders 3 tabs with the right text and icons
- [x] unit tests for `PrCommitsView` to verify that:
   - n commits are rendered if n commits are passed in
   - load more button is only rendered if there's more commits to be loaded
   - clicking loadMore calls the function that loads more stuff

### TODOs for future prs:
- display co-author avatars
- refactor header rendering in `IssueishDetailView` to improve readability
- figure out why keyboard nav doesn't work and fix it


annthurium:
@simurai a couple of thoughts:

1) thanks for taking a pass at this!

2) I cannot read the text on the active tab with my theme.  How do we work around these kinds of theme problems?
screen shot 2018-09-13 at 4 29 46 pm

3) Can we keep the button text as "show more" and "hide more" instead of ellipses?  Accessibility guidelines for designing for users with cognitive impairments tell me that buttons should say what they actually do.  One source for this information (that's actually a really beautifully designed poster): https://github.com/UKHomeOffice/posters/blob/master/accessibility/dos-donts/posters_en-UK/svg/autistic-spectrum.svg

4) I'm having a hard time visually connecting which commits go with which sha on the right, without the borders.  Is there something else we could do to address this?  

5) Also, without the borders on the tabs, it's not clear to me that they are tabs.  

6) Thanks for fixing the text breaking on long words.

simurai:
### 3.

That's a nice infographic. :heart_eyes_cat: My thinking was that people _should_ be familiar with the ellipsis icon becaused .com uses a similar button.

image

But it's totally ok to not copy .com 100%. Maybe we can add a border/background color for the text label, so it looks different enough from the title. I'll give that a try.

### 2.

What UI theme are you using? The active tab uses `@text-color-selected` color. I assume it's white in your theme because the `@background-color-selected` is probably dark. We could give `@text-color-highlight` a try and see if it has enough contrast. The `atom-light-ui` theme has the same issue. Or not change the color for the active tab, but I thought it might be a good idea to use something else besides the border. Which leads us to..

### 4. + 5.

The borders are using this variable: `@base-border-color`. It looks like your theme has set it to transparent or the same color as the background color. It might be that the theme author didn't like a border somewhere and then instead of overriding it only there, globally tried to make all borders invisible.

Finding the right variable to make it look good in all (most) themes is a constant challenge. But it's also our fault because we didn't document what each variable means and show examples how it gets used. The variable name is the :see_no_evil: **only** documentation and it's up to theme authors to decide how to interpret those names.

Anyways, I can make an issue for that theme. Bundled themes look like this:

One UI | Atom UI
--- | ---
one-light-ui | atom-light-ui
one-dark-ui | atom-dark-ui

### 7.

Yeah, I was also thinking if we should keep commits consistent throughout the package or if recent commits should just be a compact version. We might can answer that once we have a [log view](https://github.com/atom/github/pull/1419). And I agree, the commits are not _that_ useful yet. And on .com I rarely look at the commits tab and "Files Changes" is most of the time enough, especially because you can filter the diff by commit. Should we still keep it for now? Be able to see the commits and click to .com could still be useful in some situations and having 3 tabs is more balanced. The biggest risk is adding it now and then taking it away or change it later again.

annthurium:
@kuychaco, as far as what kind of feedback I'm looking for in code review:
- did I follow all the graphql conventions we've set in the package?  Or are there things that could be improved?
- Are there ways I could refactor to enhance readability?

simurai:
Something that came to mind when looking at the many commits of https://github.com/atom/github/pull/1512 in the issueish pane: Should the **order be reversed**? So that new commits appear at the top and you don't have to scroll down and potentially click on "load more"  a few times, just to see if new commits are made. The timeline also has the same issue.

We don't have to change it now, just something to consider one day.

lib/views/issueish-detail-view.js
```diff
@@ -105,6 +107,85 @@ export class BareIssueishDetailView extends React.Component {
...
+        <TabPanel>
+          <GithubDotcomMarkdown
+            html={issueish.bodyHTML || '<em>No description provided.</em>'}
+            switchToIssueish={this.props.switchToIssueish}
+          />
+          {this.renderEmojiReactions(issueish)}
+          <PrTimelineContainer
+            {...childProps}
+            switchToIssueish={this.props.switchToIssueish}
+          /> :
```
simurai: There is a `:` at the bottom left in the Overview tab:

screen shot 2018-09-21 at 9 31 58 pm

Is it the ☝️ above `:` that could be removed?

annthurium: oh, that's a mistake!  Thanks for pointing it out.  Will remove.

lib/views/pr-commits-view.js
```diff
@@ -0,0 +1,143 @@
...
+    ) {
+      url
+      commits(
+        first: $commitCount, after: $commitCursor
+      ) @connection(key: "prCommitsView_commits") {
+        pageInfo { endCursor hasNextPage }
+        edges {
+          cursor
+          node {
+            commit {
```
kuychaco: This is just a nice to have -- we could extract this as a fragment on PrCommitView. See [timeline/commits-view.js](https://github.com/atom/github/blob/master/lib/views/timeline-items/commits-view.js#L85) and [timeline/commit-view.js](https://github.com/atom/github/blob/master/lib/views/timeline-items/commit-view.js#L69-L86) as an example.

This keeps our query fragments co-located with the views that use the resulting data. I think this may also allow relay to do some extra checking for us to make sure that data from our queries match with what is actually being used.

test/integration/checkout-pr.test.js
```diff
@@ -161,6 +163,7 @@ describe('check out a pull request', function() {
     });
   }
 
+  // achtung! this test is flaky
```
kuychaco: hmm weird. Any idea why? 🤔

test/views/issueish-timeline-view.test.js
```diff
@@ -109,7 +109,7 @@ describe('IssueishTimelineView', function() {
     const wrapper = shallow(buildApp({
       relayHasMore: () => false,
     }));
-    assert.isFalse(wrapper.find('.github-PrTimeline-load-more-link').exists());
+    assert.isFalse(wrapper.find('.github-PrTimeline-loadMoreButton').exists());
```
kuychaco: Not sure about the appropriate convention here, just noticing that we've switched from hyphen-separated to camelCase... same below.

test/views/pr-commit-view.test.js
```diff
@@ -0,0 +1,69 @@
...
+import {shallow} from 'enzyme';
+
+import PrCommitView from '../../lib/views/pr-commit-view';
+
+const defaultProps = {
+  committerAvatarUrl: 'https://avatars3.githubusercontent.com/u/3781742',
+  committerName: 'Margaret Hamilton',
+  date: '2018-05-16T21:54:24.500Z',
+  messageHeadline: 'This one weird trick for getting to the moon will blow your mind 🚀',
+  abbreviatedOid: 'bad1dea',
```
kuychaco: 😆

kuychaco:
> Should the order be reversed?

Oh oops. Yeah the order should totally be recent items first

test/views/issueish-timeline-view.test.js
annthurium: I think @simurai made that change - I'm happy to switch to kabob-case for the sake of consistency.

annthurium:
@kuychaco @simurai dotcom renders the newest commits at the bottom of the commits tab.  While we don't have to stay consistent with dotcom in every way, I think that reversing the order here might be confusing to users.  I'm willing to test this with uxr and revisit it, though!

lib/views/issueish-detail-view.js
```diff
@@ -113,6 +194,9 @@ export class BareIssueishDetailView extends React.Component {
       issue: issueish.__typename === 'Issue' ? issueish : null,
       pullRequest: issueish.__typename === 'PullRequest' ? issueish : null,
     };
+    // todo(tt, 9/2018): it could enhance readability to extract header rendering into
+    // 2 functions: one for rendering an issue header, and one for rendering a pr header.
+    // however, the tradeoff there is having some repetitive code.
```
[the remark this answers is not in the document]
annthurium: good suggestion!  Let's do that.

test/views/issueish-timeline-view.test.js
simurai: The [naming convention](https://github.com/atom/github/blob/master/docs/naming-convention.md#child-element) *would* be camelCase for child elements of a component. So it can easily be read in groups: `github-PrTimeline-loadMoreButton` -> It's the `loadMoreButton` inside the `PrTimeline` component inside the `github` package. We don't follow that convention everywhere, which isn't a big deal.

Issue #1591, opened 2018-07-18
Instrument allll the things


annthurium:
Here are some things we could instrument to learn more about how users are interacting with the GitHub package:

- [ ] Staging.  IMO this is one killer feature of using git inside Atom vs on the command line. Are users doing partial staging of hunks and lines with the context menu?  With the buttons at the top of the diff view?  With keyboard shortcuts?  Or are they just smashing the "Stage All" and "Unstage All" buttons?

- [ ] Measure system latency of Git operations.  How long do these things actually take in the real world?

I'm sure there are more things.  What user behaviors would you like to understand better?

smashwilson:
Is there anything meaningful we could measure about the graphical merge conflict resolution widgets? Something like:

- [ ] How much is the graphical conflict resolution being used?
  - [ ] Count of conflicting changes resolved with the editor decorations, maybe as a percentage of times that merge conflicts are observed.
  - [ ] How many people have disabled it altogether in their package settings?
  - [ ] How often do people dismiss the conflict decorations to resolve them manually?

Also: how about GPG signed commits?

- [ ] How many GPG-signed commits are created with our package?
  - [ ] How many of those successfully use the Atom prompt helper?

Maybe some metrics related to remote protocols and authentication?

- [ ] What's the breakdown of protocols used by git remotes - https, ssh, git, ... ?
- [ ] How many remote operations trigger the in-Atom prompt helper?

---

> What does this instrument mean? Collect usage data?

Yah - thanks to a ton of work from @annthurium, we have a way to [collect telemetry from Atom usage](https://github.com/atom/telemetry) from those who consent on the welcome pane and send it to our internal GitHub metrics pipeline (instead of Google Analytics). Our goal is to have data we can work with to make decisions about what work to prioritize, identify discoverability gaps in our UX, and things like that.

the-j0k3r:
>Also: how about GPG signed commits?

 >   How many GPG-signed commits are created with our package?

Mine all

 >       How many of those successfully use the Atom prompt helper?

none

annthurium:
oh, I can't believe I didn't add think of this initially, but we should instrument all the entry points for the GitHub package.  (Keyboard shortcuts, menus, clicking on the tile in the status bar).

kuychaco:
Another candidate for instrumenting: undoing the last commit

kuychaco:
Ooh, and discarding/undoing discards

dwelle:
[...]

PR #1685, opened 2018-09-08
Add metrics collection for committing partial changes, undoing the last commit, and opening an expanded commit msg editor

kuychaco:
### Description of the Change

This PR introduces some of the additional metrics discussed in #1591. 

So far I've added events for the following:
* undo last commit 
  * _I use this feature A LOT. would be interesting to see how frequently others use it._
* commit, with metadata indicating:
  * if the commit included partial changes (that is, unstaged changes remained uncommitted in the working directory) 
    * _are people taking advantage of how easy it is to stage/unstage subsets of changes in Atom? what percentage of total commits commit partial changes? As we make changes to our diff view and implement other discoverability/usability improvements, how will this be affected?_ 
  * the number of co-authors for the commit 
    * _what percentage of users are committing with co-authors? as we ship more features that promote collaboration how does this change?_

Other thoughts and considerations:

**Expanded commit message editor usage**
I considered using the "verbatim" option passed to Repository#commit to report whether the commit was made from an expanded commit editor or the small commit box in the Git Panel because I know that is what dictates its value, but it seemed icky to be relying on that inference. It also felt not great to pass down an additional option such as `expandedCommitMessageEditor` to the repository model strictly for collecting metrics on it. If we decide this is something we really want to instrument, it's better and more direct to do so where the expanded editor is opened (such as CommitController#openCommitMessageEditor).

**Gathering data on button use vs. keyboard shortcuts vs. context menu use**
It could be interesting and useful to gather data on how users are initiating operations in our package. This info could yield insights about how discoverable each of these entry points are. That said, the way that keybindings and context-menus are set up in Atom is based on commands which result in functions being called. We can easily instrument the function calls in this package as well as button presses, but I don't think there's an easy/straight-forward way to instrument the keyboard shortcut or context menu usage. @smashwilson @annthurium 💭s?

### Alternate Designs

To gather information on partial-staging I considered tracking staging/unstaging events. At the end of the day though, I think what we care most about is how frequently users are taking advantage of the fact that you can easily make commits with partial changes using our UI. Tracking all of the staging/unstaging events seems like it would be a lot of noise with little additional value. Open to thoughts and suggestions though!

### Benefits

We'll have interesting insights into user behavior, and be able to ask/answer questions such as those above in italics. This information could potentially help us identify UX issues, inform project decisions and priorities, and track the impact of future work and features introduced.

### Possible Drawbacks

This PR simply introduces additional metrics collection. No drawbacks that I can think of. 

### Applicable Issues

#1591

TODO: 
- [x] Possibly instrument commit editor expansion

### DIFF
--- a/lib/models/repository-states/present.js
+++ b/lib/models/repository-states/present.js
@@ -17,6 +17,7 @@ import Remote from '../remote';
 import RemoteSet from '../remote-set';
 import Commit from '../commit';
 import OperationStates from '../operation-states';
+import {addEvent} from '../../reporter-proxy';
 
 /**
  * State used when the working directory contains a valid git repository and can be interacted with. Performs
@@ -344,6 +345,7 @@ export default class Present extends State {
       async () => {
         try {
           await this.git().reset('soft', 'HEAD~');
+          addEvent('undoLastCommit', {package: 'github'});
         } catch (e) {
           if (/unknown revision/.test(e.stdErr)) {
             // Initial commit
--- a/lib/models/repository-states/present.js
+++ b/lib/models/repository-states/present.js
@@ -250,6 +250,7 @@ export default class Present extends State {
         addEvent('commit', {
           package: 'github',
           partial: unstagedCount > 0,
+          amend: !!options.amend,
           coAuthorCount: coAuthors ? coAuthors.length : 0,
         });
         // note: this is somewhat redundant to the `incrementCounter` call in `GitShellOutStrategy` on all git commands,
--- a/lib/models/repository-states/present.js
+++ b/lib/models/repository-states/present.js
@@ -234,7 +234,7 @@ export default class Present extends State {
         Keys.branches,
       ],
       // eslint-disable-next-line no-shadow
-      () => this.executePipelineAction('COMMIT', (message, options) => {
+      () => this.executePipelineAction('COMMIT', async (message, options) => {
         const opts = (!options || !options.coAuthors)
           ? options
           : {
@@ -244,7 +244,16 @@ export default class Present extends State {
             }),
           };
 
-        return this.git().commit(message, opts);
+        await this.git().commit(message, opts);
+        const {unstagedFiles, mergeConflictFiles} = await this.getStatusesForChangedFiles();
+        const unstagedCount = Object.keys({...unstagedFiles, ...mergeConflictFiles}).length;
+        const {amend, expandedCommitMessageEditor} = opts || {};
+        addEvent('commit', {
+          package: 'github',
+          partial: unstagedCount > 0,
+        });
+        // note: this is somewhat redundant to the `incrementCounter` call in `GitShellOutStrategy` on all git commands,
+        // here we get useful additional metadata
       }, message, options),
     );
   }
@@ -530,9 +539,9 @@ export default class Present extends State {
           this.transitionTo('TooLarge');
           return {
             branch: {},
-            stagedFiles: [],
-            unstagedFiles: [],
-            mergeConflictFiles: [],
+            stagedFiles: {},
+            unstagedFiles: {},
+            mergeConflictFiles: {},
           };
         } else {
           throw err;
--- a/lib/models/repository-states/present.js
+++ b/lib/models/repository-states/present.js
@@ -235,22 +235,22 @@ export default class Present extends State {
       ],
       // eslint-disable-next-line no-shadow
       () => this.executePipelineAction('COMMIT', async (message, options) => {
-        const opts = (!options || !options.coAuthors)
-          ? options
-          : {
-            ...options,
-            coAuthors: options.coAuthors.map(author => {
-              return {email: author.getEmail(), name: author.getFullName()};
-            }),
-          };
+        // check with Ash that I retained the logic he was going for...
+        const coAuthors = options && options.coAuthors;
+        const opts = !coAuthors ? options : {
+          ...options,
+          coAuthors: coAuthors.map(author => {
+            return {email: author.getEmail(), name: author.getFullName()};
+          }),
+        };
 
         await this.git().commit(message, opts);
         const {unstagedFiles, mergeConflictFiles} = await this.getStatusesForChangedFiles();
         const unstagedCount = Object.keys({...unstagedFiles, ...mergeConflictFiles}).length;
-        const {amend, expandedCommitMessageEditor} = opts || {};
         addEvent('commit', {
           package: 'github',
           partial: unstagedCount > 0,
+          coAuthorCount: coAuthors ? coAuthors.length : 0,
         });
         // note: this is somewhat redundant to the `incrementCounter` call in `GitShellOutStrategy` on all git commands,
         // here we get useful additional metadata
--- a/lib/models/repository-states/present.js
+++ b/lib/models/repository-states/present.js
@@ -235,7 +235,6 @@ export default class Present extends State {
       ],
       // eslint-disable-next-line no-shadow
       () => this.executePipelineAction('COMMIT', async (message, options) => {
-        // check with Ash that I retained the logic he was going for...
         const coAuthors = options && options.coAuthors;
         const opts = !coAuthors ? options : {
           ...options,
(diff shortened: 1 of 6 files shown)


lib/models/repository-states/present.js
```diff
@@ -233,17 +234,27 @@ export default class Present extends State {
...
+          }),
+        };
 
-        return this.git().commit(message, opts);
+        await this.git().commit(message, opts);
+        const {unstagedFiles, mergeConflictFiles} = await this.getStatusesForChangedFiles();
+        const unstagedCount = Object.keys({...unstagedFiles, ...mergeConflictFiles}).length;
+        addEvent('commit', {
+          package: 'github',
+          partial: unstagedCount > 0,
```
annthurium: Since there are several UI options for partial staging, it would be interesting to instrument those as well.  Not something you have to implement here, just something to consider for future iterations.

lib/models/repository-states/present.js
```diff
@@ -344,6 +355,7 @@ export default class Present extends State {
       async () => {
         try {
           await this.git().reset('soft', 'HEAD~');
+          addEvent('undoLastCommit', {package: 'github'});
```
annthurium: nit: for the counter names, we're using dash separation.  Could you make this `undo-last-commit` instead?

lib/models/repository-states/present.js
```diff
@@ -17,6 +17,7 @@ import Remote from '../remote';
 import RemoteSet from '../remote-set';
 import Commit from '../commit';
 import OperationStates from '../operation-states';
+import {addEvent} from '../../reporter-proxy';
```
annthurium: one thing I worry about is, we currently have some metrics tracking in git-shell-out-strategy.  Which is basically as close to the "back end" as you can go.  We also have tracking on the UI layer, which is useful for seeing how many users are clicking specific buttons and whatnot.  If we are adding instrumentation to the middle layers of the application, how do we ensure we're not double counting events? Would it make more sense to move this instrumentation down a layer to keep it all in one place?

lib/models/repository-states/present.js
```diff
@@ -528,9 +540,9 @@ export default class Present extends State {
           this.transitionTo('TooLarge');
           return {
             branch: {},
-            stagedFiles: [],
-            unstagedFiles: [],
-            mergeConflictFiles: [],
+            stagedFiles: {},
```
annthurium: just for my own education: why did you change this function to return objects instead of arrays?  (I'm sure there's a perfectly good reason, I just don't know what it is 😃 )

annthurium:
a few more thoughts on the questions you raised in the body of the pull request:

- I agree that it would be icky to pass extra meta data about `expandedCommitMessageEditor` to the repository just for metrics tracking.  It looks like the only way to toggle the expanded commit editor is through that button, though?  You could instrument the button to get a sense of how often users are toggling it on.  If it's very occasional, we can safely assume this is not a popular feature.

- I haven't come up with a great strategy for instrumenting context menu actions as of yet.  That would be a good thing to add. At a glance, perhaps we could leverage the `ContextMenuInterceptor` component and just track everything we're intercepting.

test/models/repository.test.js
```diff
@@ -597,6 +599,116 @@ describe('Repository', function() {
...
+        // do whole commit
+        assert.deepEqual(reporterProxy.addEvent.callCount, 0);
+        await repo.commit('Commit everything');
+        assert.deepEqual(reporterProxy.addEvent.callCount, 1);
+        args = reporterProxy.addEvent.lastCall.args;
+        assert.deepEqual(args[0], 'commit');
+        assert.isFalse(args[1].partial);
+      });
+
+      it.only('reports if the commit was an amend', async function() {
```
annthurium: I was like "why did code coverage decrease so much" and then I was like "ohh." I do the same thing myself at least once a week.

lib/models/repository-states/present.js
kuychaco: Yeah, I had that thought but then a couple of considerations caused me to pause. Maybe you have thoughts that would help with un-pausing :). 

Do you have specific questions you'd like answered by this information or examples of conclusions we might be able to draw?

These were the considerations that caused me to pause initially:
* as mentioned in the description, tracking all of the staging/unstaging events seems like it could be a lot of noise with little additional value. For example, keyboard users might stage files by hitting `enter` a bunch of times for each line/hunk/file they want to stage... so for a given partial commit, there could be a ton of staging events. Maybe this could provide valuable information, but nothing came to mind for me immediately
* the distinction between partial staging as opposed to staging everything can't necessarily be inferred from instrumenting the UI controls. For example, clicking the "Stage selection" button could result in partial staging of a file, or it could result in staging all changes if the user has selected all of the changes before clicking the button. 

💭 welcome if you have any!

lib/models/repository-states/present.js
kuychaco: heh, yeah good question. it's not immediately clear.

This was to fix an oversight. I checked the format of the `results` returned from this function and objects are expected rather than arrays. Probably doesn't make a difference for the error case here since they're all empty objects anyways, but consistency is good

kuychaco:
Regarding gathering metrics on expanded commit editor usage, if we do choose to do so

> It looks like the only way to toggle the expanded commit editor is through that button, though? You could instrument the button to get a sense of how often users are toggling it on. 

You can also use a command, so the optimal location to instrument would be the `CommitController#openCommitMessageEditor` method, which both the button and command call through to

> If it's very occasional, we can safely assume this is not a popular feature.

Yeah, that and/or there’s potential to improve discoverability :)

lib/models/repository-states/present.js
```diff
@@ -233,17 +234,28 @@ export default class Present extends State {
...
-            }),
-          };
+      () => this.executePipelineAction('COMMIT', async (message, options) => {
+        const coAuthors = options && options.coAuthors;
+        const opts = !coAuthors ? options : {
+          ...options,
+          coAuthors: coAuthors.map(author => {
+            return {email: author.getEmail(), name: author.getFullName()};
+          }),
+        };
```
kuychaco: @smashwilson can you double check to see if I retained the logic you intended with this change?

test/models/repository.test.js
```diff
@@ -654,6 +766,40 @@ describe('Repository', function() {
...
+      const workingDirPath = await cloneRepository('multiple-commits');
+      const repo = new Repository(workingDirPath);
+      await repo.getLoadPromise();
+
+      sinon.stub(reporterProxy, 'addEvent');
+      sinon.stub(repo.git, 'reset').throws();
+
+      assert.deepEqual(reporterProxy.addEvent.callCount, 0);
+      try {
+        await repo.undoLastCommit();
```
smashwilson: If you want to be _sure_ this is a failure, you can replace this try-catch with:

```js
await assert.isRejected(repo.undoLastCommit());
```

test/controllers/commit-controller.test.js
```diff
@@ -297,6 +298,30 @@ describe('CommitController', function() {
...
       });
+
+      describe('openCommitMessageEditor', function() {
+        it('records an event', async function() {
+          const wrapper = shallow(app, {disableLifecycleMethods: true});
+
+          sinon.stub(reporterProxy, 'addEvent');
+          // open expanded commit message editor
+          await wrapper.find('CommitView').prop('toggleExpandedCommitMessageEditor')('message in box');
+          assert.deepEqual(reporterProxy.addEvent.callCount, 1);
```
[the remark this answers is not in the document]
kuychaco: oh haha yeah, copy/paste oversight. Thanks for pointing this out!

test/controllers/commit-controller.test.js
```diff
@@ -297,6 +298,30 @@ describe('CommitController', function() {
...
+          let args = reporterProxy.addEvent.lastCall.args;
+          assert.deepEqual(args[0], 'open-commit-message-editor');
+          assert.deepEqual(args[1], {package: 'github'});
```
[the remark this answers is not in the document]
kuychaco: oooh I like!

test/models/repository.test.js
kuychaco: I _do_ want to be sure! :) 

Thanks for all the great suggestions for cleaning up test assertions @smashwilson. I like the improved cleanliness and readability ✨

PR #1687, opened 2018-09-10
Temporarily disable keytar-reliant tests on AppVeyor

smashwilson:
All Atom beta and dev-channel builds are currently experiencing issues building native modules correctly. See atom/atom#17997 for the tracking issue. Because it doesn't look like a trivial thing, I'm disabling the failing tests here on AppVeyor until we can work out the root cause.


smashwilson:
@annthurium @kuychaco Okay, `git pull origin master` into your WIP branches and we won't be blocked by _this_ anyway :smile:

PR #1695, opened 2018-09-18
Add prompts for metrics, tests, documentation, and hypothesis to validate

kuychaco:

### Requirements

### Description of the Change

Add prompts to ensure that relevant metrics, tests, and documentation are added to pull requests, as well as hypotheses to test once changes are released.

### Alternate Designs

@sguthals, @smashwilson, @annthurium, and I discussed capturing these in a [project board](https://github.com/atom/github/projects/12#card-12742461) or in a [tracking issue](https://github.com/github/VisualStudio/issues/1899) and decided to give this method a try in order to minimize the number of places we need to look for information. 

### Benefits

These prompts will encourage contributors and reviewers to think more holistically about the code changes being introduced and help us keep track of what we need to validate after releasing. 

The metrics prompt will help us make data-driven decisions, better understand our users and their behaviors, and draw conclusions about the impact of the changes we've introduced. 

The test prompt will hopefully minimize bugs and help code reviewers and future contributors understand the intention of the code introduced.

The documentation prompt will surface useful information to users, maintainers, developers, etc

Documenting all of these in PR descriptions will give maintainers a quick way to see what was introduced in a given release cycle and what needs to be validated by simply looking at the info for each merged Pull Request.

### Possible Drawbacks

None that I can think of.

### Applicable Issues

See alternative approaches in this [project board card](https://github.com/atom/github/projects/12#card-12742461) and [tracking issue](https://github.com/github/VisualStudio/issues/1899)

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -3,8 +3,7 @@
 ### Requirements
 
 * Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
-* All new code requires tests to ensure against regressions.
-* Some sections may not be applicable to your changes, and a simple "N/A" will suffice.
+* Suggestion: You can use checklists to keep track of progress for the sections on metrics, tests, documentation, and user research.
 
 ### Description of the Change
 
@@ -36,7 +35,7 @@ We must be able to understand the design of your change from this description. I
 
 ### Tests
 
-<!-- What tests were added / will be added to help protect against future regressions? Write "N/A" if not applicable. -->
+<!-- What unit or integration tests were (or will be) added added to help protect against future regressions? For manual testing, be sure to describe in detail what steps you took. If you chose not to include a specific test, please explain why. Write "N/A" if not applicable. -->
 
 ### Documentation
 
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -9,6 +9,8 @@
 
 <!--
 
+Please start with a short (5-10 word) description and elaborate and provide context as you see fit.
+
 We must be able to understand the design of your change from this description. If we can't get a good idea of what the code will be doing from the description here, the pull request may be closed at the maintainers' discretion. Keep in mind that the maintainer reviewing this PR may not be familiar with or have worked with the code here recently, so please walk us through the concepts.
 
 -->
@@ -28,3 +30,21 @@ We must be able to understand the design of your change from this description. I
 ### Applicable Issues
 
 <!-- Enter any applicable Issues here -->
+
+### Metrics
+
+<!-- What metrics are associated with this code change and what questions can they help us answer? -->
+
+### Tests
+
+<!-- What tests were added / will be added to help protect against future regressions? -->
+
+### Documentation
+
+<!-- Describe the documentation added or improved -->
+
+### Validation
+
+<!-- What hypothesis need to be validated/invalidated once this change is released? -->
+
+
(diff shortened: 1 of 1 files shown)


PULL_REQUEST_TEMPLATE.md
```diff
@@ -28,3 +30,25 @@ We must be able to understand the design of your change from this description. I
...
+
+### Tests
+
+<!-- What tests were added / will be added to help protect against future regressions? -->
+
+### Documentation
+
+<!-- Describe the documentation added or improved -->
+
+### Pre-Release Validation
```
annthurium: - If I was a community contributor, I might be confused by this question.

- I don't think we have hypotheses to answer for every pull request.  And if we collect those hypotheses on pull requests, that doesn't make it easy to aggregate them.  Instead, I think it makes the most sense to list our hypotheses in a release checklist, since that's the cadence we're aiming for with our research efforts.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -28,3 +30,25 @@ We must be able to understand the design of your change from this description. I
 ### Applicable Issues
 
 <!-- Enter any applicable Issues here -->
+
+### Metrics
+
+<!-- What metrics are associated with this code change and what questions can they help us answer? -->
+
+### Tests
```
annthurium: yay, so glad this is a section!  Test plans are very important.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -28,3 +30,25 @@ We must be able to understand the design of your change from this description. I
 ### Applicable Issues
 
 <!-- Enter any applicable Issues here -->
+
+### Metrics
```
annthurium: 💯

kuychaco:
> Should we consider aligning our pull request template with atom/atom?

Good question! I checked out the new PR template on `atom/atom` and it looks pretty cool. I'm a bit hesitant to bring over all of the changes they made over there (especially because of the extra step of having to copy/paste templates over), but definitely open to discussing it. For now I've drawn inspiration from them and re-used some of the wording, namely in the Tests section. What do you think @annthurium? 💭

PULL_REQUEST_TEMPLATE.md
```diff
@@ -28,3 +28,40 @@ We must be able to understand the design of your change from this description. I
...
+
+-->
+
+### Documentation
+
+<!-- Describe the documentation added or improved. Write "N/A" if not applicable. -->
+
+### User Experience Research
+
+#### Hypotheses to validate prior to releasing
```
annthurium: I'm still not convinced that this is useful information to collect at the pull request level of granularity.  The pull request template already feels heavy/cumbersome to fill out, for me.  Probably even more so for new contributors. And are we doing anything with this information after we write it out?  We're not going to have enough time to research hypotheses for every pull request.  And I don't think pull requests are where our UXR findings should live because it's hard to see themes in aggregate. 

At the same time, I acknowledge that you said you would find value in writing out the answers to these questions on your own pull requests. How would you feel about clearly marking these questions as optional?  Optional is different from "N/A" in my opinion.  "N/A" is "there are situations where this might not apply" optional is "do this if you would find it valuable."  If you find value in answering these questions in writing, I don't want to stop you.  But I don't think they should be required.

kuychaco: Yeah I hear you. Definitely don't want to be increasing friction for contributors.

Definitely agree with making explicit that these are optional (you make a good point about "N/A" not conveying optionality). I'm gonna play with the wording some more and try scaling back this section quite a bit to see how it feels. 

I'm also open to dropping it from the template entirely. In terms of where best to capture this information and have these discussions, @daviwil brought up a good point and suggested that RFCs are a good candidate. That said, we don't always create an RFC for our work and not every change that would benefit from UXR would justify having an RFC. 

As we get into the swing of things with UXR work I'm sure other ideas for keeping track of UXR hypotheses will come up.

sguthals:
I really like this new template and the changes. Thanks team for working together on this and @kuychaco for taking the lead!

PR #1702, opened 2018-09-25
Move "Undo Discard" button behind context menu

### LINKED ISSUE #1260: Require confirmation on undo discard
### Description

The package should confirm the user if they really want to undo discarded changes. There's no spacing between that button and the first diff item so I accidentally press it often and have to manually search through the diff to see what it restored and undo that
### Versions

```
Atom    : 1.22.1
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0

apm  1.18.8
npm  3.10.10
node 6.9.5 x64
python 2.7.10
git 2.14.2
```
Using bundled v0.7.0 of this package

kuychaco:
### Description of the Change

This PR removes the `Undo Discard` button and hides this functionality behind a context menu that is accessed from the Unstaged Changes list **header**. 

git_ ___src_test-repo
fullscreen_9_24_18__11_43_pm

Note: A similar context menu currently exists in the Merge Conflicts list header for bulk resolve actions.

When multiple files are selected, the context menu item is pluralized to `Discard Changes in Selected Files`. I chose this more verbose description (rather than simply `Discard Changes`) to make it clear that this would discard changes for the Unstaged Changes file list, and not for the file diff view which has separate discard state.

The Stage/Unstage all buttons no longer disappear when there are no changes. Now they simply become disabled. This gives something for the new kebab menu to sit up against.

### Alternate Designs

Initially, I moved the `Undo Discard` button into the header and hid the text when the width was too narrow (see https://github.com/atom/github/commit/b82e732ca7e1d6e642fdb4f133447064647d51a5):

git_ ___src_test-repo
git_ ___src_test-repo

This solution still leaves room for the user to accidentally click the `Undo Discard` button. It's right next to the `Stage All` button and, depending on styling and theme, the boundary between the two buttons may not be super visible, making it easier for users to mistakenly hit the `Undo discard` button when trying to hit the `Stage All` button.

Hiding the undo-discard functionality behind a context menu on the header makes it harder to accidentally trigger this action. See images in section above.

### Benefits

This UX makes it harder to accidentally undo the last discard.

### Possible Drawbacks

The undo discard functionality might be less discoverable since it is hidden behind a menu. Users may be confused by the disappearance of a button from the UI. 

### Applicable Issues

Issue -- #1242
Supersedes PR #1252
Fixes #1260 

### Metrics

- [x] Instrument `undoLastDiscard` method
- [x] Instrument discard methods
- [x] Instrument UI entry points (buttons, commands, header menu)

### Tests

Manual testing steps:
* click on <kbd>...</kbd> menu
* verify that undo last discard is greyed out when there is no discard history
* verify that "Discard Changes in Selected File" is pluralized correctly
* verify functionality of all menu items (discard all, discard selected files, undo last discard)
* verify that Stage/Unstage All buttons remain in UI and are disabled when there are no changes
* verify that header context menus no longer show "Stage All" and "Discard All" 

No context menu unit tests added as there is currently no way to write UI tests for context menus in electron --
https://github.com/electron/spectron/issues/21

Added metrics tests for recording the following:
* discarding all changes from the StagingView
* discarding selected changes from the StagingView
* undoing the last discard from the StagingView
* discarding selected lines from the FilePatchController
* undoing the last discard from the FilePatchController

### Documentation

- [ ] Update flight manual - https://github.com/atom/flight-manual.atom.io/pull/488 (add issue to merge this in project board for correct release)

Some light UXR before shipping would be helpful. Especially to assess discoverability.
After shipping, we can monitor usage of the discard/undo discard functionality. It would be interesting to tweet about it and see if that affects feature usage. 

Questions:
- [x] How's the wording of the context menu items?
- [x] When there are no unstaged changes the `Stage All` button disappears and the ellipses icon moves to the right. This might look jarring. Maybe we should keep it fixed in the middle of the header, or place it to the right of the `Stage All` button. If we place it to the right then the `Stage All` button won't line up with the `Unstage All` button in the section below...
- [x] Would vertical ellipses look better than horizontal ellipses? Note that the merge conflict section uses the horizontal ellipses for its bulk resolve menu. Also, seems like the vertical ellipses icon is not available in the octicon set bundled with Atom.

/cc @simurai for 💭s on ☝️ 

TODO:
- [x] Conditionally show `Undo Last Discard` only if there is undo history
- [x] Show items in menus at all times, but make unclickable if no undo history 
- [x] Remove context menu items for header and put in kebab

Future steps:
- [ ] Consider prompting users to confirm undo discard if head has moved since last discard 
- [ ] Update flight manual when changes hit stable

### DIFF
--- a/lib/controllers/file-patch-controller.js
+++ b/lib/controllers/file-patch-controller.js
@@ -539,10 +539,11 @@ export default class FilePatchController extends React.Component {
     return this.props.discardLines(this.state.filePatch, lines, this.repositoryObserver.getActiveModel());
   }
 
-  undoLastDiscard() {
+  undoLastDiscard({eventSource} = {}) {
     addEvent('undo-last-discard', {
       package: 'github',
       component: 'FilePatchController',
+      eventSource,
     });
     return this.props.undoLastDiscard(this.props.filePath, this.repositoryObserver.getActiveModel());
   }
--- a/lib/controllers/file-patch-controller.js
+++ b/lib/controllers/file-patch-controller.js
@@ -530,11 +530,12 @@ export default class FilePatchController extends React.Component {
     return textEditor;
   }
 
-  discardLines(lines) {
+  discardLines(lines, {eventSource} = {}) {
     addEvent('discard-unstaged-changes', {
       package: 'github',
       component: 'FilePatchController',
       lineCount: lines.length,
+      eventSource,
     });
     return this.props.discardLines(this.state.filePatch, lines, this.repositoryObserver.getActiveModel());
   }
--- a/lib/controllers/file-patch-controller.js
+++ b/lib/controllers/file-patch-controller.js
@@ -8,6 +8,7 @@ import Switchboard from '../switchboard';
 import FilePatchView from '../views/file-patch-view';
 import ModelObserver from '../models/model-observer';
 import {autobind} from '../helpers';
+import {addEvent} from '../reporter-proxy';
 
 export default class FilePatchController extends React.Component {
   static propTypes = {
@@ -530,10 +531,19 @@ export default class FilePatchController extends React.Component {
   }
 
   discardLines(lines) {
+    addEvent('discard-unstaged-changes', {
+      package: 'github',
+      component: 'FilePatchController',
+      lineCount: lines.length,
+    });
     return this.props.discardLines(this.state.filePatch, lines, this.repositoryObserver.getActiveModel());
   }
 
   undoLastDiscard() {
+    addEvent('undo-last-discard', {
+      package: 'github',
+      component: 'FilePatchController',
+    });
     return this.props.undoLastDiscard(this.props.filePath, this.repositoryObserver.getActiveModel());
   }
 
--- a/lib/views/staging-view.js
+++ b/lib/views/staging-view.js
@@ -608,9 +608,12 @@ export default class StagingView extends React.Component {
 
     const menu = new Menu();
 
-    if (this.props.unstagedChanges.length) {
+    const selectedItemCount = this.getSelectedItemFilePaths().length;
+    const pluralization = selectedItemCount > 1 ? 's' : '';
+
+    if (this.props.unstagedChanges.length && selectedItemCount) {
       menu.append(new MenuItem({
-        label: 'Discard Selected Files',
+        label: 'Discard Changes in Selected File' + pluralization,
         click: () => this.discardChanges(),
       }));
     }
--- a/lib/views/staging-view.js
+++ b/lib/views/staging-view.js
@@ -15,6 +15,7 @@ import ResolutionProgress from '../models/conflicts/resolution-progress';
 import RefHolder from '../models/ref-holder';
 import FilePatchController from '../controllers/file-patch-controller';
 import Commands, {Command} from '../atom/commands';
+import SimpleTooltip from '../atom/simple-tooltip';
 import {autobind} from '../helpers';
 
 const debounce = (fn, wait) => {
@@ -57,6 +58,7 @@ export default class StagingView extends React.Component {
     commandRegistry: PropTypes.object.isRequired,
     notificationManager: PropTypes.object.isRequired,
     workspace: PropTypes.object.isRequired,
+    tooltips: PropTypes.object.isRequired,
     openFiles: PropTypes.func.isRequired,
     attemptFileStageOperation: PropTypes.func.isRequired,
     discardWorkDirChangesForPaths: PropTypes.func.isRequired,
@@ -171,6 +173,12 @@ export default class StagingView extends React.Component {
         element.scrollIntoViewIfNeeded();
       }
     }
+
+    if (this.header.clientWidth < 365) {
+      this.header.classList.add('github-StagingView-header--narrow-width');
+    } else {
+      this.header.classList.remove('github-StagingView-header--narrow-width');
+    }
   }
 
   render() {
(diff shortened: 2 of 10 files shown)


kuychaco:
> ellipsis icon 👉 dedicated menu for discarding, since "Stage All Changes" already has its dedicated button.

@simurai agreed 👍 

I ended up removing the "Stage All" and "Discard All" from the header context menu. Now that we have a dedicated UI element for exposing these actions, it seems unnecessary to have a right-click context menu for the same actions on the header.

lib/controllers/file-patch-controller.js
```diff
@@ -530,10 +531,19 @@ export default class FilePatchController extends React.Component {
...
+      lineCount: lines.length,
+    });
     return this.props.discardLines(this.state.filePatch, lines, this.repositoryObserver.getActiveModel());
   }
 
   undoLastDiscard() {
+    addEvent('undo-last-discard', {
+      package: 'github',
+      component: 'FilePatchController',
+    });
```
smashwilson: Yess metrics

lib/views/staging-view.js
```diff
@@ -377,20 +381,30 @@ export default class StagingView extends React.Component {
     this.subs.dispose();
   }
 
+  getSelectedItemFilePaths() {
```
smashwilson: 👍 for refactoring this out.

lib/views/staging-view.js
```diff
@@ -586,6 +606,35 @@ export default class StagingView extends React.Component {
     menu.popup(remote.getCurrentWindow());
   }
 
+  showActionsMenu(event) {
+    event.preventDefault();
+
+    const menu = new Menu();
+
+    const selectedItemCount = this.getSelectedItemFilePaths().length;
```
smashwilson: This could also be `this.state.selection.getSelectedItems().size` to save an Array construction, yeah?

(Micro-optimization, not important 😆 )

test/controllers/file-patch-controller.test.js
```diff
@@ -298,6 +299,31 @@ describe('FilePatchController', function() {
         assert.isTrue(cursorCall.args[0].isEqual([4, 0]));
       });
     });
+
+    describe('discardLines()', function() {
+      it('records an event', function() {
+        const wrapper = mount(component);
+        sinon.stub(reporterProxy, 'addEvent');
+        wrapper.instance().discardLines(['once upon a time', 'there was an octocat named mona lisa']);
```
smashwilson: :octocat:

lib/views/staging-view.js
kuychaco: Indeed. I'm all for optimization 👍

lib/views/staging-view.js
```diff
@@ -377,20 +381,30 @@ export default class StagingView extends React.Component {
...
   openFile() {
-    const filePaths = Array.from(this.state.selection.getSelectedItems(), item => item.filePath);
+    const filePaths = this.getSelectedItemFilePaths();
     return this.props.openFiles(filePaths);
   }
 
   discardChanges() {
-    const filePaths = Array.from(this.state.selection.getSelectedItems(), item => item.filePath);
+    const filePaths = this.getSelectedItemFilePaths();
+    addEvent('discard-unstaged-changes', {
```
annthurium: it looks like we're measuring similar events from the `StagingView` and the `FilePatchController`.  Since the `FilePatchController` is above the `StagingView` in the component tree, I'm assuming both these events will be fired every time a discard happens.  Which makes me worry about double counting, if you're trying to figure out the total usage of the "undo discard" feature.

What's really useful to learn from instrumenting the view layer is how the user is interacting with the UI.  Are they actually clicking the new button?  Instrumenting that button more specifically makes double counting less likely and clarifies what you're looking to learn.

lib/views/staging-view.js
```diff
@@ -586,6 +606,35 @@ export default class StagingView extends React.Component {
     menu.popup(remote.getCurrentWindow());
   }
 
+  showActionsMenu(event) {
+    event.preventDefault();
+
+    const menu = new Menu();
+
+    const selectedItemCount = this.state.selection.getSelectedItems().size;
+    const pluralization = selectedItemCount > 1 ? 's' : '';
```
annthurium: if we ever i18n our package, all our hard coded pluralization is going to be fun to convert....

lib/views/staging-view.js
```diff
@@ -797,6 +846,11 @@ export default class StagingView extends React.Component {
       return;
     }
 
+    addEvent('undo-last-discard', {
```
annthurium: same comment as above - would recommend refactoring this so that you're logging menu views as a ui interaction.

lib/views/staging-view.js
kuychaco: > Since the FilePatchController is above the StagingView in the component tree, I'm assuming both these events will be fired every time a discard happens.

They're actually in different parts of the component hierarchy and instrument separate discard/undo discard states. There should be no double counting.

> What's really useful to learn from instrumenting the view layer is how the user is interacting with the UI.

I totally agree with this. The tricky thing is the we can't easily instrument all UI actions to distinguish between different entry points. A button, yes, that's easy. Menu usage versus keybinding versus command palette usage, not so easy. I can't think of an easy straightforward way to instrument those, they all go through the command. 

In order to catch all discard and undo discard actions I had to instrument the `undoLastDiscard` method. I had the thought to also instrument the button but that would lead to duplicate events. I guess I could pass some meta data along to record if the event was triggered by a button versus command. But we'd have to keep in mind that the command could be triggered by a menu or keybinding as well as the command palette. Okay I'll do that... Maybe add an `eventSource` property.

lib/views/staging-view.js
kuychaco: Yuuuup 😕

kuychaco:
@annthurium what do you think of my addition of metadata for event collection in https://github.com/atom/github/pull/1702/commits/32b0ada679c58579773ad6a74c2597842a938710. I still have to do something similar for discarding changes.

kuychaco:
> ah, I like eventSource! That does solve the problem elegantly. Thanks for addressing my review feedback!

@annthurium Totally! Thanks for pushing me to think past my initial block on this :)

PR #1704, opened 2018-09-25
discoverability improvements

annthurium:
### Description of the Change

There are a lot more Atom users than there are users of the Git/Github package.  This points to a discoverability problem.  There are menus and commands to open the Git/GitHub panes, but in order to do that you have to know the panes exist. 

This pull request changes the icon and text in the status bar to say "Git (n)" where n is the number of changed files, and the icon is a diff icon.  This pull request also adds a GitHub icon and text to the status bar, which toggles the GitHub tab when clicked.

screen shot 2018-09-25 at 3 08 04 pm

### Alternate Designs

I thought about not showing the GitHub status bar icon if there are no GitHub.com remotes present.  The GitHub tab isn't very useful in that case.  It's not really doing any harm to show the GitHub tab -- it's just not a great use of real estate.  Thus, I decided the complexity of hiding the icon wasn't worth the hassle.

### Benefits

Hopefully, more users will learn that the git and github panes exist, use them, and find the meaning that's been missing from their lives.

We plan on doing usability testing to validate the assumption that the UI improves discoverability.

### Possible Drawbacks

If we keep adding stuff to the status bar, eventually it's going to get too crowded.  For right now, it's fine.

### Applicable Issues

https://github.com/atom/github/issues/1573

### Metrics

When a user clicks on the GitHub status bar icon, we record an event.

### Tests

Unit tests to validate that...
 - `ChangedFilesCountView`:
  - renders the new icon
  - renders the text "Git"
- `GithubTileView`:
  - renders the new icon
  - renders the text "GitHub"
  - records an event when clicked
  - calls the passed in prop when clicked
- `StatusBarTileController`:
  - toggles the GitHub tab when clicked

Manual testing to validate that...
 - the git tab shows / hides when the status bar icon is clicked
 - the github tab shows / hides when the status bar icon is clicked
 - clicking the github tab doesn't cause any errors if no github.com remotes are present.

### DIFF
--- a/lib/controllers/status-bar-tile-controller.js
+++ b/lib/controllers/status-bar-tile-controller.js
@@ -5,7 +5,7 @@ import BranchView from '../views/branch-view';
 import BranchMenuView from '../views/branch-menu-view';
 import PushPullView from '../views/push-pull-view';
 import ChangedFilesCountView from '../views/changed-files-count-view';
-import GithubStatusBarTile from '../views/github-status-bar-tile';
+import GithubTileView from '../views/github-tile-view';
 import Tooltip from '../atom/tooltip';
 import Commands, {Command} from '../atom/commands';
 import ObserveModel from '../views/observe-model';
@@ -94,7 +94,7 @@ export default class StatusBarTileController extends React.Component {
     return (
       <Fragment>
         {this.renderTiles(repoProps)}
-        <GithubStatusBarTile didClick={this.props.toggleGithubTab} />
+        <GithubTileView didClick={this.props.toggleGithubTab} />
         <ChangedFilesCountView
           didClick={this.props.toggleGitTab}
           changedFilesCount={repoProps.changedFilesCount}
--- a/lib/controllers/status-bar-tile-controller.js
+++ b/lib/controllers/status-bar-tile-controller.js
@@ -5,6 +5,7 @@ import BranchView from '../views/branch-view';
 import BranchMenuView from '../views/branch-menu-view';
 import PushPullView from '../views/push-pull-view';
 import ChangedFilesCountView from '../views/changed-files-count-view';
+import GithubStatusBarTile from '../views/github-status-bar-tile';
 import Tooltip from '../atom/tooltip';
 import Commands, {Command} from '../atom/commands';
 import ObserveModel from '../views/observe-model';
@@ -20,8 +21,13 @@ export default class StatusBarTileController extends React.Component {
     confirm: PropTypes.func.isRequired,
     repository: PropTypes.object.isRequired,
     toggleGitTab: PropTypes.func,
+    toggleGithubTab: PropTypes.func,
     ensureGitTabVisible: PropTypes.func,
   }
+  // todo: ensureGitTabVisible is unused.
+  // it looks like we used to pop open the git tab if merge conflicts
+  // were present.  Is that something we want to keep doing?
+  // if so we should fix it.
 
   constructor(props) {
     super(props);
@@ -88,6 +94,7 @@ export default class StatusBarTileController extends React.Component {
     return (
       <Fragment>
         {this.renderTiles(repoProps)}
+        <GithubStatusBarTile didClick={this.props.toggleGithubTab} />
         <ChangedFilesCountView
           didClick={this.props.toggleGitTab}
           changedFilesCount={repoProps.changedFilesCount}
--- a/lib/views/changed-files-count-view.js
+++ b/lib/views/changed-files-count-view.js
@@ -28,17 +28,13 @@ export default class ChangedFilesCountView extends React.Component {
   }
 
   render() {
-    const label =
-      (this.props.changedFilesCount === 1)
-        ? '1 file'
-        : `${this.props.changedFilesCount} files`;
     return (
       <button
         ref="changedFiles"
         className="github-ChangedFilesCount inline-block"
         onClick={this.handleClick}>
-        <Octicon icon="diff" />
-        {label}
+        <Octicon icon="git-commit" />
+        {`Git (${this.props.changedFilesCount})`}
         {this.props.mergeConflictsPresent && <Octicon icon="alert" />}
       </button>
     );
--- a/lib/views/github-status-bar-tile.js
+++ b/lib/views/github-status-bar-tile.js
@@ -1,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Octicon from '../atom/octicon';
+
+import {addEvent} from '../reporter-proxy';
+import {autobind} from '../helpers';
+
+export default class GithubStatusBarTile extends React.Component {
+  static propTypes = {
+    didClick: PropTypes.func.isRequired,
+  }
+
+  constructor(props) {
+    super(props);
+    autobind(this, 'handleClick');
+  }
+
+  handleClick() {
+    addEvent('click', {package: 'github', component: 'GithubStatusBarTile'});
+    this.props.didClick();
+  }
+
+  render() {
+    return (
+      <button
+        className="github-StatusBarTile inline-block"
+        onClick={this.handleClick}>
+        <Octicon icon="mark-github" />
+        GitHub
+      </button>
+    );
+  }
+}
--- a/lib/controllers/root-controller.js
+++ b/lib/controllers/root-controller.js
@@ -161,6 +161,7 @@ export default class RootController extends React.Component {
           tooltips={this.props.tooltips}
           confirm={this.props.confirm}
           toggleGitTab={this.gitTabTracker.toggle}
+          toggleGithubTab={this.githubTabTracker.toggle}
           ensureGitTabVisible={this.gitTabTracker.ensureVisible}
         />
       </StatusBar>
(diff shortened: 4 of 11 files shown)


lib/controllers/status-bar-tile-controller.js
```diff
@@ -20,8 +21,13 @@ export default class StatusBarTileController extends React.Component {
     confirm: PropTypes.func.isRequired,
     repository: PropTypes.object.isRequired,
     toggleGitTab: PropTypes.func,
+    toggleGithubTab: PropTypes.func,
     ensureGitTabVisible: PropTypes.func,
   }
+  // todo: ensureGitTabVisible is unused.
```
annthurium: @kuychaco : do you have thoughts on this?

kuychaco: Hmmm looks like we're never actually calling this method. So maybe we can just drop it?

kuychaco: Oh oops, didn't see the comments below as well. Let's ask @smashwilson about the desired merge conflict behavior. Maybe this is a regression?

lib/controllers/status-bar-tile-controller.js
```diff
@@ -5,6 +5,7 @@ import BranchView from '../views/branch-view';
 import BranchMenuView from '../views/branch-menu-view';
 import PushPullView from '../views/push-pull-view';
 import ChangedFilesCountView from '../views/changed-files-count-view';
+import GithubStatusBarTile from '../views/github-status-bar-tile';
```
kuychaco: Here's a suggestion for the class name based on package convention (ending with the word "view") and context (being in the `status-bar-tile-controller.js` file) -- `GitHubTileView`. And the file could be `github-tile-view.js`.

Similarly, since you changed the tile for opening the Git panel it might be more appropriately named something like `GitTileView` and the file `git-tile-view.js`. 

Corresponding test files would need to be changed accordingly.

lib/views/changed-files-count-view.js
```diff
@@ -28,17 +28,13 @@ export default class ChangedFilesCountView extends React.Component {
...
-        : `${this.props.changedFilesCount} files`;
     return (
       <button
         ref="changedFiles"
         className="github-ChangedFilesCount inline-block"
         onClick={this.handleClick}>
-        <Octicon icon="diff" />
-        {label}
+        <Octicon icon="git-commit" />
+        {`Git (${this.props.changedFilesCount})`}
```
kuychaco: > This pull request changes the icon and text in the status bar to say "Git (n)" where n is the number of changed files

Interestingly, when I first saw this my initial thought was that n was the number of git commits... I notice currently we have the number as well as the word "Files" which makes it very clear that the number refers to the changed file count. I wonder if there's a way we could make it more clear with your new version... adding the word "Files" might take up too much space. I think it would even be fine to just drop the file count...

@simurai do you have 💭s?

lib/views/github-status-bar-tile.js
```diff
@@ -0,0 +1,33 @@
...
+    didClick: PropTypes.func.isRequired,
+  }
+
+  constructor(props) {
+    super(props);
+    autobind(this, 'handleClick');
+  }
+
+  handleClick() {
+    addEvent('click', {package: 'github', component: 'GithubStatusBarTile'});
```
kuychaco: Not sure what convention we're going with for event names (or if we have even discussed one yet)... I had the thought that a more descriptive event name could be handy, something like `toggle-github-tab`. That way whoever is combing through the data we're collecting can more easily infer what the events are for in terms of extracting information about user behavior.

kuychaco: That said, I can also see the value in simply stating that we're capturing a click event on a particular view element... 

Oh okay, so we want to decouple the UI action (click on an element) from the effect of the action (toggling the tab). We've instrumented the `reveal` and `hide` methods of the `TabTracker` which tells us the number of times a given tab is opened/closed. We also want to know what method was used to reveal/hide the tab, which is what we're doing here. Does that sounds about right?

lib/views/github-status-bar-tile.js
```diff
@@ -0,0 +1,33 @@
...
+    this.props.didClick();
+  }
+
+  render() {
+    return (
+      <button
+        className="github-StatusBarTile inline-block"
+        onClick={this.handleClick}>
+        <Octicon icon="mark-github" />
```
kuychaco: :octocat: ✨

lib/controllers/status-bar-tile-controller.js
```diff
@@ -88,6 +94,7 @@ export default class StatusBarTileController extends React.Component {
     return (
       <Fragment>
         {this.renderTiles(repoProps)}
+        <GithubStatusBarTile didClick={this.props.toggleGithubTab} />
```
simurai: How about moving the GitHub tile to the far right?

screen shot 2018-09-26 at 1 55 36 pm

Then it's more in order of this typical 🤔 workflow:
1. Choose a branch
2. Make sure it's up-to-date
3. Make changes
4. Create PR

lib/views/changed-files-count-view.js
simurai: > I think it would even be fine to just drop the file count...

That would be the least confusing, but we'd miss out on informing the user that there are uncommitted changes, which might be important information to keep.

> adding the word "Files" might take up too much space.

screen shot 2018-09-26 at 2 21 53 pm

Hmm.. yeah.

Or how do you like some sort of "notification badge":

image

Atom IDE uses something similar. It could also be just grey to be a bit more subtle.

lib/controllers/status-bar-tile-controller.js
annthurium: good call!  I'll change the name.

annthurium:
@simurai @kuychaco I'm going to merge this, as I believe I have addressed all feedback.  I am super open to addressing additional concerns in the future, if you have any.

lib/views/changed-files-count-view.js
kuychaco: Sounds good to me! Would we do this UXR prior to merging? Or just prior to releasing?

lib/controllers/status-bar-tile-controller.js
simurai: @annthurium That's a good point, although the "far right" sometimes gets used by the "updates" and "blue squirrel".

Another option would be to move "GitHub" to the left, before the branch name:

image

Just so that "GitHub" doesn't get sandwiched between the Git "actions" and the Git tab.

lib/views/changed-files-count-view.js
```diff
@@ -28,17 +28,13 @@ export default class ChangedFilesCountView extends React.Component {
...
-        ? '1 file'
-        : `${this.props.changedFilesCount} files`;
     return (
       <button
         ref="changedFiles"
         className="github-ChangedFilesCount inline-block"
         onClick={this.handleClick}>
-        <Octicon icon="diff" />
-        {label}
+        <Octicon icon="git-commit" />
```
bmario: I think changing the icon from `diff` to `git-commit` isn't a good choice. The pane is about changed files, it tells you how many files have local changes, yet it uses the symbol for commits. When I first saw the new layout, I'd expected to get the equivalent of `git log` and were confused about what the number is going to tell me.

simurai: The Git panel includes different things:
- Changes
- Commit message editor
- Recent commits

The "Recent commits" is some sort of mini `git log`:

commits

So it doesn't seem easy to find a good icon that fits everything. Using the git icon should hint at: "Create and view commits".

bmario: I'd agree, that for the pane the git commit icon is a better fit, but I'm talking about the status bar button.

The issue is, you're optimizing this for people who aren't even using it. And in doing so, you are breaking the current UI for the users you already have. You shouldn't take that lightly. The new layout doesn't convey more information, nor does it free up precious space in the status bar. Once you found out that the Git(Hub) pane exists and that the buttons will open those panes, these buttons take up space because they got bulky. 

I agree that the "n files" label wasn't perfect. What I'd suggest is merely splitting up the buttons. Make one button with just the diff icon and the number of changed files as the label. And a second button with the git-commit icon and git as its label. Both buttons open the diff/commit git pane on click, but the information is logically split. Plus, users who already discovered the git pane can hide the git button without losing functionality. Additionally, all buttons should have a tooltip like every other thing in the status bar does.
I'd even make a PR for that if you like that idea.

simurai: > nor does it free up precious space in the status bar.
> What I'd suggest is merely splitting up the buttons. Make one button with just the diff icon and the number of changed files as the label. And a second button with the git-commit icon and git as its label.

Here a mockup:

after

But wouldn't two buttons take up even more space than just a single button?

before

Also since the two buttons do the **same** thing, could that be confusing? People would click on both and then think something must be broken since in both cases it opens the same panel. Maybe this could be made less confusing if the focus moves to the changes and the recent commits. Like this:

screen shot 2018-12-10 at 12 13 23 pm

But having the focus in two different places seems less useful for users that use the mouse to open the git panel.

Issue #1697, opened 2018-09-18
Cannot handle local branch being named differently from its remote branch


stuartchaney:
### Description
1. Team Member creates a PR on a branch named feature/replace-iphone-image-template

2. When checking out this PR on Atom. A branch is created named 
pr-81/stuartchaney/feature/replace-iphone-image-template

The branch created on my local system should be feature/replace-iphone-image-template

When clicking fetch from Atom this will lead to the following

![](https://cl.ly/b8683baa617f/Image%202018-09-18%20at%2010.12.05%20AM.png)

**Expected behavior:**

local branch created from checking out PR should be feature/replace-iphone-image-template

**Actual behavior:**

local branch created from checking out PR is pr-81/stuartchaney/feature/replace-iphone-image-template

**Reproduces how often:**
100%

### Versions

1.30.0

vanessayuenn:
The underlying issue is that we're not properly handling the case where the local branch is named differently from its upstream. Related issue: #918.

Issue #918, opened 2017-06-12
Pulling the tracking git branch via UI causes error


vvs:
Doing git pull via the new Atom UI (pressing the "down arrow" button in the bottom-right corner, and then pressing "Pull" button, or "Fetch" button) causes the following error:

Unable to pull
fatal: Couldn't find remote ref my-local-branch-name

Doing the 'git pull' from the command line works fine.

Please note that there is no 'my-local-branch-name' on the server indeed, the remote branch name is different, but the local branch is set to track that remote branch, let's call it "my-remote-branch".

**Reproduces how often:**

100%

### Versions
Atom : 1.18.0-beta3
Electron: 1.3.15
Chrome : 52.0.2743.82
Node : 6.5.0

Let me know if you need more info (and how to get that info).

vvs:
Some more info. Looking at the logs, I see that atom tries to perform the following command:

`git  pull origin my-local-branch-name`

Which is obviously wrong, since my-local-branch-name is a local branch.

BinaryMuse:
Very good point — seems as though we look up the remote for the current branch, but we don't look up the upstream tracking branch as configured with the `branch.branchname.merge` (and in fact you can specify multiple values in order to have git perform an octopus merge when you pull).

/cc @kuychaco for 💭 s

PR #1708, opened 2018-09-27
Fix: handle branch names mismatch

vanessayuenn:
### Description of the Change

This PR introduces the use of `refspec` in push and pull operations (fetch does it a bit differently), in order to handle situations where the local branch is named differently from its tracking branch. This set of circumstances has become less of an edge case since the introduction of PR-checkout -- we prefix the branch name with PR number, etc. 

### Benefits

- Able to fetch, pull, push checked out PRs
- Able to be creative in naming your local branches 👯 

### Possible Drawbacks

This PR changes how the underlying git operations work, which is used by many things in the package; so the possible setback is if things break in edge cases that I haven't thought of.

### Applicable Issues

fixes: #1697 
fixes: #918 

### Metrics

N/A

### Tests

Added unit tests to verify that push & pull & fetch work as expected despite the difference between local branch name and its upstream branch name. 

Manually tested push & pull & fetch operations in these cases:
- local branch named differently from upstream
- local branch named the same as upstream
- checked out PR from a branch in the original repo
- checked out PR opened from a forked repo

### DIFF
--- a/lib/models/branch.js
+++ b/lib/models/branch.js
@@ -67,6 +67,23 @@ export default class Branch {
     return this.getRemoteRef().replace(/^(refs\/)?((heads|remotes)\/)?/, '');
   }
 
+  getRefSpec(action) {
+    debugger
+    if (this.isRemoteTracking()) {
+      return '';
+    }
+    const remoteBranch = action === 'PUSH' ? this.push : this.upstream;
+    const remoteBranchName = remoteBranch.getShortRemoteRef();
+    const localBranchName = this.getName();
+    if (remoteBranchName !== localBranchName) {
+      const refSpec = action === 'PUSH'
+        ? `${localBranchName}:${remoteBranchName}`
+        : `${remoteBranchName}:${localBranchName}`;
+      return refSpec;
+    }
+    return localBranchName;
+  }
+
   getSha() {
     return this.attributes.sha || '';
   }
@@ -94,6 +111,7 @@ export default class Branch {
   isPresent() {
     return true;
   }
+
 }
 
 export const nullBranch = {
--- a/lib/models/branch.js
+++ b/lib/models/branch.js
@@ -75,10 +75,11 @@ export default class Branch {
     const remoteBranchName = remoteBranch.getShortRemoteRef();
     const localBranchName = this.getName();
     if (remoteBranchName && remoteBranchName !== localBranchName) {
-      const refSpec = action === 'PUSH'
-        ? `${localBranchName}:${remoteBranchName}`
-        : `${remoteBranchName}:${localBranchName}`;
-      return refSpec;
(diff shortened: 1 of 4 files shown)


lib/models/branch.js
```diff
@@ -67,6 +67,23 @@ export default class Branch {
...
+    }
+    const remoteBranch = action === 'PUSH' ? this.push : this.upstream;
+    const remoteBranchName = remoteBranch.getShortRemoteRef();
+    const localBranchName = this.getName();
+    if (remoteBranchName && remoteBranchName !== localBranchName) {
+      if (action === 'PUSH') {
+        return `${localBranchName}:${remoteBranchName}`;
+      } else if (action === 'PULL') {
+        return `${remoteBranchName}:${localBranchName}`;
+      }
```
smashwilson: Using strings for enumerated values always makes me a bit nervous - it feels like it's easy to typo them somewhere (or miss a callsite when refactoring, say) and get surprising behavior. Sometimes I'll use Symbols instead, or objects with a `.when()` method, to push the "getting a valid action" work down into the language more.

With that said: the real answer here is "port our package to TypeScript" anyway. So this is just fine for now :smile:

test/controllers/status-bar-tile-controller.test.js
```diff
@@ -646,6 +646,50 @@ describe('StatusBarTileController', function() {
...
+
+    it('pulls from the correct branch', async function() {
+      const prePullSHA = await repository.git.exec(['rev-parse', 'HEAD']);
+      await repository.git.exec(['reset', '--hard', 'HEAD~2']);
+      sinon.spy(repository, 'pull');
+      await wrapper
+        .instance()
+        .pull(await wrapper.instance().fetchData(repository))();
+      const postPullSHA = await repository.git.exec(['rev-parse', 'HEAD']);
+      assert.isTrue(repository.pull.called);
```
annthurium: nit: do you want to assert that `repository.pull` was called with the args you expect?

test/controllers/status-bar-tile-controller.test.js
```diff
@@ -646,6 +646,50 @@ describe('StatusBarTileController', function() {
...
+
+    it('pushes to the correct branch', async function() {
+      await repository.git.commit('new local commit', {allowEmpty: true});
+      const localSHA = await repository.git.exec(['rev-parse', 'another-name']);
+      sinon.spy(repository, 'push');
+      await wrapper
+        .instance()
+        .push(await wrapper.instance().fetchData(repository))();
+      const remoteSHA = await repository.git.exec(['rev-parse', 'origin/master']);
+      assert.isTrue(repository.push.called);
```
annthurium: same nit as above - might be good to specify which args `push` should be called with.

PR #1734, opened 2018-10-08
Test to catch electron-link and mksnapshot failures

smashwilson:
I'm adding a unit test that uses electron-link and electron-mksnapshot to transpile the package source, preprocess it, then snapshot and verify it in a way that's consistent with [the way Atom does](https://github.com/atom/atom/blob/master/script/lib/generate-startup-snapshot.js). This should help us catch snapshotting problems before we tag and publish a release.

It's a bit of a heavy test - it takes about 20s for me on a cold cache. When you're running it locally, the transpilation and electron-link preprocessing are cached from run to run, which helps quite a bit  (down to around 7s).

We've hit a fair amount of pain with only discovering snapshot problems when we try to release, so I think it's worth including despite the slowdown. There _are_ alternatives, though: I could extract the snapshot generation into an independent script and add it as an `npm run` script, so we could choose explicitly when to run it. 

@atom/github-package: What do you think?


smashwilson:
Oh yuck, AppVeyor is hitting the same problems building native dependencies.

PR #1744, opened 2018-10-12
Add some instrumentation

### LINKED ISSUE #1727: Add metrics for cloning
Right now cloning is only accessible from the command palette. I wouldn't be surprised if we had some discoverability issues here. Let's collect some data on usage, and come up with ways to improve discoverability.

### LINKED ISSUE #1728: figure out a way of instrumenting context menu actions
Right now, we don't have a great way of collecting UI metrics for context menu actions.  Understand context menu actions is critical for discoverability, so we can understand how users are using the package.

I don't even have good suggestions as to what this might look like - more investigation is needed.

vanessayuenn:
### Description of the Change

This PR adds reporting metrics for cloning repos, as well as various context menu actions within the package. 

### Alternate Designs

This implementation listens to all command dispatch event, filters out the ones triggered via context menus, and acts upon them. The drawback is that it won't be clear _exactly_ which context menu was used to trigger that command. An alternative approach would be to hook into each event we're interested in individually, and reports back if the event came from a context menu, and which one. But this approach would result in the same code scattered all over the place, which isn't ideal either. 

### Metrics

We can know how much usage context menus get, how discoverable the `clone repo` functionality is. 

### Tests

I only added tests for the clone repo instrumentation, but not the context menu ones as I can't think of how one can test context menus in our situation. 

### Related Issues

resolves #1727 
resolves #1728

### DIFF
--- a/lib/controllers/root-controller.js
+++ b/lib/controllers/root-controller.js
@@ -472,6 +472,7 @@ export default class RootController extends React.Component {
         {detail: e.stdErr, dismissable: true},
       );
     } finally {
+      addEvent('clone-repo', {package: 'github'});
       this.setState({cloneDialogInProgress: false, cloneDialogActive: false});
     }
   }
--- a/lib/controllers/root-controller.js
+++ b/lib/controllers/root-controller.js
@@ -94,6 +94,16 @@ export default class RootController extends React.Component {
     this.subscription = new CompositeDisposable(
       this.props.repository.onMergeError(this.gitTabTracker.ensureVisible),
     );
+
+    this.props.commandRegistry.onDidDispatch(event => {
+      if (event.type && event.type.startsWith('github:')
+      && event.detail && event.detail[0] && event.detail[0].contextCommand) {
+        addEvent('context-menu-action', {
+          package: 'github',
+          command: event.type,
+        });
+      }
+    });
   }
 
   componentDidMount() {
--- a/lib/controllers/root-controller.js
+++ b/lib/controllers/root-controller.js
@@ -476,13 +476,13 @@ export default class RootController extends React.Component {
     this.setState({cloneDialogInProgress: true});
     try {
       await this.props.cloneRepositoryForProjectPath(remoteUrl, projectPath);
+      addEvent('clone-repo', {package: 'github'});
     } catch (e) {
       this.props.notificationManager.addError(
         `Unable to clone ${remoteUrl}`,
         {detail: e.stdErr, dismissable: true},
       );
     } finally {
-      addEvent('clone-repo', {package: 'github'});
       this.setState({cloneDialogInProgress: false, cloneDialogActive: false});
     }
   }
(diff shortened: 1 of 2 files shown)


lib/controllers/root-controller.js
```diff
@@ -94,6 +94,16 @@ export default class RootController extends React.Component {
     this.subscription = new CompositeDisposable(
       this.props.repository.onMergeError(this.gitTabTracker.ensureVisible),
     );
+
+    this.props.commandRegistry.onDidDispatch(event => {
```
annthurium: thanks for taking this on, @vanessayuenn !

It makes me nervous that there are no unit tests for this code, although I agree with you that this is tricky to test.

What if we stubbed `this.props.commandRegistry.onWillDispatch` and added `event.detail[0].contextCommand`?  Would this allow us to simulate a context event from a unit test?  Of course, it would have to . be scoped to just this unit test to prevent leakage.

Or maybe @smashwilson and @kuychaco have ideas here?

kuychaco: @vanessayuenn well done digging to find this `contextCommand` flag! Super handy! ✨ 

As far as how to unit test this, looks like there's an [undocumented] `detail` argument we can supply to the `dispatch` method on `commandRegistry` -- https://github.com/atom/atom/blob/v1.31.2/src/command-registry.js#L270

So it'd look something like this:

```
atom.commands.dispatch(
  '.github-FilePatchListView-path', 
  'github:discard-changes-in-selected-files', 
  {contextCommand: true}
)
```

We can test that we don't record events if `contextCommand` is not true, as well as if the event type does not start with `github:`.

One thing to note -- some of our context menu actions are bound to core commands (like `core:confirm`) so we wouldn't be catching these context menu actions. For example:

```
'.github-FilePatchView.is-staged': [
    {
      'label': 'Unstage Selection'
      'command': 'core:confirm'
    }
  ]
```

This is fine. Just something to keep in mind when we're looking at metrics, as well as when we're deciding what commands to bind context menu actions to.

test/controllers/root-controller.test.js
```diff
@@ -351,6 +352,7 @@ describe('RootController', function() {
       resolveClone();
 
       assert.isTrue(cloneRepositoryForProjectPath.calledWith('git@github.com:atom/github.git', '/home/me/github'));
+      assert.isTrue(reporterProxy.addEvent.calledWith('clone-repo', {package: 'github'}))
```
kuychaco: Looks like `addEvent` gets called asynchronously after the `cloneRepositoryForProjectPath` promise is resolved. We can account for this using our async assertion (courtesy of @BinaryMuse ✨):

```
      await assert.async.isTrue(reporterProxy.addEvent.calledWith('clone-repo', {package: 'github'}));
```

lib/controllers/root-controller.js
vanessayuenn: Thanks for the input & ideas!

PR #1746, opened 2018-10-15
Process proposal

smashwilson:
This is me writing down what I heard from our process discussion during last week's minisummit.

I still need to write up @annthurium's idea about shipping unreferenced components, and I'd like to hear @vanessayuenn's thoughts about this when she's around too before we ship :smile:


docs/core-team-process.md
```diff
@@ -0,0 +1,77 @@
+# Core @atom/github team process
+
+This guide describes the way that the core @atom/github team works together day-to-day.
+
+We value:
+
+* **Trust** in each other's judgement and instincts.
```
annthurium: 💯

docs/core-team-process.md
```diff
@@ -0,0 +1,77 @@
...
+
+:+1: _Advantages:_
+
+* Makes it less likely that one developer may block the others when their tasks take longer than expected.
+* More asynchronous-friendly.
+* Leaves a trail of documentation for each task.
+
+:-1: _Disadvantages:_
+
+* Decomposing tasks well is challenging.
```
annthurium: another disadvantage is having to coordinate when you're merging things.  If `user-story-c` finishes before `user-story-b`, and then you want to build a `user-story-d` that depends on `user-story-c`, that's going to be somewhat annoying.  (Maybe this is covered by "decomposing tasks well is challenging"?  ¯\_(ツ)_/¯

docs/core-team-process.md
```diff
@@ -0,0 +1,77 @@
...
+* Resilient to time off and asynchronicity.
+* Minimizes the need to context switch up and down abstraction layers while working.
+
+:-1: _Disadvantages:_
+
+* Diminishes variety of work done by any individual developer, which could become boring.
+* Reduces the familiarity developed by any single developer to a single abstraction layer within the codebase.
+* Timing may become difficult. It's possible that one "seam" may take much more time to implement than the others, which could lead to a bottleneck.
+* Some efforts will not be decomposable into easily identified seams for division of labor.
+
```
vanessayuenn: 🤔 This approach relies heavily on this to be true:
>Each developer continuously negotiates the interface with neighboring layers by an active Slack conversation, correcting their direction based on feedback.

-- with `continuous negotiation` being the key point, which makes me think this is not very timezone-friendly. 

I am imagining a scenario where developers A, B, and C agree on a certain direction to finish a task. Developer A (6 hours ahead of the others) goes ahead with the agreed upon direction for the model layer. Developer B starts their day 6 hours later, upon making changes to the view component, realizes the original appraoch wouldn't work because of Reasons™️; they let developer A knows, and the team renegotiates on the direction, but that may result in having to throw out the work developer A has already done.

docs/core-team-process.md
```diff
@@ -0,0 +1,77 @@
+# Core @atom/github team process
+
+This guide describes the way that the core @atom/github team works together day-to-day.
+
+We value:
```
vanessayuenn: Love that the values are put as the first and foremost thing. 🥇

docs/core-team-process.md
```diff
@@ -0,0 +1,77 @@
...
+
+* Overlap times need to be negotiated, either by pair programming or using another method to divvy up work. If we all overlap significantly it functionally decays to one of the other solutions.
+* Hand-offs are high communication touchpoints, but the rest of the time is more isolated.
+
+## Ambient socialization
+
+In addition to these strategies, we can take advantage of other technologies to help us feel connected in an ambient way.
+
+* We all open [Teletype](https://teletype.atom.io/) portals as we work, even when not actively pairing, and share the URL in Slack. We join each other's portals in a window on a separate Atom window and watch each other's progress as a background process.
+* We stream to the world on [Twitch](https://twitch.tv) as we work. We sometimes jump into each other's streams to chat or catch up.
```
vanessayuenn: I love this idea! It can also increase transparency to external contributors, and perhaps make "contributing to oss" look less scary to people who are on the fence/curious about contributing.

docs/core-team-process.md
smashwilson: Yeah, you're right. When I'm working alone I tend to iterate pretty quickly up and down abstraction layers. As soon as we're stretched across timezones we risk lots of running down blind alleys.

Here's what's in my head right now: each set of developers who are online synchronously can divide work as we divide in Seams; as that set changes when people come online and drop offline, we do Handoffs to pass context along.

For example, say:

1. You come online first in your morning and work solo for a few hours, shifting up and down the stack just like you would otherwise.
2. When I come online, I get caught up on the work you've pushed so far and chat with you to sync up on progress. We divvy up areas of work to focus on for the next few hours, chatting in Slack as we go.
3. @annthurium and @kuychaco come online next. We bring them up to speed and subdivide further. Maybe they pair on the view work while we work on the model and controller.
4. When you're done for the day, you summarize how far you got on your bit. One of the three of us catches up, picks up where you left off, and keeps it going. I do the same when I log off.
5. When @annthurium and @kuychaco are finishing up they leave a quick writeup of their collective progress.
7. The next morning, you read the diff and the writeup and get traction on continuing through your day.
8. ...and repeat. ♻️ 

Does that sound rational?

docs/core-team-process.md
smashwilson: Yeah, I think all of these approaches have some higher risk of merge conflicts. This one defers the resolution until it would be maximally painful though.

docs/core-team-process.md
smashwilson: I'm also [smashwilson](https://www.twitch.tv/smashwilson) on Twitch :wink:

I've never streamed myself _coding_ before though. Oddly I think I'm most nervous about people judging my taste in music in doing that 😜

docs/core-team-process.md
vanessayuenn: it does!

Issue #1689, opened 2018-09-11
Multi-File Diff


smashwilson:
We're accumulating several places where we have a need to present a diff that spans multiple files in a single view, a la the ["files changed" tab on a GitHub pull request](https://github.com/atom/github/pull/1685/files):

* The pull request details item proposed by the RFC in #1706.
* The CommitPaneItem proposed in #1655.

The easiest place to introduce this is by modifying the `FilePatchItem` to display diffs for multiple files at once when you cmd- or ctrl-click multiple files in the unstaged or staged changes lists in the git tab, so let's use that as our first use case.

We can't really start building this until the FilePatchItem is revamped in #1512, but let's start talking about what the design for a multi-file diff view should look like, so we can hit the ground running once it lands 🐎 .

Essentially what we need to design is a re-usable React component that:

* Shows the list of files contained in a multi-file diff.
* Shows the combined diff, with some kind of prominent divider between each file's portions of the diff.
* Makes it easy to jump quickly to a specific file you care about, or back to the file list to get to another file.
* Makes it apparent which file a part of a long diff belongs to, so you don't lose track of what file you're looking at if its diff is larger than you can fit in the pane at once.

/cc @annthurium, @kuychaco, and @simurai for input on what this should look like and how it should behave :smile:

kuychaco:
@smashwilson thanks for starting the conversation!

One question I have upfront is -- where does it make sense to replicate what dotcom has, and where does it make sense to deviate?

For instance, for the following I think we can draw heavy inspiration from dotcom...
* Shows the list of files contained in a multi-file diff.
* Makes it easy to jump quickly to a specific file you care about, or back to the file list to get to another file.

... and create a similar sort of dropdown menu that is in a floating header that is always visible:
<img width="719" alt="add_metrics_collection_for_committing_partial_changes__undoing_the_last_commit__and_opening_an_expanded_commit_msg_editor_by_kuychaco_ _pull_request__1685_ _atom_github" src="https://user-images.githubusercontent.com/7910250/45379610-63085a80-b5b5-11e8-8bf3-6ad356aea258.png">

From what I can tell dotcom doesn't really do this...
* Makes it apparent which file a part of a long diff belongs to, so you don't lose track of what file you're looking at if its diff is larger than you can fit in the pane at once.

... so maybe we can add the filename as another floating header that remains visible while scrolling (located right below the header that contains the file-navigator)

simurai:
[...]

Issue #1655, opened 2018-08-20
Commit pane item


smashwilson:
Let's design a pane item to show the contents of a single commit. :art: The work in #1512 will allow us to show much larger diffs in a performant way, which opens the door to a full-commit view.

We can access it by:

* Clicking on an entry in the recent commits view
* (Maybe) showing the combined staged changes while the keyboard focus is in the commit message editor, like a `git commit -v`
* (Eventually) clicking on a commit entry in the issueish pane item
* (Eventually) from diving into a single commit on the log view (#1419)

We'll want to show the full diff with some kind of navigation controls to allow you to move among the changed files. It'd also be a good place to put controls for contextual actions that you can take on that commit: amending the most recent, hard and soft resets, revert, checkout, create and checkout a new branch.

simurai:
> We'll want to show the full diff with some kind of navigation controls to allow you to move among the changed files.

Does this mean that in the same pane, there are diffs of multiple files stacked on top of each other, like on .com and the navigation controls would change the scroll position? That would be nice for allowing:

> (Maybe) showing the combined staged changes while the keyboard focus is in the commit message editor

Like when there are 2 changed files you don't have to switch between them and could just scroll and see the diff of both files.

simurai:
Mockups for:

> (Maybe) showing the combined staged changes while the keyboard focus is in the commit message editor, like a `git commit -v`

![commit-preview](https://user-images.githubusercontent.com/378023/47568149-54aaae00-d96b-11e8-833a-98356b69db29.png)

In the mockup, there is a "Preview" button that can be clicked to see a preview of the next commit... buuuuuut, it _could_ show automatically on focus when there is already a `FilePatchView` open. Like when you clicked on a file to see the diff, then why not re-use that same pane for the preview when focusing the commit message editor. 👍 

If you click the "Preview" button, but there is no commit message, it just shows `...`.

![image](https://user-images.githubusercontent.com/378023/47568070-17deb700-d96b-11e8-8707-25d4ee0d3667.png)


> Clicking on an entry in the recent commits view

![commit-pane](https://user-images.githubusercontent.com/378023/47567450-8a4e9780-d969-11e8-9a51-96d327d0716b.png)

PR #1753, opened 2018-10-22
RFC for multi-file diff

vanessayuenn:
This is a WIP RFC summarizing the dicussion in #1689 (btw should we close that issue in favor of keeping the discussion here?). A few of the sections are left as `TBD` for now and will be filled in as the conversation continues and things become clearer in terms of direction. 

📜[Rendered Version](https://github.com/atom/github/blob/multi-diff-rfc/docs/rfcs/004-multi-file-diff.md)

Cc-ing @kuychaco @smashwilson @annthurium @simurai for input & mock-up help. 🙇‍♀️


kuychaco:
Thanks @vanessayuenn for getting this off to a great start! I appreciated the thoughtfulness of your "Drawbacks" section. It inspired me to think of an alternative approach that might be easier and more straight forward to design and implement.

> cmd-click to select multiple files might not be as universally known as we assume, so that might affect discoverability of this feature.

I agree with you that it might be tricky to get the discoverability and UX of this just right. Instead we can opt to lessen scope and _simply offer a way for users to view the collective set of staged changes that they are about to commit._

This could look like a button above the commit message box that, when clicked, opens a multi-file diff pane item called something like "Commit Preview" and shows a summary of what will go into the user's next commit based on what is currently staged.

Benefits to this approach:

**Minimal scope:** 
It's the simplest use case for a multi-file diff that I can think of, allowing us to focus primarily on building out the multi-file diff item and not having to worry about other complex or nuanced UX interactions

**Pareto Principle (aka 80/20 rule)**
> 80% of the results will come from just 20% of the actions
This proposal covers the most common use case for viewing changes for multiple files prior to committing. When I imagine a user's workflow, I actually can't think of any other use cases. I'm sure they exist, but it's probably not worth the added complexity to support them.

**Educate and delight users:** 
By using helpful language like "View changes to be committed" and "Commit Preview" we can help make Git more transparent and understandable to users. For example, when teaching Git I often encounter users who are confused by Git's two-stage commit process and don't immediately see the value for it. Thus they never partially stage any changes. 

We can also encourage good practices like reviewing all the changes you are about to commit. This would be a nice addition to the flow that currently exists in our panel: (from top to bottom) view unstaged changes -> stage changes to be committed -> view all staged changes to be committed -> write commit message that summarizes all changes -> hit commit button -> see commit appear in recent commits list

Here's my best shot at creating a visual for ya:

git_ ___src_test-repo_and_how-we-work_md_ ___github_github

@simurai what does your UX spidey sense say about this?

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,65 @@
...
+
+
+## Explanation
+
+#### Mock-ups coming soon
+
+
+#### Unstaged Changes pane
+- User can `cmd+click` and select multiple files from the list of unstaged changes, and the pane on the left (see multi-file diff section below) will show diffs of the selected files. That pane will continue to reflect any further selecting/unselecting on the Unstaged Changes pane.
+- Once there is at least one file selected, `Stage All` button should be worded as `Stage Selected`.
```
kuychaco: > Once there is at least one file selected, `Stage All` button should be worded as `Stage Selected`.

Personally I think I prefer the current functionality where the `Stage All` button remains, regardless of how many changed files are selected. It's kind of like in the diff view where there's a `Stage File` button at the top. It's a quick way to stage everything regardless of what's selected. I could possibly be convinced otherwise though, especially if there's UXR that shows that users would expect the button to change.

Users can currently stage selected files by hitting enter or right-clicking and selecting `Stage`. Admittedly these are less discoverable than a button that says `Stage Selected`

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,65 @@
...
+
+#### Staged Changes pane
+- Same behavior as Unstaged Changes pane.
+- Once there is at least one file selected, `Unstage All` button should be worded as `Unstage Selected`.
+
+#### Multi-file diff view
+_(note: The following is a summary of what we would like the UX to achieve, but I don't have a clear visuals of what that looks like yet.)_
+
+- Shows diffs of multiple files as a stack.
+- Each diff should show up as its own block, and the current functionality should remain independent of each block.
```
kuychaco: > Each diff should show up as its own block, and the current functionality should remain independent of each block.

Not quite sure what this means 🤔. Could you please elaborate?

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,65 @@
...
+
+
+## Drawbacks
+
+- `cmd-click` to select multiple files might not be as universally known as we assume, so that might affect discoverability of this feature.
+- There might be performance concerns having to render many diffs at once.
+
+## Rationale and alternatives
+
+An alternative would be to _not_ implement multi-file diff, as other editors like VS Code also only has per-file diff at the time of writing. However, not implementing this would imply that [the proposed new PR review flow](https://github.com/atom/github/blob/master/docs/rfcs/003-pull-request-review.md) will have to find another solution to display all changes in a PR.
```
kuychaco: Another drawback to this alternative worth mentioning is that only displaying one-file diffs means that users have to do a lot more clicking to view all of their changes. Imagine that they have 3 files changed each with a very small change (say a model file, a view file, a test file) and they want to see them all next to each other. Or say there was a variable rename and only 10 lines are changed, but they are each in a different file. It'd be a bit of a pain to click through to view each one. 

Also, if we didn't implement multi-file diffs then we couldn't show commit contents since they often include changes across multiple files

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,65 @@
...
+- Once there is at least one file selected, `Stage All` button should be worded as `Stage Selected`.
+
+#### Staged Changes pane
+- Same behavior as Unstaged Changes pane.
+- Once there is at least one file selected, `Unstage All` button should be worded as `Unstage Selected`.
+
+#### Multi-file diff view
+_(note: The following is a summary of what we would like the UX to achieve, but I don't have a clear visuals of what that looks like yet.)_
+
+- Shows diffs of multiple files as a stack.
```
kuychaco: It would be cool if each diff was collapsable. I've seen this is some of @simurai's mockups. "Collapse/Expand All" capabilities would be nice as well

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,65 @@
...
+- Each diff should show up as its own block, and the current functionality should remain independent of each block.
+- It should be easy to jump quickly to a specific file you care about, or back to the file list to get to another file. Dotcom does so by creating a `jump to` drop down.
+- As user scrolls through a long list of diffs, there should be a sticky heading which remains visible showing the filename of the diff being viewed.
+
+
+## Drawbacks
+
+- `cmd-click` to select multiple files might not be as universally known as we assume, so that might affect discoverability of this feature.
```
kuychaco: Really good point 👍

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Rationale and alternatives
+
+An alternative would be to _not_ implement multi-file diff, as other editors like VS Code also only has per-file diff at the time of writing. However, not implementing this would imply that [the proposed new PR review flow](https://github.com/atom/github/blob/master/docs/rfcs/003-pull-request-review.md) will have to find another solution to display all changes in a PR.
+
+## Unresolved questions
+
+- What unresolved questions do you expect to resolve through the implementation of this feature before it is released in a new version of the package?
+
+TBD
```
kuychaco: One question I have is how exactly to construct the multi-file diffs. Do we have one TextEditor component that has different sections for each file. Or do we create a new type of pane item that contains multiple TextEditors stacked on top of one another, one for each file diff. 

If we do the former we could probably get something shipped sooner (we could just get the diff of the staged changes from Git, add a special decoration for file headers, and present all the changes in one editor). But to pave the way for a more complex code review UX I think taking extra time to do the latter will serve us well. For example, I can imagine reviewers wanting to collapse some files, or mark them as "Done", in which case it would be easier if we treated each diff as its own component.

docs/rfcs/004-multi-file-diff.md
kuychaco: Forgot you said to just push to the branch. I went ahead and added this :)

docs/rfcs/004-multi-file-diff.md
kuychaco: Pushed a commit to add this 👍

docs/rfcs/004-multi-file-diff.md
kuychaco: Added this to the out-of-scope section cuz I think we can punt on this until we get closer to tackling code review in the editor

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,67 @@
...
+- [new PR review flow](https://github.com/atom/github/blob/master/docs/rfcs/003-pull-request-review.md) that shows all changed files proposed in a PR
+
+
+## Explanation
+
+#### Mock-ups coming soon
+
+
+#### Unstaged Changes pane
+- User can `cmd+click` and select multiple files from the list of unstaged changes, and the pane on the left (see multi-file diff section below) will show diffs of the selected files. That pane will continue to reflect any further selecting/unselecting on the Unstaged Changes pane.
```
simurai: Another way is to `click + drag up/down` on multiple files. Kinda hard to discover (I just rediscovered it :smile:), but quite nice once you know.

simurai:
> @kuychaco This could look like a button above the commit message box that, when clicked, opens a multi-file diff pane item called something like "Commit Preview" and shows a summary of what will go into the user's next commit based on what is currently staged.

I really like the idea of a "Commit Preview". 👍 Here 3 alternatives to the button above the commit message box.

### 1. Use the headers to select all files

Clicking on "Unstaged Changes" or "Staged Changes" would select all files (for each section) and thus show the multi-file diff on the left.

screen shot 2018-10-24 at 5 31 52 pm

**Downside**: Harder to discover. Might might can add a hover style.

### 2. Add a "Preview" button next to the commit button

The flow would be more like 1. Stage 2. Write commit message 3. Preview (optional) 4. Commit

screen shot 2018-10-24 at 5 41 43 pm

**Downside**: There is no preview button for all unstaged changes. Long branch names get cut off.

### 3. Show a preview automatically when the commit message has focus

No buttons to click, always get a preview when typing the commit message.

screen shot 2018-10-24 at 5 46 27 pm

**Downside**: This could feel jarring and unexpected.

----

Anyways, I'll try to add some mockups tomorrow.

vanessayuenn:
@kuychaco I am very into your idea of commit preview! I originally typed a long response but I realized I wasn't really adding anything new but rather just repeating what you said. 😅 Anyway I concur with your rationale of the new approach, and I think we should make the multi-diff view the focus of the work, and introduce this feature to be user-facing in the most efficient and frictionless way possible. Commit preview checks both of these boxes, and I would be happy to update the RFC to reflect this change of direction.

docs/rfcs/004-multi-file-diff.md
vanessayuenn: I realize now that it was quite badly worded. I meant the current functionality (e.g. the `open file`, `stage file`, etc from the header) of the diff view for the file it's showing should remain to be file-specific, even though the view would show more than one file. Does that clarify things? If not I can attempt to illustrate with mock-ups. Words are hardddd...

docs/rfcs/004-multi-file-diff.md
annthurium: I knew about cmd-click, but didn't realize Atom supported it in the staged files area.  I wonder if there's a way we can educate users about this, either in product education / tutorials, or with marketing and educational materials (videos, etc.)

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,67 @@
...
+- It should be easy to jump quickly to a specific file you care about, or back to the file list to get to another file. Dotcom does so by creating a `jump to` drop down.
+- As user scrolls through a long list of diffs, there should be a sticky heading which remains visible showing the filename of the diff being viewed.
+Nice-to-have UX that doesn't necessarily need to be implemented
+- Each file diff can be collapsed (this can potentially be )
+
+
+## Drawbacks
+
+- `cmd-click` to select multiple files might not be as universally known as we assume, so that might affect discoverability of this feature.
+- There might be performance concerns having to render many diffs at once.
```
annthurium: Hopefully @smashwilson's work on making diffs use a `TextEditor` will make performance a non issue.  Regardless, we should find a way of measuring perf in general and for diff rendering specifically.

docs/rfcs/004-multi-file-diff.md
annthurium: I'm also not sure what this means.  In a visual sense, do you mean each file in the diff is separated visually by a border and some margin?

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Unresolved questions
+
+- What unresolved questions do you expect to resolve through the implementation of this feature before it is released in a new version of the package?
+
+How exactly do we construct the multi-file diffs? Do we have one TextEditor component that has different sections for each file. Or do we create a new type of pane item that contains multiple TextEditor components stacked on top of one another, one for each file diff... If we do the former we could probably get something shipped sooner (we could just get the diff of the staged changes from Git, add a special decoration for file headers, and present all the changes in one editor). But to pave the way for a more complex code review UX I think taking extra time to do the latter will serve us well. For example, I can imagine reviewers wanting to collapse some files, or mark them as "Done", in which case it would be easier if we treated each diff as its own component.
```
annthurium: I agree - it seems more robust to have each file be its own `TextEditor`. It also seems like editable diffs would be easier if we go this route.  If we have one large `TextEditor` that has data from multiple files, when you save a buffer, we have to map that to which file that data lives in, which seems annoying.

docs/rfcs/004-multi-file-diff.md
kuychaco: Yup I think that makes sense! Thanks for the clarification. Maybe "Each diff retains the controls it currently has in its header (e.g. the open file, stage file, undo last discard, etc)"

kuychaco:
@simurai @vanessayuenn glad you're on board with the commit preview direction :)

@simurai thanks for proposing UX ideas! I like the "add a 'Preview' button next to the commit button" one the best. One thing to consider is that the "Abort Merge" button shows up in the same spot when there's a merge conflict. As long as things don't get too squishy I think this could be a great place for the "Preview" button. 

I do think that "Use the headers to select all files" would suffer discoverability issues and agree that "Show a preview automatically when the commit message has focus" would be jarring and unexpected.

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,72 @@
...
+
+#### Unstaged Changes pane
+- User can `cmd+click` and select multiple files from the list of unstaged changes, and the pane on the left (see multi-file diff section below) will show diffs of the selected files. That pane will continue to reflect any further selecting/unselecting on the Unstaged Changes pane.
+- Once there is at least one file selected, `Stage All` button should be worded as `Stage Selected`.
+
+#### Staged Changes pane
+- Same behavior as Unstaged Changes pane.
+- Once there is at least one file selected, `Unstage All` button should be worded as `Unstage Selected`.
+
+![staged changes](https://user-images.githubusercontent.com/378023/47497740-0a0b3200-d896-11e8-85af-7c644af9ca37.png)
```
simurai: Added some mockups, but I'm not so sure about this part:

screen shot 2018-10-25 at 8 48 48 pm

The buttons to stage file and hunks look kinda weird when they are close together. Maybe we could only show them on a file when the file is collapsed.

docs/rfcs/004-multi-file-diff.md
smashwilson: I'm 👍 for leaving the "Stage All" behavior as it is, I think. At the very least we can revisit independently from this RFC to keep our scope under control :wink:

docs/rfcs/004-multi-file-diff.md
smashwilson: How does it look if the buttons were the same width? Or is that just more confusing?

docs/rfcs/004-multi-file-diff.md
smashwilson: Oh interesting. So, effectively, we move the FilePatchItem's header controls as-is to the per-file banner within the multi file patch view? I like that - it'd be a lot easier than trying to figure out which file the file-specific buttons map to 😆

docs/rfcs/004-multi-file-diff.md
smashwilson: It should help, depending on the way we implement!

It'd likely be worth it to track down some Giant-Ass Diffs to test this with to suss out edge cases. I'm thinking a Go project where they just took out vendored dependencies, say.

docs/rfcs/004-multi-file-diff.md
smashwilson: > [..] If we do the former we could probably get something shipped sooner [..]

Actually, I think it'd be _easier_ to implement the multiple-`TextEditor` approach:

```js
<div className="github-MultiFilePatchView">
  <AtomTextEditor /* file one */ />
  <AtomTextEditor /* file two */ />
  <AtomTextEditor /* file three */ />
</div>
```

Using one `<AtomTextEditor>` to render all patches, with the file headers as block decorations, would involve some more complicated fiddling in the model layer to keep everything straight - even more-so for editability, as @annthurium noted above.

The tradeoff I see is that the single `TextEditor` approach is likely to scale better to patches with large numbers of files - by pushing complexity into the decorations within the `TextEditor`, we gain the performance boost associated with the editor's tiling and scrolling optimizations.

smashwilson:
> This could look like a button above the commit message box that, when clicked, opens a multi-file diff pane item called something like "Commit Preview" and shows a summary of what will go into the user's next commit based on what is currently staged.

👍 I'm also on board with this, by the way.

gauravchl:
Wow this is so cool!
While working on commit template, i had a similar idea of "Commit Preview" with multi diffs but for existing commits. 
So i published an atom package for it: https://github.com/gauravchl/commit-book

Couldn't resist to share it here :)


screen shot 2018-10-26 at 10 59 53 am
screen shot 2018-10-26 at 11 00 04 am

docs/rfcs/004-multi-file-diff.md
simurai: Certain diffs could also be loaded only on demand. Like when the diff is very large. Or when the whole file got deleted:

image

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,89 @@
...
+A new button added above the commit message box that, when clicked, opens a multi-file diff pane item called something like "Commit Preview" and shows a summary of what will go into the user's next commit based on what is currently staged.
+
+![commit preview button](https://user-images.githubusercontent.com/378023/47554979-afc9aa00-d945-11e8-9953-45925e3278b9.png)
+
+#### Multi-file diff view
+
+![commit preview](https://user-images.githubusercontent.com/378023/47555097-e6072980-d945-11e8-9c29-05624825d9f8.png)
+
+- Shows diffs of multiple files as a stack.
+- Each diff retains the file-specific controls it currently has in its header (e.g. the open file, stage file, undo last discard, etc).
```
simurai: In the mockups I changed it a bit to:

Description | Mockup
--- | ---
Open a file by clicking on the left of a hunk header (`<>` icon). It open the file, but also scrolls to the appropriate line. | open file
Unstage the entire file or individual hunks by clicking on the up arrow icons. | image
When selecting lines, an additional "Unstage Selected Lines" button is shown. | image

For the Commit Preview, there is no discard button, since I _think_ discarding is only possible for unstaged changes.

The above changes could easily be moved to "Out of scope" and considered at a later point. Then we can just re-use what we currently have for each file:

image

simurai:
Here a [prototype](https://www.figma.com/proto/PF3F3BU7I9F3T6kt9afF29/Multi-file-diff?node-id=1%3A6&viewport=114%2C438%2C0.5&scaling=contain) to click around. Single-file and multi-file diffs have the same layout to make adding/removing files less jarring. "File blocks" just get added or removed, but otherwise the layout stays the same.

And here an alternative where the "Preview" button is next to the commit button. It also shows the avatar and commit message above the diff:

4-commit-preview-2 copy

It would be like a [commit pane](https://github.com/atom/github/issues/1655), just without the SHA and the `committed 1 day ago`. Once you click commit, they would get added.

docs/rfcs/004-multi-file-diff.md
```diff
@@ -0,0 +1,89 @@
...
+"Find" input field for filtering diffs based on search term (which could be a file name, an author, a variable name, etc). When filtering, files that have no match get collapsed. This allows you to uncollapse files (and seeing their diff) without having to clear the filter. Matches get highlighted with a yellow overlay as well as a stripe on the side, similar to git-diff in the editor.
+
+Unfiltered | Filtered
+--- | ---
+![without filter](https://user-images.githubusercontent.com/378023/47497740-0a0b3200-d896-11e8-85af-7c644af9ca37.png) | ![with filter](https://user-images.githubusercontent.com/378023/47540019-116e2200-d90e-11e8-8d22-d305328d55c4.png)
+
+**Alternative**: It might be possible to re-use the find+replace UI to filter the multi-file diff. And maybe even have "replace" working.
+
+#### Other out of scope UX considerations
+- whether `cmd+click` to select multiple files is discoverable
```
kuychaco: Is this still relevant given that we're just going with the "Commit preview" button?

docs/rfcs/004-multi-file-diff.md
kuychaco: >  I think discarding is only possible for unstaged changes.

This is true 👍

smashwilson:
✨ 🚧 🚧 ✨ Some random thoughts about the implementation we can pursue. Names and architecture are purely illustrative and up to whomever is ultimately writing the code :smile:

* Let's have a `MultiFilePatch` model that includes an ordered list of `FilePatch` objects.
* We'll need to extend the parser to understand multi-file `git diff` output. There may be some tricky work in here in relation to the deleted-symlink-and-created-file and deleted-file-and-created-symlink edge cases.
* Add a method to our git repository plumbing to get the multi-diff from all staged changes, _or_, preferably, find a way to get this information using existing git repository methods.
* Let's have a `CommitPreview` component hierarchy for the React side of things: `CommitPreviewItem` for the workspace item stuff, `CommitPreviewContainer` for the loading and error logic.
* For the rest of the components, we have a decision to make. We can either: _(a)_ generalize the existing `FilePatch` classes (controller and view) to be usable in both situations, or _(b)_ create new components that are specialized for the multi-file case. I'm leaning a little toward (a) but that might change when we get down into the code.
* Finally, we'll need to do the addition of the "preview" button within the `CommitView` and the corresponding wiring for the preview action in `CommitController`. We'll want to be sure everything looks 👌 even while the "abort merge" button is visible.

PR #1713, opened 2018-10-03
Feature: Add commit template support

gauravchl:
### Description of the Change

This changes bring the support for commit template file.
It will automatically populate the commit message from template file into commit message box.

Steps to test this feature:
1. Create a template file (~/.gitmessage.txt) with some default commit message.
2. Update git config `git config commit.template ~/.gitmessage.txt`.
3. Reload the atom, it should load default commit message into commit input box.

More details about commit template is available [here](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_code_commit_template_code).

### Alternate Designs 

- Show a button "load message from template" next to "add co-authors"
- Don't load any commit message by default to input field, load only when user click on load button.
- If there is a message in commit input field, show another button "save message to template". Clicking on this button should update the commit.template file.

### Benefits

Users don't have to save common commit messages elsewhere and copy/paste manually.

### Possible Drawbacks

N/A

### Applicable Issues

https://github.com/atom/github/issues/1469

### Metrics

- Added a new method 'getCommitMessageFromTemplate' to git-shell-out-stratagy 
- A new fixture repo added with commit.template for testing(35 files)

### Tests

Added unit tests to verify that 
- It populates the commit message from template
- Reload the commit message from template after committing
- It correctly updates the state when switching repos
- `git.getCommitMessageFromTemplate()` works

PR #1756, opened 2018-10-25
Re-introduce commit message template

kuychaco:
### Description of the Change

This PR re-introduces the commit message template feature, originally introduced in #1713. It re-works some of the implementation, cleans up tests, and additionally fixes the merge message regression. 

With these change, the repository model now manages and updates the commit message based on file system events indicating that the config has changed or a merge has been initiated/completed. 

/cc @annthurium 

### Alternate Designs

N/A

### Benefits

Folks will be able to make commits with the aid of commit message templates

### Possible Drawbacks

The usual drawbacks associated with code changes. This feature in particular touches a part of the system that is complex and fairly edge-casey...

### Applicable Issues

Original PR - #1713
Reverted PR - #1754

### Metrics

N/A

### Tests

Added unit tests for:
* restores template after committing
* config commit.template change
* when merge is initiated 
  * merge message replaces 
    * empty commit message
    * untouched commit template
  * dirty message is left untouched
* when merge is aborted
  * clean commit message is restored
  * commit template is restored

Manually tested:
* When `tab`bing to commit box, cursor is at beginning of template (rather than at the very end)
* Commit button is disabled when commit message contains a template. It is enabled only when commit message contains non-commented lines
* Commit message updates correctly when expanding/closing expanded commit message editors
* Commit message updates correctly to show merge message when merge is initiated and aborted
* Commit message template is restored after committing

@smashwilson to test manually on Linux and Windows

N/A

TODO:
- [x] clean up git strategies tests
- [x] de-dupe file system watching tests
- [x] verify template path logic works on windows and linux


lib/views/commit-view.js
```diff
@@ -588,6 +591,11 @@ export default class CommitView extends React.Component {
 
     if (focus === CommitView.focus.EDITOR) {
       if (this.refEditor.map(focusElement).getOr(false)) {
+        if (this.editor && this.editor.getText().length > 0 && !this.isValidMessage()) {
```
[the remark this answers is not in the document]
annthurium: although...are there any weird edge cases where somebody moves focus around, and they might actually want the cursor at the end?

lib/git-shell-out-strategy.js
```diff
@@ -415,6 +415,39 @@ export default class GitShellOutStrategy {
...
+    // this regex attempts to get the specified user's home directory
+    // Ex: on Mac ~kuychaco/ is expanded to the specified user’s home directory (/Users/kuychaco)
+
+    // Regex translation:
+    // ^~ line starts with tilde
+    // ([^/]*)/ captures non-forwardslash characters before first slash
+    const regex = new RegExp('^~([^/]*)/');
+    templatePath = templatePath.trim().replace(regex, (_, user) => {
+      // if no user is specified, fall back to using the home directory.
+      return `${user ? path.join(path.dirname(homeDir), user) : homeDir}/`;
```
annthurium: @smashwilson: can you test this on non-Mac operating systems?  We want to make sure if a username is specified, it does the right thing for our Linux and or Windows users.

lib/models/repository-states/present.js
```diff
@@ -39,20 +39,46 @@ export default class Present extends State {
     this.operationStates = new OperationStates({didUpdate: this.didUpdate.bind(this)});
 
     this.commitMessage = '';
+    this.commitMessageTemplate = null;
+    this.fetchInitialMessage();
```
smashwilson: I wonder if the commit message, commit message template, and merge message tracking belong more naturally in `Repository` rather than `Present`. Most methods in `Present` are intended to call a single git command with `GitShellOutStrategy`, manipulate the cache, and translate input and output to our model objects, while `Repository` methods implement higher-order composite behavior. Commit message derivation feels like a "composite" behavior to me 🤔 

With that said: we already have a bunch of behavior in `Repository` and `Present` that blurs these lines, and it would not be trivial to move all of this there. So maybe we could leave this as a future refactoring as part of a broader effort to make the division among those abstraction layers clear.

lib/models/repository-states/present.js
```diff
@@ -39,20 +39,46 @@ export default class Present extends State {
...
+    }
   }
 
   getCommitMessage() {
     return this.commitMessage;
   }
 
+  getCommitMessageTemplate() {
+    return this.git().getCommitMessageTemplate();
+  }
```
smashwilson: It's odd to me that `getCommitMessageTemplate()` is a git operation, but `setCommitMessageTemplate()` is an accessor for an instance variable. Can we clarify the names to be less misleading somehow?

Also: should the result of this git operation be cached?

Fake edit: on reflection, caching this would cause some weird behavior because we don't have a filesystem watcher on the commit message template itself, so we wouldn't know when to invalidate that cache entry.

lib/models/repository-states/present.js
```diff
@@ -144,6 +171,46 @@ export default class Present extends State {
...
+    } else if (this.commitMessageTemplate) {
+      return this.commitMessage === this.commitMessageTemplate;
+    }
+    return false;
+  }
+
+  async updateCommitMessageAfterFileSystemChange(events) {
+    for (let i = 0; i < events.length; i++) {
+      const event = events[i];
+      const endsWith = (...segments) => event.path.endsWith(path.join(...segments));
```
smashwilson: Might be worth extracting this to `helpers.js`?

It _is_ a one-liner, but eh.

lib/views/commit-view.js
```diff
@@ -460,13 +460,18 @@ export default class CommitView extends React.Component {
     }
   }
 
+  isValidMessage() {
+    // ensure that there are at least some non-comment lines in the commit message.
+    // Commented lines are stripped out of commit messages by git, by default configuration.
+    return this.props.messageBuffer.getText().replace(/^#.*$/gm, '').trim().length !== 0;
```
smashwilson: :+1: Cool, cool.

lib/views/commit-view.js
```diff
@@ -580,6 +585,11 @@ export default class CommitView extends React.Component {
 
     if (focus === CommitView.focus.EDITOR) {
       if (this.refEditorComponent.map(focusElement).getOr(false)) {
+        if (this.props.messageBuffer.getText().length > 0 && !this.isValidMessage()) {
+          // there is likely a commit message template present
+          // we want the cursor to be at the beginning, not at the and of the template
+          this.refEditorComponent.get().getModel().setCursorBufferPosition([0, 0]);
```
smashwilson: Hah. Weird edge case, good catch!

test/integration/file-patch.test.js
```diff
@@ -732,6 +732,8 @@ describe('integration: file patches', function() {
       });
 
       it('may be partially staged', async function() {
+        this.retries(5); // FLAKE
```
smashwilson: 🙇

test/models/repository.test.js
```diff
@@ -2125,4 +2124,193 @@ describe('Repository', function() {
...
+
+      sub.add(observer.onDidChange(events => {
+        observedEvents.push(...events);
+        eventCallback();
+      }));
+
+      return {repository, observer};
+    }
+
+    function expectEvents(repository, ...suffixes) {
```
smashwilson: Yeah, these would be great to extract to a helper and clean up a bit... We have a few places that test Real Filesystem Events :tm: where these would be nice.

lib/git-shell-out-strategy.js
```diff
@@ -422,6 +422,39 @@ export default class GitShellOutStrategy {
...
+     * https://git-scm.com/docs/git-config#git-config-pathname
+     */
+    const homeDir = os.homedir();
+    // this regex attempts to get the specified user's home directory
+    // Ex: on Mac ~kuychaco/ is expanded to the specified user’s home directory (/Users/kuychaco)
+
+    // Regex translation:
+    // ^~ line starts with tilde
+    // ([^/]*)/ captures non-forwardslash characters before first slash
+    const regex = new RegExp('^~([^/]*)/');
```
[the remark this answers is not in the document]
annthurium: We didn't really have a strong reason for preferring the regexp constructor over literals here.  We went with what the original author had.

Out of curiosity, we looked to see if there was a performance difference.  Answer: not really.
https://www.measurethat.net/Benchmarks/Show/1734/1/regexp-constructor-vs-literal

We did give it a more descriptive name, and moved it out of this function so as to avoid creating a new regex every time.

lib/git-shell-out-strategy.js
```diff
@@ -422,6 +422,39 @@ export default class GitShellOutStrategy {
...
+    templatePath = templatePath.trim().replace(regex, (_, user) => {
+      // if no user is specified, fall back to using the home directory.
+      return `${user ? path.join(path.dirname(homeDir), user) : homeDir}/`;
+    });
+
+    if (!path.isAbsolute(templatePath)) {
+      templatePath = path.join(this.workingDir, templatePath);
```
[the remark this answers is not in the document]
kuychaco: Yup! See https://stackoverflow.com/questions/21998728/how-to-specify-a-git-commit-message-template-for-a-repository-in-a-file-at-a-rel

Tested this locally to verify

lib/models/repository-states/present.js
annthurium: cool - should I open an issue to refactor this in the future?  I'm on board with cleaning it up.

lib/models/repository-states/present.js
annthurium: hahaha yup Katrina and I went through that exact conversation re caching.

renamed the method to `fetchCommitMessageTemplate`.  I'm not 100% sure that's better, but perhaps folks won't think it's the inverse operation of `setCommitMessageTemplate`?

lib/models/repository-states/present.js
annthurium: sure!  Good suggestion.  Extracted.

test/models/repository.test.js
kuychaco: How does this look b3c9ec7?

smashwilson:
I just found a bug on Windows. If I run `git config --unset commit.template` at the command line, the git tab errors out and leaves this stack in the console:

```
C:\Users\smash\src\atom\node_modules\text-buffer\lib\text-buffer.js:1319 Uncaught TypeError: Cannot read property 'split' of null
    at Diff.lineDiff.tokenize (C:\Users\smash\src\atom\node_modules\text-buffer\node_modules\diff\lib\diff\line.js:22:32)
    at Diff.diff (C:\Users\smash\src\atom\node_modules\text-buffer\node_modules\diff\lib\diff\base.js:36:39)
    at Object.diffLines (C:\Users\smash\src\atom\node_modules\text-buffer\node_modules\diff\lib\diff\line.js:47:19)
    at transact (C:\Users\smash\src\atom\node_modules\text-buffer\lib\text-buffer.js:801:29)
    at TextBuffer.transact (C:\Users\smash\src\atom\node_modules\text-buffer\lib\text-buffer.js:1316:16)
    at TextBuffer.setTextViaDiff (C:\Users\smash\src\atom\node_modules\text-buffer\lib\text-buffer.js:796:10)
    at CommitController.componentDidUpdate (C:/Users/smash/src/github/lib/controllers/commit-controller.js:113:30)
    at commitLifeCycles (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:14256:22)
    at commitAllLifeCycles (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:15342:7)
    at HTMLUnknownElement.callCallback (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:100:14)
    at Object.invokeGuardedCallbackDev (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:138:16)
    at invokeGuardedCallback (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:187:29)
    at commitRoot (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:15481:7)
    at completeRoot (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:16496:34)
    at performWorkOnRoot (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:16440:9)
    at performWork (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:16358:7)
    at performSyncWork (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:16330:3)
    at requestWork (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:16230:5)
    at scheduleWork$1 (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:16096:11)
    at Object.enqueueSetState (C:\Users\smash\src\github\node_modules\react-dom\cjs\react-dom.development.js:11185:5)
    at ObserveModel.Component.setState (C:\Users\smash\src\github\node_modules\react\cjs\react.development.js:273:16)
    at ObserveModel.didUpdate (C:/Users/smash/src/github/lib/views/observe-model.js:39:12)
    at ModelObserver._refreshModelData (C:/Users/smash/src/github/lib/models/model-observer.js:52:14)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

From the message, I'm guessing that the fetched commit template is `null` instead of an empty string unexpectedly?

smashwilson:
Path logic is 👍 on Windows: I verified both `~/...` and `~username/...` paths. Checking Linux next...

smashwilson:
:+1: to path logic on Linux as well. (And I was able to reproduce the `--unset` bug I mentioned above, so it must not be platform-specific.)

lib/models/repository-states/present.js
```diff
@@ -196,7 +196,10 @@ export default class Present extends State {
 
       if (filePathEndsWith(event.path, '.git', 'config')) {
         // this won't catch changes made to the template file itself...
-        const template = await this.fetchCommitMessageTemplate();
```
kuychaco: Great! Thanks for tackling this @annthurium! 

I have a readability/logic suggestion -- in the case where there is no template we can update the commit message directly and null out our cached `commitMessageTemplate`. This way, we use the term `template` exclusively for what would be provided by Git as the commit message template and not artificially making it an empty string. And we maintain the current cohesion between `this.commitMessageTemplate` and the return value of `fetchCommitMessageTemplate()` (they are both `null` in the case of no template configuration).

```
const template = await this.fetchCommitMessageTemplate();
if (template === null) {
  this.setCommitMessage('');
  this.setCommitMessageTemplate(null);
} else if (this.commitMessageTemplate !== template) {
  this.setCommitMessageTemplate(template);
  this.setCommitMessage(template);
}
```

annthurium: cool, thanks for the suggestion.

since the commit message template is set in either case, I'll do that outside the if statements.  Other than that, did it the way you suggest.

PR #1757, opened 2018-10-25
Update dugite to 1.79.0

smashwilson:
Thanks, Greenkeeper!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2523,9 +2523,9 @@
       }
     },
     "dugite": {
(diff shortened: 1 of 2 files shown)


shiftkey:
@smashwilson I did a smoke test of this branch in Desktop last night (I think) with Ubuntu 18.04 and a clone that went through the `libcurl.so.4` it provided and seemed to work fine, but let me know if you spot anything unexpected with the packaging.

Issue #761, opened 2017-05-08
Flake: "GithubPackage ... beforeEach hook for refreshes the appropriate Repository"


smashwilson:
[AppVeyor build](https://ci.appveyor.com/project/Atom/github/build/1242/job/96l8rptl8ul2opq6/tests)

_GithubPackage / when there is a change in the repository / "before each" hook for "refreshes the appropriate Repository and Atom GitRepository when a file is changed in workspace 2"_

PR #1765, opened 2018-10-29
Retry known test flakes

smashwilson:
Flag known test flakes with `this.retries(5)` to limit the pain until we can schedule time to work them down again. Each of these flakes has already been filed with [an issue labelled flaky-test](https://github.com/atom/github/issues?q=is%3Aopen+is%3Aissue+label%3Aflaky-test).

### DIFF
--- a/test/github-package.test.js
+++ b/test/github-package.test.js
@@ -608,6 +608,8 @@ describe('GithubPackage', function() {
     let workdirPath2, atomGitRepository2, repository2;
 
     beforeEach(async function() {
+      this.retries(5); // FLAKE
+
       [workdirPath1, workdirPath2] = await Promise.all([
         cloneRepository('three-files'),
         cloneRepository('three-files'),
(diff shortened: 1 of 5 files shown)


test/github-package.test.js
```diff
@@ -608,6 +608,8 @@ describe('GithubPackage', function() {
     let workdirPath2, atomGitRepository2, repository2;
 
     beforeEach(async function() {
+      this.retries(5); // FLAKE
```
kuychaco: @smashwilson do we actually want this in the `beforeEach`? Is the goal with this to retry all of the tests in this `describe` block?

smashwilson: My intent was to retry the `beforeEach` itself if it throws. I was going off of #761 which puts the failure there instead of within any of the specific tests.

I'm... not actually sure if that's what Mocha will do here, though? The [documentation](https://mochajs.org/#retry-tests) doesn't make it clear. It doesn't break though?

Issue #1720, opened 2018-10-04
Editable diffs in FilePatchItems


smashwilson:
Once #1712 lands, the `FilePatchItem` will be rendering diffs with an Atom `TextEditor` component, but it will be in read-only mode. Let's let users edit diffs directly from the diff view! This saves you the trouble of needing to toggle to an editor and back again when you notice typos, `console.log()` statements, or `.only()` tests when reviewing unstaged or staged changes.

Editing a diff's contents should mark the item as "modified" and be treated by Atom like an unsaved text editor: it should cancel shutdown with a prompt to save and so on. Saving the item should apply the modifications to the underlying resource.

Saving edits to a diff within an "Unstaged changes" file patch item modifies the **underlying file in your working copy**, as though you'd pressed `cmd-O` and edited the file there.

Saving edits to a diff within a "Staged changes" file patch item modifies the **staged index entry for this file**. This leaves the file on disk unchanged, but the saved modifications will be applied to the file that ultimately becomes a part of the next commit.

Editing addition or unchanged lines is straightforward: the edits are written through to the equivalent locations in the "new" file. When editing a removal, though, the removed line needs to "split" into an addition and a removal with the edit and the cursor placed on the addition line.

smashwilson:
I'd like to make it really clear somehow:

* When you can and cannot edit the diff that you're looking at
* If you do edit the diff, what will actually be changed when you save your edits

@simurai, do you have any thoughts about how we can accomplish this? Maybe we could add some kind of badge or flair to the patch header with some kind of pen icon when it's writeable, and some text describing the edit destination?

smashwilson:
Another consideration: when you're editing a diff that will be applied to a file in your working copy, we should update the underlying `TextBuffer` immediately, to update any other `TextEditors` that have that file open. This lets us stay consistent with the behavior when you have multiple `TextEditors` open on the same file. That also means we'll have to clean up after ourselves if the user closes the modified `FilePatchItem` and discards their changes.

This gets extra fun once we tackle #1689 and could have to modify different `TextBuffers` depending on the locations of specific edits 🙃

simurai:
> Maybe we could add some kind of badge or flair to the patch header with some kind of pen icon when it's writeable, and some text describing the edit destination?

Yeah, some sort of `read-only` badge would be great. Hovering could show a tooltip with the reason?

In VS Code, the cursor (caret) doesn't blink and if you start typing, a tooltip appears:

[...]

PR #1767, opened 2018-10-30
Commit Preview

smashwilson:
🌐 Collaborative pull request to implement the first chunk of behavior described in #1753.

- [ ] Let's have a `MultiFilePatch` model that includes an ordered list of `FilePatch` objects.
- [ ] We'll need to extend the parser to understand multi-file `git diff` output. There may be some tricky work in here in relation to the deleted-symlink-and-created-file and deleted-file-and-created-symlink edge cases.
- [ ] Add a method to our git repository plumbing to get the multi-diff from all staged changes, _or_, preferably, find a way to get this information using existing git repository methods.
- [ ] Let's have a `CommitPreview` component hierarchy for the React side of things: `CommitPreviewItem` for the workspace item stuff, `CommitPreviewContainer` for the loading and error logic.
- [ ] For the rest of the components, we have a decision to make. We can either: _(a)_ generalize the existing `FilePatch` classes (controller and view) to be usable in both situations, or _(b)_ create new components that are specialized for the multi-file case. I'm leaning a little toward (a) but that might change when we get down into the code.
- [ ] Finally, we'll need to do the addition of the "preview" button within the `CommitView` and the corresponding wiring for the preview action in `CommitController`.

### DIFF
--- a/lib/items/commit-preview-item.js
+++ b/lib/items/commit-preview-item.js
@@ -1,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Emitter} from 'event-kit';
+
+import {WorkdirContextPoolPropType} from '../prop-types';
+
+export default class CommitPreviewItem extends React.Component {
+  static propTypes = {
+    workdirContextPool: WorkdirContextPoolPropType.isRequired,
+    workingDirectory: PropTypes.string.isRequired,
+  }
+
+  static uriPattern = 'atom-github://commit-preview?workdir={workingDirectory}'
+
+  static buildURI(relPath, workingDirectory) {
+    return `atom-github://commit-preview?workdir=${encodeURIComponent(workingDirectory)}`;
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.emitter = new Emitter();
+    this.isDestroyed = false;
+    this.hasTerminatedPendingState = false;
+  }
+
+  terminatePendingState() {
+    if (!this.hasTerminatedPendingState) {
+      this.emitter.emit('did-terminate-pending-state');
+      this.hasTerminatedPendingState = true;
+    }
+  }
+
+  onDidTerminatePendingState(callback) {
+    return this.emitter.on('did-terminate-pending-state', callback);
+  }
+
+  destroy() {
+    /* istanbul ignore else */
+    if (!this.isDestroyed) {
+      this.emitter.emit('did-destroy');
+      this.isDestroyed = true;
+    }
+  }
+
+  onDidDestroy(callback) {
+    return this.emitter.on('did-destroy', callback);
+  }
+
+  render() {
+    return null;
+  }
+}
--- a/test/items/commit-preview-item.test.js
+++ b/test/items/commit-preview-item.test.js
@@ -1,0 +1,63 @@
+import React from 'react';
+import {mount} from 'enzyme';
+
+import CommitPreviewItem from '../../lib/items/commit-preview-item';
+import PaneItem from '../../lib/atom/pane-item';
+import WorkdirContextPool from '../../lib/models/workdir-context-pool';
+import {cloneRepository} from '../helpers';
+
+describe('CommitPreviewItem', function() {
+  let atomEnv, repository, pool;
+
+  beforeEach(async function() {
+    atomEnv = global.buildAtomEnvironment();
+    const workdir = await cloneRepository();
+
+    pool = new WorkdirContextPool({
+      workspace: atomEnv.workspace,
+    });
+
+    repository = pool.add(workdir).getRepository();
+  });
+
+  afterEach(function() {
+    atomEnv.destroy();
+    pool.clear();
+  });
+
+  function buildPaneApp(override = {}) {
+    const props = {
+      workdirContextPool: pool,
+      ...override,
+    };
+
+    return (
+      <PaneItem workspace={atomEnv.workspace} uriPattern={CommitPreviewItem.uriPattern}>
+        {({itemHolder, params}) => {
+          return (
+            <CommitPreviewItem
+              ref={itemHolder.setter}
+              workingDirectory={params.workingDirectory}
+              {...props}
+            />
+          );
+        }}
+      </PaneItem>
+    );
+  }
+
+  function open(wrapper, options = {}) {
+    const opts = {
+      workingDirectory: repository.getWorkingDirectoryPath(),
+      ...options,
+    };
+    const uri = CommitPreviewItem.buildURI(opts.workingDirectory);
+    return atomEnv.workspace.open(uri);
+  }
+
+  it('constructs and opens the correct URI', async function() {
+    const wrapper = mount(buildPaneApp());
+    await open(wrapper);
+    assert.isTrue(wrapper.update().find('CommitPreviewItem').exists());
+  });
+});


styles/commit-preview-view.less
```diff
@@ -0,0 +1,20 @@
+@import "variables";
+
+.github-StubItem-git-commit-preview { // TODO Rename class
```
simurai: It seems this class gets auto-generated? Will there be some sort of `CommitPreviewView`?

styles/commit-preview-view.less
```diff
@@ -0,0 +1,20 @@
...
+    & + .github-FilePatchView {
+      margin-top: @component-padding;
+      border-top: 1px solid @base-border-color;
+    }
+  }
+
+
+  // hack hack hack
+  .github-FilePatchView {
+    height: 500px;
```
simurai: I think it's not possible to set the height with CSS alone. The reason is that `<atom-text-editor>`s have scrolling inside them and don't expand to whatever height their content is, like normal `<div>`s.

Do `<atom-text-editor>`s have some sort of "auto-height without scrolling" feature? If not we might have to inline each `<atom-text-editor>` with the current `height` of the content. And update it on every change. We probably also lose the performance boost, because now the native `overflow: auto` does the scrolling and not the editor.

Hmmm.. should we consider a single `<atom-text-editor>`? Probably comes with other problems. :grimacing:

simurai: Ok, nevermind. Turns out there is a `autoHeight` option.

lib/views/file-patch-view.js
```diff
@@ -228,7 +228,8 @@ export default class FilePatchView extends React.Component {
         buffer={this.props.filePatch.getBuffer()}
         lineNumberGutterVisible={false}
         autoWidth={false}
-        autoHeight={false}
+        // TODO only set to true for commit previews, but not for single FilePatchViews
+        autoHeight={true}
```
simurai: Would it be possible to disable `autoHeight` again for "single" `FilePatchView`s? Currently they don't scroll anymore.

smashwilson: Fixed ✅  I've added a prop to `FilePatchView` that controls whether or not `autoHeight` is set and set it appropriately in the parent components.

styles/commit-preview-view.less
smashwilson: That class is actually a temporary one that's applied to "stub items," which are used during deserialization for timing reasons. I've added a `github-CommitPreview-root` class to the `<div>` that should always be there 👌

annthurium:
addressing @simurai's questions:

> Should it be a pending pane? It feels like the Commit Preview is something more temporary. Or do people want to keep it open all the time, maybe in a split view?
@kuychaco brought this up too.  I'm not really sure what the difference is between a pending pane and a non pending pane, tbh.  I'll learn about that and come back with a more informed opinion.

> Similar and maybe depends on the above, should the Commit Preview pane get automatically closed after committing? Maybe not if there are more unstaged changes since you might want to make a 2nd commit next.
Yeah, I think we should not close the preview automatically, for that reason. 

> Should the Commit Preview button be a 3-way toggle? Similar how the Git and GitHub status bar buttons work. If the Commit Preview pane is already open, but not visible (the active tab), users might just want to focus it again, and not close it altogether.
Is your question prompted by performance concerns?  I think the API we're using hides the pane, rather than destroying it.

> Should we remove the blinking cursor? And use the default mouse cursor? To not make it look like the diff is editable? Here some options: #1720 (comment)
It depends on when editable diffs are on the roadmap.  If we're going to do that soon (like before the end of the year), it might not make sense to implement this.  Otherwise yeah, it's confusing if the diff looks like it's editable and it's not.

simurai:
> I'm not really sure what the difference is between a pending pane and a non pending pane

Sorry, here a gif that tries to show the issue:

pending

Notice that when clicking on individual files, Atom automatically closes the tab of the previous file. But after opening the "Commit Preview" and clicking on a single file, Atom opens a new tab for that file and keeps the "Commit Preview" tab permanently opened. It's not necessarily a "bug", some people might prefer this behaviour. Ohh.. btw. "pending pane" is a core setting, that can be turned off/on:

screen shot 2018-11-02 at 11 45 59 am

---

> Is your question prompted by performance concerns?

More a UX concern. Actually, it now works like a 3-way toggle. 👍👍👍 Just that the label of the button "Close Commit Preview" could be misleading. Example:

1. The "Commit preview" tab is open, but not the active tab
2. When clicking on "Close Commit Preview", it doesn't close the "Commit preview" tab, it makes it the active tab. Which is the right behaviour, just wrong button label.

toggle

Maybe "Close Commit Preview" could be renamed to "Toggle Commit Preview". Then the button label doesn't need to be updated when the "Commit Preview" tab isn't the active tab.

---

Re. button label. I asked about it in our design review. Some felt it's confusing that it says "Commit Preview", because it feels more like a "Show all staged changes". Curious if more people feel that way too.

---

> Otherwise yeah, it's confusing if the diff looks like it's editable and it's not.

Ok, I might play around a bit to see if it can be made more "read-only" looking.

---

Anyways, since this PR is our first exploration into multi-file diffs, we don't have to iron out every little detail. :sweat_smile: It might be easier to make certain decisions after using this feature for a while.

smashwilson:
> Re. button label. I asked about it in our design review. Some felt it's confusing that it says "Commit Preview", because it feels more like a "Show all staged changes". Curious if more people feel that way too.

Hmm, interesting. I can see your point about "commit preview" being unclear... but I feel like "show all staged changes" doesn't do enough to convey _why_ you want to do that. It's also a bit wordy.

Maybe its proximity to the commit editor and commit button will be enough to signal the "why"?

kuychaco:
Can someone please double check this off-by-one thing for the "Open in File" logic? See https://github.com/atom/github/pull/1767/commits/3a3c68a08e87255e16e059ab1d56de05ddd41739 (implementation) and https://github.com/atom/github/pull/1767/commits/a60cfc5bdbbd425e8a1588a47dece84f5fec56db (tests)

It's getting late in the day and my brain isn't braining quite as well anymore...

simurai:
Here mockups for 2 alternatives to the "Commit Preview" button:

### 1. "View All" button in the header

> Put the button in the Staged Changes header. “View All” button to the left of “Unstage All”?

button

👉 [In context](https://user-images.githubusercontent.com/378023/48334833-554b8000-e69f-11e8-824b-d680d4d43235.png).

---

### 2. "Preview" button, next to the commit button

Here the alternative that already came up before:

screen shot 2018-11-12 at 5 24 45 pm

👉 [In context](https://user-images.githubusercontent.com/378023/48335140-36012280-e6a0-11e8-98d2-c8e1e00688f0.png)

About the problem that it gets too crowded with the "Abort Merge" button... we could move the "Abort Merge" button above it, since it's not visible permanently:

screen shot 2018-11-12 at 5 29 26 pm

👉 [In context](https://user-images.githubusercontent.com/378023/48335326-b758b500-e6a0-11e8-8f0f-a9d1a632b739.png)

---

Maybe try option 1? Space wise, I think it's still ok. 👍

lib/watch-workspace-item.js
```diff
@@ -0,0 +1,40 @@
...
+
+  function itemMatches(item) {
+    return item.getURI && uPattern.matches(item.getURI()).ok();
+  }
+
+  if (!component.state) {
+    component.state = {};
+  }
+  let itemCount = workspace.getPaneItems().filter(itemMatches).length;
+  component.state[stateKey] = itemCount > 0;
```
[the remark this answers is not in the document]
kuychaco: oops this is an outdated comment that I made long ago. Here's the updated line of code this question corresponds to -- https://github.com/atom/github/pull/1767/files#diff-c4986cbd0e872a78fcf3b9df64cbad97R30

And I think the answer is that we're initializing the state so we're use an object literal. Whereas `setState` is only used to update state once the component mounts...

lib/controllers/multi-file-patch-controller.js
```diff
@@ -183,30 +188,47 @@ export default class FilePatchController extends React.Component {
   }
 
   async discardRows(rowSet, nextSelectionMode, {eventSource} = {}) {
+    // (kuychaco) For now we only support discarding rows for MultiFilePatches that contain a single file patch
```
annthurium: super, duper nit: is there a reason this comment is directed at Katrina?

lib/git-shell-out-strategy.js
```diff
@@ -656,6 +656,23 @@ export default class GitShellOutStrategy {
     return rawDiffs;
   }
 
+  async getStagedChangesPatch() {
+    const output = await this.exec([
+      'diff', '--staged', '--no-prefix', '--no-ext-diff', '--no-renames', '--diff-filter=u',
```
annthurium: another question just for my own education: how did you know what arguments to pass to `diff` here?

test/builder/patch.js
```diff
@@ -0,0 +1,326 @@
...
+    return this.layeredBuffer.wrapReturn({
+      filePatch: new FilePatch(this.oldFile, this.newFile, patch),
+    });
+  }
+}
+
+class FileBuilder {
+  constructor() {
+    this._path = 'file.txt';
+    this._mode = '100644';
```
annthurium: ugh, I can never remember what the file modes mean.  Is this worth documenting with a comment?

lib/containers/changed-file-container.js
```diff
@@ -50,11 +52,13 @@ export default class FilePatchContainer extends React.Component {
...
+
     return (
-      <FilePatchController
-        filePatch={data.filePatch}
-        isPartiallyStaged={data.isPartiallyStaged}
-        hasUndoHistory={data.hasUndoHistory}
+      <MultiFilePatchController
+        multiFilePatch={multiFilePatch}
+        hasUndoHistory={hasUndoHistory}
+        isPartiallyStaged={isPartiallyStaged}
```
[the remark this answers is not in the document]
annthurium: I kinda like listing it out though.  Just to make it clearer what props are getting passed in.  It's easy to pass in unnecessary props when you're spreading.

lib/watch-workspace-item.js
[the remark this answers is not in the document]
smashwilson: > And I think the answer is that we're initializing the state so we're use an object literal. Whereas setState is only used to update state once the component mounts...

:+1: Bingo. You _can't_ use `setState` in a React component's constructor, and you _can't_ use direct assignment elsewhere in the component. The workspace item watcher is awkward in that it's an external class that wants to manipulate the state of a React component, so it needs to be aware of where it's expecting to be called.

This is one place where React's Hooks will be super useful, I think!

lib/models/patch/builder.js
```diff
@@ -5,27 +5,73 @@ import File, {nullFile} from './file';
 import Patch from './patch';
 import {Unchanged, Addition, Deletion, NoNewline} from './region';
 import FilePatch from './file-patch';
+import MultiFilePatch from './multi-file-patch';
 
-export default function buildFilePatch(diffs) {
+export function buildFilePatch(diffs) {
```
smashwilson: Do we still actually need `buildFilePatch()`? I believe `buildMultiFilePatch()` supersedes everything it can do.

I suppose the only benefit to keeping both is that `buildFilePatch()` enforces that we're expecting to receive exactly zero, one, or two diffs?

styles/file-patch-view.less
```diff
@@ -24,14 +23,36 @@
     padding: @component-padding;
   }
 
+  // TODO: Use better selector
```
smashwilson: ☝️ Unresolved TODO 🚨

test/builder/patch.js
smashwilson: 🚨 I would prefer extracting them to constants, personally. Probably in a static property on the File class? I can handle it 👌

test/controllers/commit-controller.test.js
```diff
@@ -29,6 +30,16 @@ describe('CommitController', function() {
     const noop = () => {};
     const store = new UserStore({config});
 
+    // Ensure the Workspace doesn't mangle atom-github://... URIs
```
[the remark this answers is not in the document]
smashwilson: It turns out that if you don't have an opener registered for a nonstandard URI pattern like this, the Workspace tries to interpret it as a file URI, which mangles it and breaks all of the tests.

```js
> atom.workspace.open('atom-github://unknown/whatever').then(item => console.log(item.getURI()))
< atom-github:/unknown/whatever
```

(Notice the single forward slash after `atom-github:/` :point_up:)

🚨 I'll add a comment for it.

test/watch-workspace-item.test.js
```diff
@@ -0,0 +1,203 @@
...
+  describe('setPattern', function() {
+    it('immediately updates the state based on the new pattern', async function() {
+      sub = watchWorkspaceItem(workspace, 'atom-github://item0/{pattern}', component, 'theKey');
+      assert.isFalse(component.state.theKey);
+
+      await workspace.open('atom-github://item1/match');
+      assert.isFalse(component.setState.called);
+
+      await sub.setPattern('atom-github://item1/{pattern}');
+      assert.isFalse(component.state.theKey);
```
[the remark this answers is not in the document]
smashwilson: Nope! `component` isn't a real React component, just a mock one, so the `.state` property and `.setState()` method are (intentionally) unrelated. What I'm testing here is that `setPattern()` is using `component.setState()` to modify the "component's" state with the new pattern instead of directly setting it on `component.state`.

lib/models/repository-states/present.js
```diff
@@ -119,6 +119,8 @@ export default class Present extends State {
       const includes = (...segments) => fullPath.includes(path.join(...segments));
 
       if (filePathEndsWith(fullPath, '.git', 'index')) {
+        keys.add(Keys.stagedChanges);
+        // todo: stagedChangesSinceParentCommit appears to be dead code and can be removed
```
smashwilson: 🚨 Unresolved TODO

lib/views/file-patch-header-view.js
```diff
@@ -40,8 +46,31 @@ export default class FilePatchHeaderView extends React.Component {
...
+  }
+
+  renderPath() {
+    const dirname = path.dirname(this.props.relPath);
+    const basename = path.basename(this.props.relPath);
+
+    if (dirname === '.') {
+      return <span className="gitub-FilePatchHeaderView-basename">{basename}</span>;
+    } else {
+      // TODO: double check that the forward-slash works cross-platform...
```
smashwilson: 🚨 Unresolved TODO

lib/git-shell-out-strategy.js
annthurium: great, thanks! that's really helpful.

lib/models/repository-states/present.js
annthurium: this is some unrelated dead code we discovered, but due to the possibility of introducing additional bugs I say we punt on this unrelated change and clean it up later.

lib/controllers/multi-file-patch-controller.js
kuychaco: This is true ☝️. Maybe there's a more clear format for this?

lib/models/repository-states/present.js
kuychaco: Based on my knowledge of how this data was used, and looking at where/how it's still being referenced in the codebase, I'm confident that it's safe to clean it up without fear of introducing bugs. I'll go ahead and do so.

lib/containers/changed-file-container.js
```diff
@@ -5,9 +5,9 @@ import yubikiri from 'yubikiri';
 import {autobind} from '../helpers';
 import ObserveModel from '../views/observe-model';
 import LoadingView from '../views/loading-view';
-import FilePatchController from '../controllers/file-patch-controller';
+import MultiFilePatchController from '../controllers/multi-file-patch-controller';
 
-export default class FilePatchContainer extends React.Component {
+export default class ChangedFileContainer extends React.Component {
```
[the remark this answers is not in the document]
kuychaco: I'm glad! Tbh I wasn't sure about how others would think about the naming when I first did the re-structuring, so your validation helps me close out that concern :)

lib/models/patch/builder.js
kuychaco: Maybe? I'm fine just using `buildMultiFilePatch()`. From what I can tell we just need to handle the `emptyDiffFilePatch` case in it. And we'd need to consolidate the tests and see about which `buildFilePatch` tests need to be converted to `buildMultiFilePatch` tests. Makes the code cleaner and potentially reduces friction to change in the future. I'm 👍 if you wanna do this. I'm also fine leaving it for now and opening up a code cleanup issue for it

lib/views/multi-file-patch-view.js
```diff
@@ -304,13 +303,47 @@ export default class FilePatchView extends React.Component {
...
+
+                tooltips={this.props.tooltips}
+
+                undoLastDiscard={() => this.undoLastDiscardFromButton(filePatch)}
+                diveIntoMirrorPatch={() => this.props.diveIntoMirrorPatch(filePatch)}
+                openFile={() => this.didOpenFile({selectedFilePatch: filePatch})}
+                toggleFile={() => this.props.toggleFile(filePatch)}
+              />
+              {this.renderExecutableModeChangeMeta(filePatch)}
+              {this.renderSymlinkChangeMeta(filePatch)}
```
annthurium: @smashwilson: I moved this function call from line 312 to here, and now the integration tests in `file-patch.test.js` are failing.  😱 

The tests pass when I move the rendering back to the way it was, so I'm confident that this is the problem, but I don't understand why.

Katrina and I tried to figure it out today and failed.  Would you be willing to take a look?

lib/views/file-patch-header-view.js
smashwilson: I'm glad I checked this! It turns out that while git itself always uses forward slashes, we explicitly translate paths in patches with `toNativePathSep`:

whoops

Fortunately the fix is to just use `path.sep` instead:

new-and-improved

lib/controllers/multi-file-patch-controller.js
```diff
@@ -22,9 +22,10 @@ export default class FilePatchController extends React.Component {
     config: PropTypes.object.isRequired,
 
     destroy: PropTypes.func.isRequired,
-    discardLines: PropTypes.func.isRequired,
-    undoLastDiscard: PropTypes.func.isRequired,
-    surfaceFileAtPath: PropTypes.func.isRequired,
+    discardLines: PropTypes.func,
+    undoLastDiscard: PropTypes.func,
+    surfaceFileAtPath: PropTypes.func,
+    handleClick: PropTypes.func,
```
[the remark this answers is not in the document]
kuychaco: I :fire:'d it. Hope that's okay!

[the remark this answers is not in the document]
smashwilson: I think it was a relic of the multi-AtomTextEditor approach so we should be good 👍

lib/views/commit-view.js
```diff
@@ -223,7 +236,7 @@ export default class CommitView extends React.Component {
 
           <button
             ref={this.refCommitButton.setter}
-            className="github-CommitView-button github-CommitView-commit btn btn-primary"
+            className="github-CommitView-button github-CommitView-commit btn btn-primary native-key-bindings"
```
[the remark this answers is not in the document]
smashwilson: Without `native-key-bindings`, the button didn't click on "enter" when it was focused which felt weird. (I think spacebar still worked, though.)

lib/controllers/root-controller.js
```diff
@@ -128,6 +129,7 @@ export default class RootController extends React.Component {
     return (
       <Commands registry={this.props.commandRegistry} target="atom-workspace">
         {devMode && <Command command="github:install-react-dev-tools" callback={this.installReactDevTools} />}
+        <Command command="github:toggle-commit-preview" callback={this.toggleCommitPreviewItem} />
```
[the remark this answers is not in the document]
annthurium: I think it would be weird to title a pane "View All Staged Changes" or "All Staged Changes."  It's ok for the button and the pane to be labeled differently imo.

Changing the name of the command to match the text that's on the button makes a lot of sense though.  I can take care of that.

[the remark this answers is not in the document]
annthurium: hmm actually...

"toggle all staged changes" seems less clear than "toggle commit preview."  As a user I'd worry that "toggle all staged changes" is going to unstage the things I have staged, or something like that, rather than just opening a pane.  I vote to leave this as is.

kuychaco:
Found another bug. It appears when the Commit Preview has a mode-change patch with no hunks, and then another patch below it with changed lines:

commit_preview_ ___src_repo-symlinks

Unstaging lines throws an error:

```
git-shell-out-strategy.js? [sm]:33 Uncaught (in promise) GitError: git apply --cached - in /Users/kuychaco/src/test-repo exited with code 128
stdout: 
stderr: error: patch with only garbage at line 4

    at new GitError (/Users/kuychaco/github/github/lib/git-shell-out-strategy.js:35:18)
    at Promise (/Users/kuychaco/github/github/lib/git-shell-out-strategy.js:329:23)
    at <anonymous>
```

Here's the diagnostic output:

commit_preview_ ___src_repo-symlinks

docs/react-component-atlas.md
```diff
@@ -66,12 +66,24 @@ This is a high-level overview of the structure of the React component tree that
...
+> > > [`<MultiFilePatchView>`](/lib/views/multi-file-patch-view.js)
+> > >
+> > > Render a sequence of git-generated file patches within a TextEditor, using decorations to include contextually
+> > > relevant controls.
+>
+> > [`<CommitPreviewItem>`](/lig/items/commit-preview-item.js)
+> > [`<CommitPreviewContainer>`](/lib/containers/commit-preview-container.js)
+> >
+> > The workspace-center pane that appears when looking at _all_ the staged changes that will be going into the next commit.
+> >
```
PaRaD0XCORP: dfdf

cristianomarquesadmo1:
Good job guys. Congratulacions to all

PR #1797, opened 2018-11-14
Prevent broken patches when staging or unstaging lines rapidly

### LINKED ISSUE #1796: Bad patches when staging rapidly
### Prerequisites

* Have you tried to reproduce the problem in [Safe Mode][safe-mode]?
* Have you followed all applicable steps in the [debugging guide][debugging-guide]?
* Have you checked the FAQs on the [message board][message-board] for common solutions?
* Have you checked that your issue [isn't already filed][already-filed]?

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

### Description

When rapidly staging or unstaging files one line at a time, a "patch does not apply" error appears in the dev console.

### Steps to Reproduce

1. Create an untracked file with many lines.
2. Click on the file in the unstaged changes list to open it in a `ChangedFileItem`.
3. Select the first row.
4. Use the keyboard shortcut (`cmd-enter`) to rapidly stage many lines at once.

**Expected behavior:**

Rapid stage (or unstage) operations should be throttled into a synchronous sequence that applies cleanly.

**Actual behavior:**

The following stacktrace appears in the dev console:

```
Uncaught (in promise) GitError: git apply --cached - in /Users/smashwilson/fixtures/patches exited with code 1
stdout: 
stderr: error: patch failed: few-patches.after.txt:1
error: few-patches.after.txt: patch does not apply

    at new GitError (/Users/smashwilson/.atom/packages/github/lib/git-shell-out-strategy.js:34:18)
    at Promise (/Users/smashwilson/.atom/packages/github/lib/git-shell-out-strategy.js:328:23)
    at <anonymous>
```

**Reproduces how often:**

Reproduces consistently with files containing ~100 different lines.

### Versions

```
Atom    : 1.34.0-nightly12
Electron: 2.0.12
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

```
apm  2.1.3
npm  6.2.0
node 8.9.3 x64
atom 1.34.0-nightly12
python 2.7.15
git 2.19.1
```

```
github@0.22.1-0 (atom/github#33b9063a)
```

First seen by @vanessayuenn.

smashwilson:
### Requirements

### Description of the Change

To prevent the generation of broken patches, the `FilePatchController` (soon to be `MultiFilePatchController`) pauses staging operations while it waits on a Promise that is _intended_ to be resolved when the `FilePatch` containing the result of the operation arrives as a component prop. However, staging operations were being resumed too soon for two reasons:

* If a new `FilePatch` is already being generated before a staging operation is begun -- triggered by an unrelated filesystem event, for example -- and arrives before the staging operation is complete, the "next patch" promise will resolve too early.

I've addressed this somewhat by waiting for the staging operation promise to resolve first before waiting for the "next patch" promise to resolve.

* Our `FilePatch` comparison noticed any difference in referential equality between old and new patches. This isn't good enough: it's very possible for the component to be re-rendered with a different `FilePatch` object that contains an identical patch.

This one I fixed by comparing the actual `toString()` serialization of the old and new `FilePatches` to determine when to resolve the "next patch" promise. I am still doing a referential equality check first so we'll stay nice and quick in the common case of a re-render with the same patch, and not doing a check at all in the common case of not having a pending staging operation underway.

### Alternate Designs

I thought about introducing a first-class way to sequence the operations performed within a `Repository` so we could explicitly express our constraint of "wait for the next FilePatch to be generated _after_ the staging operation has completed." We could do this by tracking a counter in the `Present` state that we increment each time an operation is started, and finding some way to compare its value among two different result objects. It felt pretty invasive to implement once I started looking at it, though, so I took the simpler tack.

### Benefits

Rapidly staging and unstaging lines will no longer create bad patches.

### Possible Drawbacks

Well, it might cause some merge conflicts in #1767.

### Applicable Issues

Fixes #1796.

### Metrics

_N/A_

### Tests

I've tested this by manually reproducing the bug in dev mode, making the change, and verifying that it no longer appears.

We already do have unit tests that are intended to capture this sort of situation. It turns out that writing unit tests to catch asynchronous race conditions is tricky. Who knew!

### Documentation

_N/A_

### User Experience Research

- [ ] @vanessayuenn should be unable to reproduce the problem when using this branch

### DIFF
--- a/lib/controllers/file-patch-controller.js
+++ b/lib/controllers/file-patch-controller.js
@@ -37,7 +37,6 @@ export default class FilePatchController extends React.Component {
     );
 
     this.state = {
-      lastFilePatch: this.props.filePatch,
       selectionMode: 'hunk',
       selectedRows: new Set(),
     };
--- a/lib/controllers/file-patch-controller.js
+++ b/lib/controllers/file-patch-controller.js
@@ -44,13 +44,17 @@ export default class FilePatchController extends React.Component {
     this.mouseSelectionInProgress = false;
     this.stagingOperationInProgress = false;
 
+    this.lastPatchString = null;
     this.patchChangePromise = new Promise(resolve => {
       this.resolvePatchChangePromise = resolve;
     });
   }
 
   componentDidUpdate(prevProps) {
-    if (prevProps.filePatch !== this.props.filePatch) {
+    if (
+      this.lastPatchString !== null &&
+      this.lastPatchString !== this.props.filePatch.toString()
+    ) {
       this.resolvePatchChangePromise();
       this.patchChangePromise = new Promise(resolve => {
         this.resolvePatchChangePromise = resolve;
@@ -219,11 +223,18 @@ export default class FilePatchController extends React.Component {
     if (this.stagingOperationInProgress) {
       return null;
     }
-    this.stagingOperationInProgress = true;
-    this.patchChangePromise.then(() => {
-      this.stagingOperationInProgress = false;
-    });
 
-    return fn();
+    this.stagingOperationInProgress = true;
+    this.lastPatchString = this.props.filePatch.toString();
+    const operationPromise = fn();
+
+    operationPromise
+      .then(() => this.patchChangePromise)
+      .then(() => {
+        this.stagingOperationInProgress = false;
+        this.lastPatchString = null;
+      });
+
+    return operationPromise;
   }
 }
--- a/test/controllers/file-patch-controller.test.js
+++ b/test/controllers/file-patch-controller.test.js
@@ -4,6 +4,7 @@ import React from 'react';
 import {shallow} from 'enzyme';
 
 import FilePatchController from '../../lib/controllers/file-patch-controller';
+import FilePatch from '../../lib/models/patch/file-patch';
 import * as reporterProxy from '../../lib/reporter-proxy';
 import {cloneRepository, buildRepository} from '../helpers';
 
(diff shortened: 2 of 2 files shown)


lib/controllers/file-patch-controller.js
```diff
@@ -37,21 +37,24 @@ export default class FilePatchController extends React.Component {
...
     this.state = {
-      lastFilePatch: this.props.filePatch,
       selectionMode: 'hunk',
       selectedRows: new Set(),
     };
 
     this.mouseSelectionInProgress = false;
     this.stagingOperationInProgress = false;
 
+    this.lastPatchString = null;
```
vanessayuenn: What is the benefit of storing `lastPatchString` as a variable vs keeping it in the state?

smashwilson: The state key was actually unused entirely. There were no references to it; the patch change promise was being resolved with a comparison between `this.props` and the `prevProps` given to `componentDidUpdate()`.

As for storing it as a string though - as part of the process of rendering a new patch, we "adopt" the TextBuffer from the previous one, which leaves the buffer on the old patch empty, so _every_ patch was being detected as "different". Pre-rendering the old one to a string lets us compare them even after we've messed around with it.

Issue #1715, opened 2018-10-03
Split IssueishDetailItem into PullRequestDetailItem and IssueDetailItem


smashwilson:
A bit of refactoring @kuychaco suggested in #1712. The two share _some_ functionality but (a) they will soon diverge more and (b) even now there are a bunch of `isPr` conditionals that are pretty ugly.

Issue #1611, opened 2018-07-26
Replace getDerivedStateFromProps with memoization


smashwilson:
Where possible, it would be cleaner and less error-prone to replace components that implement `static getDerivedStateFromProps()` with PureComponents or `memoize-one`:

* ["You Probably Don't Need Derived State" - Memoization](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization)

PR #1804, opened 2018-11-20
split IssueishDetailView into IssueDetailView and PullRequestDetailView

annthurium:
addresses https://github.com/atom/github/issues/1715. 

These components have enough separate functionality that we should move them into their own files.  There should be no user facing changes as a result of this refactor - it just makes it easier for us to maintain things in the future. 

## Todo
- [x] make new files for the new components
- [x] extract pr functionality from `IssueishDetailView` into `PullRequestDetailView`
- [x] extract issue functionality from `IssueishDetailView` into `IssueDetailView`
- [x] fix the graphql problems that will undoubtedly arise
- [x] extract emoji rendering functionality since it's shared by both components
- [x] do a pass on variables that could maybe be renamed?
- [x] nuke `IssueishDetailContainer` and any other dead code
- [x] fix missing prop warning in pr view when commits tab is clicked
- [ ] all unit tests are green
- [ ]  Coveralls reports no new uncovered lines
- [ ] Coveralls reports no uncovered existing lines

## Manual test plan:
- [x] open issue in Atom
    - [x] header links to issue on dotcom
    - [x] refresh button pulls down new data
    - [x] issue body text and comments are styled as expected
    - [x] link to dotcom in issue footer exists
    - [x] emoji reactions are rendered
    - [x] switching from an issue to a pull request (by clicking link to a pull request referenced in an issue)
- [x] open pull request in Atom
  - [x] header links to pull request on dotcom
  - [x] refresh button pulls down new data
  - [x] pull request meta view is styled as expected
  - [x] checkout pull request functionality works.
  - [x] overview tab contains pull request timeline
  - [x] emoji reactions are rendered
  - [x] build status contains the build status
  - [x] commits tab contains commits 
  - [x] test a pull request with more than 50 commits to make sure pagination and refetching work correctly
   - [x] switching from an pull request to an issue (by clicking link to an issue referenced in a pull request body)


lib/views/issue-detail-view.js
```diff
@@ -0,0 +1,246 @@
...
+          switchToIssueish={this.props.switchToIssueish}
+        />
+      </Fragment>
+    );
+  }
+
+  render() {
+    const repo = this.props.repository;
+    const issueish = this.props.issueish;
+    const childProps = {
```
annthurium: I think some child components that are used by both issues and pull requests want to know which is which, and we're not splitting all the components.  So I'm going to leave this as is, but double check that it's truly required by these children.

lib/controllers/issueish-detail-controller.js
```diff
@@ -213,7 +224,12 @@ export default createFragmentContainer(BareIssueishDetailController, {
       commitCount: {type: "Int!"},
       commitCursor: {type: "String"},
     ) {
-      ...issueishDetailView_repository
+      ...issueDetailView_repository
```
annthurium: todo: figure out a way to not duplicate this fragment, because it seems like there should be a way to avoid it.

lib/views/issue-detail-view.js
```diff
@@ -0,0 +1,246 @@
...
+    }),
+    issueish: PropTypes.shape({
+      __typename: PropTypes.string.isRequired,
+      id: PropTypes.string.isRequired,
+      title: PropTypes.string,
+      url: PropTypes.string.isRequired,
+      bodyHTML: PropTypes.string,
+      number: PropTypes.number,
+      state: PropTypes.oneOf([
+        'OPEN', 'CLOSED',
```
annthurium: todo: double check that these are the only states an issue can be in.

annthurium:
@kuychaco, @vanessayuenn, @smashwilson -- I'd love a code review on this when you have some time.

Specifically:

- is there anything weird about the graphql / relay code here?  I want to fix the weird missing prop error message that happens when you click on the commits tab in the pull request view.  I could make the props optional, but that feels lazy and I'd like to fix the root cause if possible since it's clearly something I introduced.
- I left a comment about this, but I have a repeated fragment in the `IssueishDetailController`.  What's the right way to avoid doing that?  
- Do you think it's out of scope to rename a bunch of props we're calling `isssueish` to be an issue or a pull request more specifically?  We still have a lot of code that's shared between both, so I wasn't sure what the best way to go here might be.
- Did I miss any dead code that needs to be 🔥 ?

Thanks!!  Open to any other feedback you have as well, of course.

lib/views/issue-detail-view.js
```diff
@@ -0,0 +1,243 @@
...
+                  icon="repo-sync"
+                  className={cx('github-IssueishDetailView-headerRefreshButton', {refreshing: this.state.refreshing})}
+                  onClick={this.handleRefreshClick}
+                />
+              </div>
+            </div>
+            <div className="github-IssueishDetailView-headerRow">
+              <div className="github-IssueishDetailView-headerGroup">
+                <IssueishBadge className="github-IssueishDetailView-headerBadge"
+                  type={issueish.__typename}
```
kuychaco: We can now just specify that it's `Issue`

kuychaco:
Oooh, we should also update the react component hierarchy doc 📝

lib/controllers/issueish-detail-controller.js
annthurium: Interesting.  This is what I had originally tried but I got a relay compiler error. But now it's working.

lib/views/pr-detail-view.js
```diff
@@ -167,7 +142,7 @@ export class BareIssueishDetailView extends React.Component {
             html={issueish.bodyHTML || '<em>No description provided.</em>'}
             switchToIssueish={this.props.switchToIssueish}
           />
-          {this.renderEmojiReactions(issueish)}
+          <EmojiReactionsView reactionGroups={issueish.reactionGroups} />
           <PrTimelineContainer
```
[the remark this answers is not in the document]
annthurium: I'd prefer to keep this small in scope and address additional refactoring in the future if that's ok.  It's too easy for refactoring to sprawl.

lib/controllers/issueish-detail-controller.js
```diff
@@ -56,44 +58,65 @@ export class BareIssueishDetailController extends React.Component {
     this.updateTitle();
   }
 
+  componentWillMount() {
```
smashwilson: 🚨 `componentWillMount` is deprecated and unsafe in new React code. Have a look at [the React docs](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount) for an explanation.

There are alternative ways to derive state from props. Helpfully the React folks have [a blog post](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html) with some of the (modern) approaches to it. Elsewhere in the codebase right now we tend to use the "lengthy and bug-prone" approach with `static getDerivedStateFromProps`, but I have an open issue (#1611) to go through and upgrade it to one of their recommended alternatives.

lib/controllers/issueish-detail-controller.js
```diff
@@ -15,7 +16,8 @@ export class BareIssueishDetailController extends React.Component {
       owner: PropTypes.shape({
         login: PropTypes.string.isRequired,
       }).isRequired,
-      issueish: PropTypes.any, // FIXME from IssueishPaneItemContainer.propTypes
+      pullRequest: PropTypes.any, // FIXME
+      issue: PropTypes.any, // FIXME from IssueishPaneItemContainer.propTypes
```
smashwilson: We can probably just delete these `FIXME` comments honestly 🙈 I don't think it's ever going to be worth it to fully spec out the Relay responses in PropTypes. My fingers hurt even thinking about it.

lib/controllers/issueish-detail-controller.js
```diff
@@ -56,44 +58,65 @@ export class BareIssueishDetailController extends React.Component {
     this.updateTitle();
   }
 
+  componentWillMount() {
+    const {repository} = this.props;
+    if (repository && repository.pullRequest && repository.pullRequest.__typename) {
+      this.setState({typename: repository.pullRequest.__typename});
+    }
```
smashwilson: Do we ever set `{typename}` to "issue"? From the diff I'm seeing, it looks like it'll either be `'PullRequest'` or `null`.

That'll work because all of the `if(typename)` clauses compare to `'PullRequest'` and have the issue case in the `else` block. It might be confusing if you saw that in the React dev tools, though.

lib/controllers/issueish-detail-controller.js
```diff
@@ -130,13 +153,13 @@ export class BareIssueishDetailController extends React.Component {
     const fromPullRefspec =
       headRemote.getOwner() === repository.owner.login &&
       headRemote.getRepo() === repository.name &&
-      headPush.getShortRemoteRef() === `pull/${issueish.number}/head`;
+      headPush.getShortRemoteRef() === `pull/${pullRequest.number}/head`;
```
smashwilson: Okay, yeah, and this is guaranteed to be defined because we gate on `this.state.typename` up above. 👍

lib/views/emoji-reactions-view.js
```diff
@@ -0,0 +1,40 @@
...
+const reactionTypeToEmoji = {
+  THUMBS_UP: '👍',
+  THUMBS_DOWN: '👎',
+  LAUGH: '😆',
+  HOORAY: '🎉',
+  CONFUSED: '😕',
+  HEART: '❤️',
+};
+
+export default class EmojiReactionsView extends React.Component {
```
smashwilson: Ahhh you extracted this from `IssueishDetailView`? Yesss 💯

lib/views/issue-detail-view.js
```diff
@@ -0,0 +1,239 @@
...
+        <GithubDotcomMarkdown
+          html={issue.bodyHTML || '<em>No description provided.</em>'}
+          switchToIssueish={this.props.switchToIssueish}
+        />
+        <EmojiReactionsView reactionGroups={issue.reactionGroups} />
+        <IssueTimelineController
+          issue={issue}
+          switchToIssueish={this.props.switchToIssueish}
+        />
+      </Fragment>
```
smashwilson: So clean

so clean

lib/views/pr-detail-view.js
```diff
@@ -89,13 +80,13 @@ export class BareIssueishDetailView extends React.Component {
 
   constructor(props) {
     super(props);
-    autobind(this, 'handleRefreshClick', 'refresh', 'renderEmojiReactions', 'renderIssueBody', 'renderPullRequestBody');
+    autobind(this, 'handleRefreshClick', 'refresh', 'renderPullRequestBody', 'recordOpenInBrowserEvent');
```
smashwilson: FYI I've been trying to convert `autobind` calls to the class property style that @kuychaco advocated in new code. So instead of:

```js
class SomeComponent extends React.Component {
  constructor(props) {
    super(props);
    autobind(this, 'boundMethod');
  }

  boundMethod() {
    console.log('You did the thing');
  }
}
```

It would be:

```js
class SomeComponent extends React.Component {
  boundMethod = () => {
    console.log('You did the thing');
  }
}
```

Turns out it apparently does the exact same thing, but requires one less import, can get rid of the need for a constructor entirely, and causes less duplication. (I still kind of miss our `@autobind` decoration, but I like having working code coverage even better). The only downside I've seen is that you can't use `cmd-R` to navigate to the definition.

(You don't _have_ to bother changing these here - this is just a general note for the future :smile:)

lib/views/pr-detail-view.js
```diff
@@ -240,32 +193,32 @@ export class BareIssueishDetailView extends React.Component {
                 />
               </div>
             </div>
-            {isPr && this.renderPrMetadata(issueish, repo)}
+            {this.renderPrMetadata(pullRequest, repo)}
```
smashwilson: Getting rid of all of this conditional rendering is great.

lib/views/pr-detail-view.js
```diff
@@ -317,7 +270,7 @@ export class BareIssueishDetailView extends React.Component {
   }
 
   recordOpenInBrowserEvent() {
-    addEvent('open-issueish-in-browser', {package: 'github', from: 'issueish-header'});
+    addEvent('open-pull-request-in-browser', {package: 'github', component: this.constructor.name});
```
smashwilson: Oh, cool, we get better granularity on these telemetry events, too.

Do we need to update anything or let @telliott27 know that this event name is changing?

test/fixtures/props/issueish-pane-props.js
```diff
@@ -34,9 +34,11 @@ export function issueishDetailControllerProps(opts, overrides = {}) {
     repositoryName: 'repository',
     ownerLogin: 'owner',
 
-    omitIssueish: false,
+    omitPullRequestData: false,
+    omitIssueData: false,
```
smashwilson: Speaking as the one who put them there, I don't love these factories, btw 💔. I'm not sure what a decent alternative to constructing these complicated prop structures that Relay gives us is, though. More builders instead... ?

docs/react-component-atlas.md
```diff
@@ -88,7 +88,8 @@ This is a high-level overview of the structure of the React component tree that
 > > [`<IssueishDetailItem>`](/lib/items/issueish-detail-item.js)
 > > [`<IssueishDetailContainer>`](/lib/containers/issueish-detail-container.js)
 > > [`<IssueishDetailController>`](/lib/controllers/issueish-detail-controller.js)
-> > [`<IssueishDetailView>`](/lib/controllers/issueish-detail-controller.js)
+> > [`<IssueDetailView>`](/lib/views/issue-detail-controller.js)
```
annthurium: oh darn it.  These filepaths are wrong.  I'll fix 'em.

lib/views/pr-detail-view.js
annthurium: gooood question.  I was just thinking about that during our meeting this morning.  

@telliott27 : will this break the existing dashboard queries?

lib/controllers/issueish-detail-controller.js
annthurium: soooo, I think that we fetch both the `Issue` and the `PullRequest` every time, and the `PullRequest` has a `__typename` of `Issue` if it's an issue?  I should verify that, though. 

What I really wanted was to pull `__typename` up a level, so that it belongs to `repository`, but I couldn't figure out how to express that in Relay.

annthurium: just verified that my initial comment is correct -- if we're rendering an issue, we have a `pullRequest` with the `__typename` of `Issue`, and also an `issue` with the `__typename` of `Issue`.

I can see that potentially being confusing or unexpected, though.  I'll try again to pull the `__typename` up a level in the graphql so it's a property of the `repository`.

smashwilson: Ohhh. Oh, I see now.

> I'll try again to pull the `__typename` up a level in the graphql so it's a property of the repository.

Hmm I don't know if you can do that or not. Maybe we could just add a comment here?

test/fixtures/props/issueish-pane-props.js
annthurium: yeah, this one is especially bad because to render an issue to test, you still need pull request data with a typename of `Issue`.  I also couldn't think of a better way to handle it, though.

lib/controllers/issueish-detail-controller.js
annthurium: thanks for catching this!  That's really important.

I ended up using `getDerivedStateFromProps` because we're only memoizing to avoid gnarly nested lookups, not because we're trying to avoid recomputing something expensive so `memoize-one` didn't really make sense here.

lib/controllers/issueish-detail-controller.js
kuychaco: I think we can actually get away with not using state entirely. According to [best practices](https://github.com/uberVU/react-guide/blob/master/props-vs-state.md#does-this-go-inside-props-or-state), we want to reserve our usage of state for attributes that the component will mutate at some point. For attributes that are received from above and are immutable we want to use props. In this case the typename will never be modified by the component.

I think this also increases readability and precision in our logic...

```
    if (this.state.typename !== 'PullRequest') {
      return this.checkoutOp.disable(checkoutStates.HIDDEN, 'Cannot check out an issue');
    }
```

would instead be 

```
    if (issueish.__typename === 'Issue') {
      return this.checkoutOp.disable(checkoutStates.HIDDEN, 'Cannot check out an issue');
    }
```

Additionally, there is some tweaking to our graphql fragment that will make this logic cleaner. I'll comment on that separately.

PR #1822, opened 2018-12-04
Fix: git tab keeps popping open when there's a merge conflict

### LINKED ISSUE #1816: Merge conflict pane keeps opening itself
### Description

There's a merge conflict in my branch. Atom desperately wants me to use the merge conlfict panel, and keeps re-opening it after I close the panel.

### Steps to Reproduce

1. Have a merge conflict
2. Open some files 
3. Close the merge panel
4. Save some files

**Expected behavior:**

Merge panel stays closed.

**Actual behavior:**

Merge panel re-opens itself whenever I save a file. This constantly gets in the way.

**Reproduces how often:**

100%

### Versions

Atom 1.33.0 x 64
Mac 10.13.6

### Additional Information

I prefer to keep Atom focused on code, and I need the screen space that panel keeps taking to see code.

I use Tower App for taking Git actions.

vanessayuenn:
### Description of the Change

This PR fixes the bug where the git tab keeps opening itself when a merge conflict is detected and remains unfixed. Instead of doing that, I moved the error signaling up one level, so now _all_ git errors resulted from a `pull` action will trigger an error pop up, and git tab will open itself (one time only). 

- [x] I think this might be an opportunity to increase test coverage for `get-repo-pipeline-manager.js`

### Alternate Designs

N/A

### Benefits

Users won't get annoyed at git tab opening itself anymore. People don't like it when robots act smarter than them, y'know. 

### Applicable Issues

Resolves #1816 

### Metrics

N/A

### Tests

See above


lib/get-repo-pipeline-manager.js
```diff
@@ -72,6 +72,7 @@ export default function({confirm, notificationManager, workspace}) {
       return result;
     } catch (error) {
       if (error instanceof GitError) {
+        repository.didPullError();
```
smashwilson: What's the reason for generalizing this from specifically merge errors to any pull errors?

test/get-repo-pipeline-manager.test.js
```diff
@@ -0,0 +1,245 @@
+import {cloneRepository, buildRepositoryWithPipeline} from './helpers';
+import {GitError} from '../lib/git-shell-out-strategy';
+
+
+describe('getRepoPipelineManager()', function() {
```
smashwilson: Thank you for doing this 🙏 

Go go gadget test coverage 📈

lib/controllers/root-controller.js
```diff
@@ -453,7 +452,7 @@ export default class RootController extends React.Component {
   componentDidUpdate() {
     this.subscription.dispose();
     this.subscription = new CompositeDisposable(
-      this.props.repository.onMergeError(() => this.gitTabTracker.ensureVisible()),
+      this.props.repository.onPullError(() => this.gitTabTracker.ensureVisible()),
```
smashwilson: 👍

smashwilson:
:shipit: :shipit: :shipit:

vanessayuenn:
@smashwilson when I was testing it, I have found the merge error check we have does not capture all cases where merge conflicts arise (e.g. in my case it's because I have set my `pull` to use rebase automatically). And looking through the other types of pull errors, it makes sense to have the git tab pulled open to address those errors. If this gets too much, I am totally down to revert the behavior back to only be triggered by merge errors.

Issue #1776, opened 2018-11-03
Error when trying to view a pull request


Chewsterchew:
### Prerequisites

- [x] Have you tried to reproduce the problem in [Safe Mode][safe-mode]?
- [x] Have you followed all applicable steps in the [debugging guide][debugging-guide]?
- [x] Have you checked the FAQs on the [message board][message-board] for common solutions?
- [x] Have you checked that your issue [isn't already filed][already-filed]?

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

I even like asked on slack and they said come here.

### Description

I went to checkout a pull request in my repository, however when I clicked on it, it failed to load.

### Steps to Reproduce

*These probably won't work, it's a unique situation*

1. Clone an repository with one open PR and open the project in atom.
2. Open GH tab
3. Click a pull request?? Not sure.

**Expected behavior:**

It to view the pull request as normal.

**Actual behavior:**

Both the github tab and main pane turn black and do nothing until I restart Atom.

**Reproduces how often:**

Only on one specific pull request. It did another from the same repository perfectly fine.

### Versions

Versions? check it out
```
apm  2.1.2
npm  6.2.0
node 8.9.3 x64
atom 1.32.1
python 3.6.4
git 2.18.0
```

### Additional Information

Here's a fun error I got.

```
<embedded>:14 TypeError: Cannot read property 'owner' of null
    at BareIssueishDetailController.nextCheckoutOp (/Applications/Atom.app/Contents/Resources/app/static/<embedded>:14:3679469)
    at BareIssueishDetailController.render (/Applications/Atom.app/Contents/Resources/app/static/<embedded>:14:3678190)
    at Gg (/Applications/Atom.app/Contents/Resources/app/static/<embedded>:14:2438932)
    at Kg (/Applications/Atom.app/Contents/Resources/app/static/<embedded>:14:2443812)
    at th (/Applications/Atom.app/Contents/Resources/app/static/<embedded>:14:2453366)
    at uh (/Applications/Atom.app/Contents/Resources/app/static/<embedded>:14:2453673)
    at Rh (/Applications/Atom.app/Contents/Resources/app/static/<embedded>:14:2457250)
[...]

PR #1828, opened 2018-12-06
Fix bug where pull request cannot be viewed if pullRequest.headRepository is null.

annthurium:
### Description of the Change

If `pullRequest.headRepository` is null, (which might be the case if, for example, a user deletes the repository), currently we throw an error and the pull request cannot be viewed.

This pull request adds a check for this condition, so we can avoid the code that's checking for a non existent property.

### Alternate Designs

This approach allows a user to view a pull request where a head repository is null, but they cannot check it out.  While it would be possible to fall back to refs/pull/${PR}/head to check out the pull request from dotcom, it seems not worth the added complexity.  My assumption is, it's relatively rare for a user to open a cross repository pull request and then delete their fork. 

If I'm wrong, and it's more common than I think, I'd be open to revisiting this decision and supporting checkout in this edge case.

### Benefits

Users can view pull requests again.

### Possible Drawbacks

I can't think of anything, other than writing code always introduces the possibility of more bugs.

### Applicable Issues

https://github.com/atom/github/issues/1776

### Metrics

n/a

### Tests

- Wrote a unit test to validate that the checkout opportunity is disabled if `pullRequest.headRepository` is null.
- Manually testing viewing a pull request within Atom where the head repository has been deleted.

### Documentation

N/A

### Release Notes

It's not a notable change, dawg.  Nothing to see here.

### DIFF
--- a/test/controllers/issueish-detail-controller.test.js
+++ b/test/controllers/issueish-detail-controller.test.js
@@ -109,6 +109,15 @@ describe('IssueishDetailController', function() {
       assert.isFalse(op1.isEnabled());
       assert.strictEqual(op1.getMessage(), 'Rebase in progress');
     });
+    it('is disabled if pullRequest.headRepository is null', function() {
+      const props = issueishDetailControllerProps({}, {});
+      props.repository.pullRequest.headRepository = null;
+      const wrapper = shallow(buildApp({}, {...props}));
+      const op = wrapper.find('Relay(BarePullRequestDetailView)').prop('checkoutOp');
+      assert.isFalse(op.isEnabled());
+      assert.strictEqual(op.getMessage(), 'Pull request head repository does not exist');
+    });
+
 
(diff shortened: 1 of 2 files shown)


test/controllers/issueish-detail-controller.test.js
```diff
@@ -109,6 +109,15 @@ describe('IssueishDetailController', function() {
       assert.isFalse(op1.isEnabled());
       assert.strictEqual(op1.getMessage(), 'Rebase in progress');
     });
+    it('is disabled if pullRequest.headRepository is null', function() {
+      const props = issueishDetailControllerProps({}, {});
+      props.repository.pullRequest.headRepository = null;
```
annthurium: I...I don't love this, but also, the way the test factories are currently set up, you can't supply a parameter to our test factories that makes an entire nested part of the Relay response null.   We should probably come up with a better pattern for these factories.

PR #1832, opened 2018-12-09
React 16.6.3

smashwilson:
Let's see if this magically works somehow. Or: let's see if our snapshot build catches that it still doesn't.


smashwilson:
Not quite, but it's a different error:

```
/tmp/github-snapshot-118119-7341-1bckwqy.x0yv/snapshot-source.js:63
    throw new Error('Cannot use `console` functions in the snapshot.');
    ^

Error: Cannot use `console` functions in the snapshot.
    at Object.consoleNoop (/tmp/github-snapshot-118119-7341-1bckwqy.x0yv/snapshot-source.js:63:11)
    at Object../node_modules/scheduler/cjs/scheduler.production.min.js (/tmp/github-snapshot-118119-7341-1bckwqy.x0yv/snapshot-source.js:41685:45)
    at customRequire (/tmp/github-snapshot-118119-7341-1bckwqy.x0yv/snapshot-source.js:94:47)
    at Object../node_modules/scheduler/index.js (/tmp/github-snapshot-118119-7341-1bckwqy.x0yv/snapshot-source.js:34932:26)
    at customRequire (/tmp/github-snapshot-118119-7341-1bckwqy.x0yv/snapshot-source.js:94:47)
    at Object../node_modules/react-dom/cjs/react-dom.production.min.js (/tmp/github-snapshot-118119-7341-1bckwqy.x0yv/snapshot-source.js:7869:122)
    at customRequire (/tmp/github-snapshot-118119-7341-1bckwqy.x0yv/snapshot-source.js:94:47)
    at Object../node_modules/react-dom/index.js (/tmp/github-snapshot-118119-7341-1bckwqy.x0yv/snapshot-source.js:847:26)
    at customRequire (/tmp/github-snapshot-118119-7341-1bckwqy.x0yv/snapshot-source.js:94:47)
    at Object../test/output/transpiled/lib/github-package.js (/tmp/github-snapshot-118119-7341-1bckwqy.x0yv/snapshot-source.js:191:23)
```

smashwilson:
After a little digging, here's the console statement that's causing it to fail:

```js
"function"!==typeof A&&get_console().error("This browser doesn't support requestAnimationFrame. Make sure that you load a polyfill in older browsers. https://fb.me/react-polyfills")
```

And here's where `A` is assigned:

```js
A="function"===typeof requestAnimationFrame?requestAnimationFrame:void 0
```

My guess is that this is a (minified) feature probe that React is doing to test for necessary browser features at first require, which of course isn't going to work because `requestAnimationFrame` isn't defined when the alias is defined.

We might be able to fake this out somehow? If we bind a stub function to `requestAnimationFrame`, though, I'm concerned that React will try to use our stub instead of the real function even when it's executing with a proper DOM.

I'll have to think about this one... it'll likely take some electron-link trickery to get it working :zap:

smashwilson:
... Ah. So this is "working" with 16.6.3 because [the feature probe was updated to check for `MessageChannel` as well](https://github.com/facebook/react/commit/8feeed10d8f79a0c01ca293890880cbe72b3788d#diff-603a307ec39e05daabd1c651dc2ffb15). electron-link's blueprint file stubs `window`, which fooled the check before and triggered the branch that attempted to run a `console.error` statement, but the new code uses the `setTimeout` fallback.

I'm pretty sure this means that React will be snapshotted using a lower-performance "old browser" codepath.

tell us how you feel, Captain Picard

Not great. So... I believe we should make electron-link fake more of a DOM environment to trick feature probing like this into using real DOM features.

Issue #1774, opened 2018-10-31
Tooltips for Git and GitHub buttons in status bar


annthurium:
While conducting a user interview yesterday, @kuychaco pointed out that there are no tooltips for the Git and GitHub buttons in the status bar.  The branch management button also does not have a hover tooltip (the menu is itself a tooltip, but ehh, that's more of an implementation detail.)

Tooltips that trigger on hover might help users understand what these buttons do before they click and or improve discoverability.  Or they might just annoy our users and take up valuable screen real estate.  Seems worth investigating.

StarTrooper08:
This is great idea of adding tooltip . So that the beginners could able understand the tools and control.
I would like to give a try to contribute to this issue #1848 . But before that can I get to know some detailed information about this scenario . Since I m a beginner to open source but have hands on programming from years . Thank you

PR #1848, opened 2018-12-18
add tooltips to git && GitHub status bar icons

annthurium:
### Description of the Change

During usability testing, some users tried to hover over the git and GitHub icons in the status bar.  I noticed that these buttons don't have tooltips, but the rest of the status bar icons do.

This pull request adds tooltips to these status bar icons that show up on hover.

### Alternate Designs

- We thought about just saying "Click to open the git || GitHub pane."  However, @mofo37 and myself thought telling users what they can accomplish by doing so would be more informative.

### Benefits

- Making our UI more consistent (all the other status bar things have tooltips)
- Possibly improving discoverability of the git and GitHub panes
- Cleaned up a few `autobind` calls while I was at it. 

### Possible Drawbacks

Users who already know what the buttons do might be annoyed by the real estate taken up by the tooltips.  However, the tooltip delay is relatively long.  We can always increase it if there's a problem.

### Applicable Issues

https://github.com/atom/github/issues/1774

### Metrics

It might be beneficial to collect metrics about when tooltips are shown.  However, actually showing the tooltip is handled by some Atom core apis, so it's a bit out of scope for us to track that here.

We do already have metrics showing how many times the icons are clicked, so if there's a jump in that, we'll know this change was beneficial for discoverability.  That said, I'm not expecting a massive increase.

### Tests

- Manually tested that tooltips show up as expected, and the buttons still open the panes.
- Unit tests for both components verifying that tooltip shows up and contains the expected message.

### Documentation
N/A

### Release Notes

Added tooltips to the git and GitHub status bar icons that are shown on hover.

This change was a result of research we've already done.  woo!

### DIFF
--- a/lib/views/github-tile-view.js
+++ b/lib/views/github-tile-view.js
@@ -2,17 +2,21 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Octicon from '../atom/octicon';
 
+import RefHolder from '../models/ref-holder';
+import Tooltip from '../atom/tooltip';
 import {addEvent} from '../reporter-proxy';
 import {autobind} from '../helpers';
 
 export default class GithubTileView extends React.Component {
   static propTypes = {
     didClick: PropTypes.func.isRequired,
+    tooltips: PropTypes.object.isRequired
   }
 
   constructor(props) {
     super(props);
     autobind(this, 'handleClick');
+    this.refTileNode = new RefHolder();
   }
 
   handleClick() {
@@ -22,12 +26,22 @@ export default class GithubTileView extends React.Component {
 
   render() {
     return (
-      <button
-        className="github-StatusBarTile inline-block"
-        onClick={this.handleClick}>
-        <Octicon icon="mark-github" />
-        GitHub
-      </button>
+      <span ref={this.refTileNode.setter}>
+        <button
+          className="github-StatusBarTile inline-block"
+          onClick={this.handleClick}>
+          <Octicon icon="mark-github" />
+          GitHub
+        </button>
+        <Tooltip
+          key="tooltip"
+          manager={this.props.tooltips}
+          target={this.refTileNode}
+          title={`<div style="text-align: left; line-height: 1.2em;">${'hi'}</div>`}
+          showDelay={atom.tooltips.hoverDefaults.delay.show}
+          hideDelay={atom.tooltips.hoverDefaults.delay.hide}
+        />
+      </span>
     );
   }
 }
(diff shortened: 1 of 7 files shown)


50Wliu:
Would it help to include the number of files changed in the tooltip somewhere? I also noticed that our other tooltips don't specify actions (i.e. they are more passive).
`x unstaged files, y staged files`

simurai:
> Would it help to include the number of files changed in the tooltip somewhere?

Ya, that seems like a good idea. Not sure if people would use it. Might be quicker to just open the panel instead of waiting for the tooltip to appear. But you have to click again to hide it. 🤔 Anyways, I like the idea to make the tooltips show additional details that don't fit in the tile. Like the GitHub tooltip could show the current PR ID (`#12345`).

> Making our UI more consistent (all the other status bar things have tooltips)

The branch picker also doesn't have a tooltip yet.

lib/views/github-tile-view.js
```diff
@@ -2,32 +2,46 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Octicon from '../atom/octicon';
 
+import RefHolder from '../models/ref-holder';
+import Tooltip from '../atom/tooltip';
 import {addEvent} from '../reporter-proxy';
-import {autobind} from '../helpers';
+
+export const tooltipMessage = 'Click to view open GitHub pull requests';
```
simurai: Should we remove the "Click to"? It feels a bit redundant and would make it quicker to read. Like on .com, it says what you can do (if you click):

image

Maybe we can also mention the checkout feature which is a good reason to view a PR in Atom. Maybe:

`View and checkout GitHub pull requests`

annthurium: @simurai the push pull tooltip says "click to push / pull / whatever" so I was keeping the copy consistent with that.

That said, I'm not strongly attached to "click to".

simurai: Yeah, I guess the "Click to fetch/pull/push" is somewhat needed to differentiate from the "Right-click for more"? Hmmm.. 🤔 Maybe it would be a good chance to teach what fetch/pull/push does? Like

- Fetch -> [Fetch to see if there are new commits | Right-click for more options]
- Pull -> [Pull to add new commits to your local branch | Right-click for more options]
- Push -> [Push to share your local commits | Right-click for more options]
- Publish -> [Publish to share your local branch | Right-click for more options]

This is probably technically wrong, but it's kinda how I would explain it. :joy: Below the other tiles:

- Branch -> [Create or switch branches]
- GitHub -> [View and checkout GitHub pull requests]
- Git -> [Stage, view and commit changes]

Feel free to merge with "Click to ...", the suggestions above probably need a bit more thought and testing with users.

PR #1849, opened 2018-12-18
Remove status section from feature proposal template

vanessayuenn:
I notice we have been inconsistent at remembering to change the status of feature requests from `proposed` to `accepted` when we merge the feature, and I don't think the status conveys much important information.

-----
[View rendered docs/feature-requests/000-template.md](https://github.com/atom/github/blob/vanessayuenn-patch-1/docs/feature-requests/000-template.md)


smashwilson (approved):
:+1: Good call.

PR #1855, opened 2018-12-19
Fix: wrong prefix on PR files changed

### LINKED ISSUE #1853: Changed files tabs show inaccurate file paths
### Description

It is a `git diff` default behaviour to use `a/` to prefix source file and `b/` to prefix destination file (see the [docs](https://git-scm.com/docs/git-diff#_generating_patches_with_p) for more detail). Within the codebase of the pacakge, we work around this by using [`--no-prefix` flag](https://github.com/atom/github/blob/master/lib/git-shell-out-strategy.js#L612) to get diffs, but changed files tab instead get the diff from GitHub API, which doesn't have a no-prefix option. As a result, some file paths in PR changed files have their prefixes included as well. 

### Steps to Reproduce

1. Open "Files" tab in PR view
2. Observe that some file paths have `b/` in front of it.

**Expected behavior:**

File patch title shows the accurate file path.

**Actual behavior:**

Some file paths show up with `b/` in front of it.

**Reproduces how often:**

100%

### Additional Information
image

vanessayuenn:
fixes #1853

### DIFF
--- a/lib/containers/pr-changed-files-container.js
+++ b/lib/containers/pr-changed-files-container.js
@@ -59,7 +59,16 @@ export default class PullRequestChangedFilesContainer extends React.Component {
   }
 
   buildPatch(rawDiff) {
-    const diffs = parseDiff(rawDiff);
+    const diffs = parseDiff(rawDiff).map(diff => {
+      // diff coming from API will have the defaul git diff prefixes a/ and b/
+      // e.g. a/file1.js and b/file2.js
+      // see https://git-scm.com/docs/git-diff#_generating_patches_with_p
+      return {
+        ...diff,
+        newPath: diff.newPath ? diff.newPath.replace(/^[a|b]\//, '') : diff.newPath,
+        oldPath: diff.oldPath ? diff.oldPath.replace(/^[a|b]\//, '') : diff.oldPath,
+      };
+    });
     return buildMultiFilePatch(diffs);
   }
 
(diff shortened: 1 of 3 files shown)


lib/containers/pr-changed-files-container.js
```diff
@@ -59,7 +59,16 @@ export default class PullRequestChangedFilesContainer extends React.Component {
   }
 
   buildPatch(rawDiff) {
-    const diffs = parseDiff(rawDiff);
+    const diffs = parseDiff(rawDiff).map(diff => {
```
annthurium: do we also need to worry about where we're using `parseDiff` in `git-shell-out-strategy`?  If so, we should consider extracting this functionality to some kind of helper.

I seem to remember some chatter that git gives us different output than the REST api does, so this might not actually be a problem.

annthurium (approved):
Thanks @vanessayuenn for fixing!

Left one question, but it's non blocking.

lib/containers/pr-changed-files-container.js
vanessayuenn: yes

vanessayuenn: >I seem to remember some chatter that git gives us different output than the REST api does,

Indeed! It is a git diff default behaviour to use `a/` to prefix the source file and `b/` to prefix the destination file. In `git-shell-out-strategy`, we work around this by using `--no-prefix` flag to get diffs, but for changed files tab, we don't have that option since we're getting the diff from the REST API.

PR #1856, opened 2018-12-19
Fetch and display readonly inline comments.

annthurium:
### Description of the Change

Another step in our quest to march ever onwards towards code review in the editor.

This pull request adds readonly comments to pull request pane, that are displayed inline with the diff.

See design here: https://github.com/atom/github/blob/master/docs/feature-requests/003-pull-request-review.md

### TODO

[implementation tracking project](https://github.com/atom/github/projects/16)

### Alternate Designs

- Some alternate visual designs were discussed here: https://github.com/atom/github/blob/master/docs/feature-requests/003-pull-request-review.md

### Benefits

- Viewing inline comments allows users to address feedback from code reviewers without context switching back and forth to GitHub dotcom.
- Lays the groundwork for editable comments, which would allow reviewers to review code within their editor.  Less context switching, and the ability to see more of the code you're reviewing, and it's easier to run the code, which should lead to easier and more thorough code reviews.

### Applicable Issues

N/A

### Metrics

- Number of times the tab that contains the comments is viewed.  This helps us understand whether users are discovering the feature.

### Test Plan

- [x] PRs with a large # of reviews and comments should load in a reasonable amount of time
- [x]  ensure that comment threads show up correctly when there are multiple comments on the same row
- [x] ensure that comment thread order is retained when viewing a pull request that has paginated comments and reviews (that is, if there are multiple comment threads, the threads show up in the right order)
- [x] ensure that comment order within a thread is retained when viewing a pull request that has paginated comments and reviews
- [x] ensure that comments are positioned on the same line of the diff as they are on github dotcom
- [x] single comments (that are NOT part of a review) should show up
- [x] resolved comments should not show up
- [x] outdated comments should not show up
- [x] replies to resolved comments should not show up
- [x] replies to outdated comments should not show up
- [x] replies to deleted comments should not show up
- [ ] errors should be logged in the console (how should we simulate this condition? Turning off internet means you get an error loading the diff, so we won't even get to the point where we start fetching stuff.)

### Release Notes

The GitHub package now shows readonly inline comments in the pull request view.

We have a user experience research script for the GitHub pane and the pull request view, that will help us understand how the information we're displaying is or is not useful to our users.  We intend to resume user research in January, after the holidays.


lib/views/pr-comments-view.js
```diff
@@ -0,0 +1,46 @@
...
+      return this.props.reviews.nodes.map(review => {
+        return review.comments.nodes.map(comment => {
+          if (comment.position !== null) {
+            const filePatch = this.props.multiFilePatch.getFilePatchByPath(comment.path);
+            const commentStartPoint = filePatch.getStartRange().start.translate(new Point(comment.position, 0));
+            const range = new Range(commentStartPoint, commentStartPoint);
+            const author = comment.author;
+            return (
+              <Marker key={`pr-comment-${comment.id}`} bufferRange={range} invalidate="never">
+                <Decoration type="block" className="github-PrComment">
```
simurai: FYI: I changed the class name from `github-FilePatchView-prComments-controlBlock` to `github-PrComment`. To reflect [make `PullRequestCommentsView` into its own component](https://github.com/atom/github/pull/1856/commits/b341c5dd9c1cca132a260a28fd5bce55abc541ac). Styles are in `/styles/pr-comment.less`.

kuychaco:
For reference on motivation for pagination -- https://github.com/github/analytics/pull/454

lib/controllers/pr-reviews-controller.js
```diff
@@ -0,0 +1,156 @@
...
+      },
+    );
+  }
+
+  render() {
+    if (!this.props.pullRequest || !this.props.pullRequest.reviews) {
+      return null;
+    }
+
+    const commentThreads = Object.keys(this.state).reverse().map(rootCommentId => {
```
annthurium: Looks like we're reversing the comment thread order time we render.  Seems wasteful. Is there a way to avoid that?

lib/views/pr-detail-view.js
```diff
@@ -345,10 +349,14 @@ export class BarePullRequestDetailView extends React.Component {
     this.props.relay.refetch({
       repoId: this.props.repository.id,
       issueishId: this.props.pullRequest.id,
-      timelineCount: 100,
```
annthurium: I noticed when testing that this changes the appearance of the timeline on the overview tab a bit, since we're only fetching half the data we used to and users have to click "load more" more often.  Not the biggest deal, just something to be aware of / make sure we're ok with before changing this const.

lib/controllers/pr-reviews-controller.js
```diff
@@ -0,0 +1,156 @@
...
+    // react batches calls to setState and does not update state synchronously
+    // therefore we need an intermediate state so we can do checks against keys
+    // we have just added.
+    const state = {};
+    sortedReviews.forEach(({comments}) => {
+      comments.edges.forEach(({node: comment}) => {
+        if (!comment.replyTo) {
+          state[comment.id] = [comment];
+        } else {
+          // TODO: look at this more closely...
```
annthurium: I'll add this to our tracking board to make sure it's addressed before we ship.

lib/containers/pr-review-comments-container.js
```diff
@@ -0,0 +1,139 @@
...
+      loadMore: PropTypes.func.isRequired,
+      isLoading: PropTypes.func.isRequired,
+    }).isRequired,
+  }
+
+  componentDidMount() {
+    this.handleComments();
+  }
+
+  handleComments = error => {
```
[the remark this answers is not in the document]
annthurium: hmm so actually, we are logging errors and stopping fetching if the errors exist, collecting the comments, and then attempting to load more comments.  So `logErrors` doesn't cover it all either.  Not sure what a better name would be but surely we can come up with something. 🤔 anyone have suggestions?

lib/controllers/pr-reviews-controller.js
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    if (this.props.relay.isLoading()) {
+      setTimeout(this._loadMoreReviews, PAGINATION_WAIT_TIME_MS);
+    } else {
+      this._loadMoreReviews();
+    }
+  }
+
+  handleError = error => {
```
annthurium: same as other comment above re: need a better name for this method.

lib/controllers/pr-reviews-controller.js
```diff
@@ -0,0 +1,155 @@
...
+    // therefore we need an intermediate state so we can do checks against keys
+    // we have just added.
+    const state = {};
+    sortedReviews.forEach(({comments}) => {
+      comments.edges.forEach(({node: comment}) => {
+        if (!comment.replyTo) {
+          state[comment.id] = [comment];
+        } else {
+          // TODO: look at this more closely...
+          // When comment being replied to is outdated...?? Not 100% sure...
```
annthurium: I tried replying to an outdated comment, and it did not show up in the pull request view.  So yeah, I have no idea how we'd get into this state.

@kuychaco are there other ways you'd suggest we investigate this?  I'm not sure what else to do.

smashwilson: Maybe when replying to a deleted or hidden comment?

lib/models/patch/multi-file-patch.js
```diff
@@ -321,6 +339,13 @@ export default class MultiFilePatch {
     return false;
   }
 
+  getBufferRowForDiffPosition = (fileName, diffRow) => {
+    // TODO verify that this works on Windows
```
smashwilson: 🚨 Oh right, I still need to do this. I think dotcom always gives us `fileName` with `/` separators, but we normalize them to `\` before display.

lib/views/pr-review-comments-view.js
```diff
@@ -0,0 +1,89 @@
...
+    return [...this.props.commentThreads].map(({rootCommentId, comments}) => {
+      const rootComment = comments[0];
+      if (!rootComment.position) {
+        return null;
+      }
+
+      const nativePath = toNativePathSep(rootComment.path);
+      const row = this.props.getBufferRowForDiffPosition(nativePath, rootComment.position);
+      const point = new Point(row, 0);
+      const range = new Range(point, point);
```
smashwilson: We could probably extract a helper for "make a one-point `Range` at position 0 of this row". 🤔

lib/views/pr-review-comments-view.js
```diff
@@ -0,0 +1,89 @@
...
+  render() {
+    const author = this.props.comment.author;
+    const login = author ? author.login : 'someone';
+    return (
+      <div className="github-PrComment">
+        <header className="github-PrComment-header">
+          <img className="github-PrComment-avatar" src={author ? author.avatarUrl : ''} alt={login} />
+          {login} commented{' '}
+          <a className="github-PrComment-timeAgo" href={this.props.comment.url}>
+            <Timeago time={this.props.comment.createdAt} />
```
smashwilson: 👍 for code re-use :smile:

test/containers/pr-review-comments-container.test.js
```diff
@@ -0,0 +1,136 @@
...
+      const wrapper = shallow(buildApp({relayHasMore, relayIsLoading, relayLoadMore: relayLoadMoreStub}));
+      // advancing the timer and resetting the stub to clear the initial calls of
+      // _attemptToLoadMoreComments when the component is initially mounted.
+      clock.tick(PAGINATION_WAIT_TIME_MS);
+      relayLoadMoreStub.reset();
+
+      wrapper.instance()._attemptToLoadMoreComments();
+      assert.strictEqual(relayLoadMoreStub.callCount, 0);
+
+      clock.tick(PAGINATION_WAIT_TIME_MS);
```
smashwilson: Oh god `setTimeout` tests 🙈

test/controllers/pr-reviews-controller.test.js
```diff
@@ -0,0 +1,284 @@
...
+        },
+      })),
+      pageInfo: {
+        startCursor: `result${o.reviewStartCursor}`,
+        endCursor: `result${o.reviewStartCursor + o.reviewSpecs.length}`,
+        hasNextPage: o.reviewStartCursor + o.reviewSpecs.length < o.reviewItemTotal,
+        hasPreviousPage: o.reviewStartCursor !== 0,
+      },
+      totalCount: o.reviewItemTotal,
+    };
```
smashwilson: Note that we _do_ have review and review comment builders for these 😄

lib/containers/issueish-detail-container.js
```diff
@@ -138,10 +146,14 @@ export default class IssueishDetailContainer extends React.Component {
...
       issueishNumber: this.props.issueishNumber,
-      timelineCount: 100,
+      timelineCount: PAGE_SIZE,
       timelineCursor: null,
-      commitCount: 100,
+      commitCount: PAGE_SIZE,
       commitCursor: null,
+      reviewCount: PAGE_SIZE,
+      reviewCursor: null,
+      commentCount: PAGE_SIZE,
```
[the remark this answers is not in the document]
telliott27: I pulled some distribution stats for number of reviews and number of review comments here: https://ghe.io/telliott27/reports/tree/master/Notebooks/PR%20Review%20Comments#pr-reviews-and-comments-percentiles

The TL;DR is that it is rare for a PR to have more than 10 reviews, and for reviews to have more than 10 comments.

lib/containers/pr-review-comments-container.js
```diff
@@ -0,0 +1,139 @@
...
+    }),
+    relay: PropTypes.shape({
+      hasMore: PropTypes.func.isRequired,
+      loadMore: PropTypes.func.isRequired,
+      isLoading: PropTypes.func.isRequired,
+    }).isRequired,
+  }
+
+  componentDidMount() {
+    this.handleComments();
```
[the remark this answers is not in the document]
annthurium: we are also calling this method in `_loadMoreComments` - it's the callback that gets passed to `relay.loadMore`.  We do want to keep calling `handleComments` each time we receive new props, because the parent component needs to know when all the comments have been received, so it can group them by `replyTo.id`.  Does the suggestion of making this a pure component still make sense under those conditions?

lib/containers/pr-review-comments-container.js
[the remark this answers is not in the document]
annthurium: I like `accumulateComments`!  Good suggestion.  Will fix.

lib/views/pr-review-comments-view.js
annthurium: yeah! I did not realize `Timeago` existed until this pull request.  We should probably be using `Timeago` in the `PullRequestCommitView` and a few other places where we're rolling our own humanizing timestamp.

lib/controllers/pr-reviews-controller.js
annthurium: okay, so I tried:

- replying to a deleted comment
- replying to a resolved comment
- replying to a hidden comment

None of these triggered that condition for line 132.  However, I did notice we are not respecting hidden comments.  That seems like a ship blocker to me since we should not be showing users comments that have been marked as abusive or spammy.

lib/models/patch/multi-file-patch.js
smashwilson: Spoiler alert: they were not handled correctly. It was the worst of both worlds: we were showing dotcom diffs with `/` separators, but _also_ throwing an exception here when trying to load comments.

paths-not-fixed-up

I've fixed the path normalization so that the "Files" tab will (a) display paths with native path separators and (b) correctly compute comment offsets with the normalized paths.

properly-normalized

yo-dawg


image

lib/views/pr-review-comments-view.js
annthurium: Hmm, this is the only place we're doing that though. So not sure we need a helper.

annthurium:
the last build never reported its status properly but it was green on azure pipelines I promise!

Issue #1858, opened 2018-12-20
(Un)staged changes does not use editor font size


50Wliu:
### Prerequisites

* Have you tried to reproduce the problem in [Safe Mode][safe-mode]?
* Have you followed all applicable steps in the [debugging guide][debugging-guide]?
* Have you checked the FAQs on the [message board][message-board] for common solutions?
* Have you checked that your issue [isn't already filed][already-filed]?

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

### Description

The TextEditors used for the (un)staged changes view do not respect the editor font size I have configured.
![unstaged-changes-font-size](https://user-images.githubusercontent.com/2766036/50303129-27a1e580-045a-11e9-99e4-cc1c05e173ce.png)
![editor-font-size](https://user-images.githubusercontent.com/2766036/50303180-47390e00-045a-11e9-972a-0351c76848e4.png)

It looks like this was noticed before with the old view in #331 and fixed in #333.

### Steps to Reproduce

1. Set the editor font size to anything other than 12
2. Open (un)staged changes
3. Compare to a normal text editor

**Expected behavior:** Font sizes to be identical

**Actual behavior:** Font sizes are different

**Reproduces how often:** 100%

### Versions

You can get this information from copy and pasting the output of `atom --version` and `apm --version` from the command line. Also, please include the OS and what version of the OS you're running.
```
Atom    : 1.35.0-dev-add03bdbe
Electron: 2.0.16
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

simurai:
That's :grimacing: intended: https://github.com/atom/github/commit/1390a6a8327fa3aaa0060a98c581c9ff8f1d0f88

When adding the changed files https://github.com/atom/github/pull/1829 to PRs, I felt that the default font size is too large. In that commit ☝️, the font size got locked to the font size of UI themes.

But I'll happily revert that commit again. An alternative could be that the file path and hunk titles use the UI font-size, but the diff/code use the font size from Atom's config. Like this:

![screen shot 2018-12-21 at 10 47 28 am](https://user-images.githubusercontent.com/378023/50320077-11edf800-050e-11e9-9a58-185e13e48b7e.png)

[...]

PR #1859, opened 2018-12-21
Adjustable diff font size

simurai:
### Description of the Change

This PR makes the font size of diffs adjustable based on Atom's editor config. The header font size is set by the UI themes.

screen shot 2018-12-21 at 10 47 28 am

Bonus:

- File headers now use the font family from UI themes. Before it was hard-coded.
- Hunk headers now use the font family from the config. Before it was hard-coded.

### Alternate Designs

N/A

### Benefits

Users can change the font size of diffs in Atom's settings. 

### Possible Drawbacks

Some might have a different preference.

### Applicable Issues

Fixes https://github.com/atom/github/issues/1858

### Metrics

N/A

### Tests

1. Open a diff
2. Press `cmd -` or `cmd +`
3. Observe how the code changes the font size, but the file and hunk header stay the same.

resize

### Documentation

N/A

### Release Notes

N/A

N/A

### DIFF
--- a/styles/file-patch-view.less
+++ b/styles/file-patch-view.less
@@ -41,7 +41,7 @@
     padding-left: @component-padding;
     border: 1px solid @base-border-color;
     border-radius: @component-border-radius;
-    font-family: system-ui;
+    font-family: @font-family;
     font-size: @font-size;
     background-color: @header-bg-color;
     cursor: default;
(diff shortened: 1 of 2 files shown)


simurai:
Let's merge this before the next release, since it could be seen as a regression.

Issue #1860, opened 2018-12-21
Add the ability to collapse and expand files


smashwilson:
Within the MultiFilePatch components.

PR #1863, opened 2018-12-27
Collapse and expand files within a multi file patch view.

annthurium:
### Description of the Change

This change adds the ability to collapse and expand individual files in the `MultiFilePatchView`.

### Alternate Designs

The collapse buttons could be on the right:

screen shot 2019-01-02 at 12 38 56 pm

### Benefits

- Improves the experience of viewing multi file patches by allowing users to reduce visual clutter.

### Possible Drawbacks

TBD

### Applicable Issues

https://github.com/atom/github/issues/1860

### Metrics

- We could measure the count of file patches that are expanded / collapsed.

### Tests

TBD

### Documentation

TBD

### Release Notes

The GitHub package now supports collapsing and expanding files in the pull request, commit preview, and commit pane views.

### DIFF
--- a/lib/views/file-patch-header-view.js
+++ b/lib/views/file-patch-header-view.js
@@ -4,6 +4,7 @@ import React, {Fragment} from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 
+import Octicon from '../atom/octicon';
 import RefHolder from '../models/ref-holder';
 import IssueishDetailItem from '../items/issueish-detail-item';
 import ChangedFileItem from '../items/changed-file-item';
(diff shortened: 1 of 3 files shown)


simurai:
Moved the chevron icon to the left of the title, so that it matches more the tree-view or search results.

screen shot 2019-01-02 at 1 12 47 pm

We could also make the whole title clickable to make it easier to collapse/uncollapse?

annthurium:
just a note for when we prioritize this work again...

@smashwilson suggested trying to add / remove text from the buffer.  However, since the TextBuffer is currently read only, that approach wouldn't work without editable diffs.

smashwilson:
> However, since the TextBuffer is currently read only, that approach wouldn't work without editable diffs.

If you're modifying it through the API, there's a [`bypassReadOnly` flag you can provide](https://atom.io/docs/api/v1.34.0/TextEditor#instance-setText). Or you can modify the `TextBuffer` directly instead of going through the `TextEditor`.

annthurium:
closing this in favor of https://github.com/atom/github/pull/1913 -- we realized that collapsing large files shares enough logic with collapsing and expanding any file that we should work on both of those at the same time.

PR #1884, opened 2019-01-04
Upgrade to the latest react and react-dom

smashwilson:
Now that atom/electron-link#15 is out, we should be able to update to the latest React builds without hitting weird snapshotting errors or silently falling back to ancient browser shims.

Fixes #1832.


smashwilson:
Oh, right, this probably needs #1885.

annthurium (commented):
do we need to do any manual testing?  Or are you pretty confident that the newest versions of these libs aren't gonna break anything??

also, is this the version with hooks??

smashwilson:
> do we need to do any manual testing? Or are you pretty confident that the newest versions of these libs aren't gonna break anything??

For React, yeah, I was going to at least take it for a spin and make sure it runs. Our tests _should_ catch anything that regressed, but just to be sure.

> also, is this the version with hooks??

lol

screen shot 2019-01-04 at 6 21 23 pm

smashwilson:
I fired it up and clicking around a bit and it looks like we're still good 👌

annthurium (approved):
cool, let's ship it!  Thanks for doing the manual testing.

PR #1885, opened 2019-01-04
Synchronize electron-link and electron-mksnapshot versions to Atom's

smashwilson:
To ensure that the snapshot test is as accurate as possible, lock electron-link and electron-mksnapshot to the version ranges specified in [atom/atom's package.json](https://github.com/atom/atom/blob/master/script/package.json#L14-L15).

Closes #1861; closes #1834.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2650,9 +2650,9 @@
       },
       "dependencies": {
         "debug": {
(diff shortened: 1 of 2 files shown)


annthurium (approved):
looks good to me!

PR #1889, opened 2019-01-08
Exclude Relay-generated GraphQL files from code coverage

smashwilson:
**Please be sure to read the [contributor's guide to the GitHub package](https://github.com/atom/github/blob/master/CONTRIBUTING.md) before submitting any pull requests.**

### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* Suggestion: You can use checklists to keep track of progress for the sections on metrics, tests, documentation, and user research.

### Description of the Change

Exclude files generated by `relay-compiler` from collected code coverage.

### Alternate Designs

I don't think there really is one.

### Benefits

It'll make our code coverage go up :wink:

### Possible Drawbacks

We might accidentally add a file with a name matching `*.graphql.js` in a directory called `__generated__` and exclude it by mistake... ?

### Applicable Issues

Fixes #1886.

### Metrics

_N/A_

### Tests

_N/A_

### Documentation

_N/A_

### Release Notes

_N/A_

### User Experience Research (Optional)

_N/A_

PR #1892, opened 2019-01-08
Revert "Exclude Relay-generated GraphQL files from code coverage"

smashwilson:
It looks like this broke CodeCov somehow. Let's revert until we can figure out why.

Reverts atom/github#1889.

### DIFF
--- a/.nycrc.json
+++ b/.nycrc.json
@@ -7,7 +7,6 @@
   "exclude": [
     "lib/views/git-cache-view.js",
     "lib/views/git-timings-view.js",
-    "lib/relay-network-layer-manager.js",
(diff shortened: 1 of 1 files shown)


smashwilson:
This commit _also_ errored out on codecov import: https://codecov.io/gh/atom/github/commit/ca5d94102e4b4b8568c7cb5dbf656b1486e1afc6

I wonder if there's something wrong with the way it's mapping paths? I'll keep looking tomorrow.

smashwilson:
Ah, looks like this was a transient CodeCov problem - coverage reports from rebuilds are being accepted. Hooray?

PR #1895, opened 2019-01-09
So it turns out we weren't actually using GraphQL at all

smashwilson:
Let's just... get rid of it?


smashwilson:
Ahhh it's a peerDependency from relay-compiler. Nevermind 🙈

PR #1897, opened 2019-01-09
PR Review Feature Request: review comment changes

smashwilson:
In our sprint planning meeting today, @vanessayuenn and @kuychaco had a number of ideas for how we could handle:

* **Displaying review summary comments.** Putting these in a separate "Reviews" tab within the `PullRequestDetailItem` feels redundant and "dotcom-y".
* **Displaying review comments within open TextEditors.** @vanessayuenn brought up that displaying the review comment content directly in the TextEditor with a block decoration could be disruptive to the editing experience: if there's a huge comment thread in the middle of the function you're trying to edit, you want to be able to read the comments _and_ see the full function's source at once, even if the review comment was left on a line in the middle. Additionally, it would be beneficial to give users a way to see where a specific review is within the review it belongs to, and to navigate quickly among them.

@vanessayuenn, @kuychaco: Please correct me if I've misrepresented what you were going for, or if you've changed your minds about anything since the ideas have had a chance to percolate somewhat :smile:

cc @simurai, because this is some of the stuff we'd like to run by you in our design meeting tonight 😇 

-----

✨ Prototype: `prototypes/1897-reviews` branch ✨ Use `github:open-mock` command to see the "Reviews dock" item.

reviews

-----
[View rendered docs/feature-requests/003-pull-request-review.md](https://github.com/atom/github/blob/aw/pr-review-update/docs/feature-requests/003-pull-request-review.md)

### DIFF
--- a/docs/feature-requests/003-pull-request-review.md
+++ b/docs/feature-requests/003-pull-request-review.md
@@ -51,7 +51,7 @@ A panel at the bottom of the pane shows the progress for resolved review comment
 
 When the pull request is checked out, an "open" button is shown in the review footer. Clicking "open" opens a `PullRequestReviewsItem` for this pull request's review comments as an item in the right workspace dock.
 
-### Files Changed (tab)
+### Files (tab)
 
 Clicking on the "Files Changed" tab displays the full, multi-file diff associated with the pull request. This is akin to the "Files changed" tab on dotcom.
 
--- a/docs/feature-requests/003-pull-request-review.md
+++ b/docs/feature-requests/003-pull-request-review.md
@@ -119,7 +119,7 @@ Clicking "Finish your review" from a comment or clicking "Review Changes" in the
 This item is opened in the workspace's right dock when the user:
 
 * Clicks the review progress bar in the GitHub tab.
-* Clicks the "open" button on the review summary footer of a `PullRequestDetailItem`.
+* Clicks the "open reviews" button on the review summary footer of a `PullRequestDetailItem`.
 * Clicks the "<>" button on a review comment in the "Files Changed" tab of a `PullRequestDetailItem`.
 
 It shows a scrollable view of all of the reviews and comments associated with a specific pull request,
@@ -200,6 +200,12 @@ It was a great improvement, but filtering the diff with radio buttons and checkb
 - Bring back the sub-navigation, but make it look less .com-y.
 - Keep using an editable editor for the diffs, but add some padding.
 - Introduce a "Reviews" footer to all sub-views to allow creating/submit a review, no matter where you are.
+
+#### Third iteration
+
+Long comments can disrupt the code editing experience.  Our third iteration keeps the review comments in a dock, a la Google Docs.  This helps code authors more easily address comments, because they can see the comments and also get them out of the way.
+
+Since this approach different from previous approaches, we performed a series of [usability studies](https://github.com/github/pe-editor-tools/blob/master/community/usability-testing/atom_rcid_research_summary.md) to validate that users would find this approach useful.
 
 
 ## Unresolved questions
--- a/docs/feature-requests/003-pull-request-review.md
+++ b/docs/feature-requests/003-pull-request-review.md
@@ -32,7 +32,7 @@ Peer review is also a critical part of the path to acceptance for pull requests 
 
 #### Header
 
-![header](https://user-images.githubusercontent.com/378023/46536358-3829d180-c8e9-11e8-9167-3d1003ab566b.png)
+![header](https://user-images.githubusercontent.com/378023/51305325-e400c280-1a7d-11e9-9b4e-b9cf2d326dd5.png)
 
 At the top of each `PullRequestDetailItem` is a summary about the pull request, followed by the tabs to switch between different sub-views.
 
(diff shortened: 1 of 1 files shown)


smashwilson:
As an update on our progress here:

* We'll use the prototypes that @simurai and @vanessayuenn are hashing out in some of our upcoming UXR sessions, to hear how prospective users feel about some of the directions we could take.
* @vanessayuenn and @kuychaco will assemble a write-up of the design directions that we're exploring, articulating and giving a name to each, so we can explore a discussion of the pros and cons of each.

simurai:
At some point we have to decide how reviews and review comments should be sorted. To get a feel for it, there are two prototypes for the "first level" of sorting:

By reviewer | By file 
--- | ---
Review comments get shown in the same order as they are made and therefore grouped by reviewer. | Review comments and summaries are split into two groups. Review comments are ordered by file and line number.
Branch: `vy/prototype-for-uxr` | Branch: `vy/prototype-sorted-by-file`
by-reviewer | by-file

There are cons and pros for both and it maybe comes down to personal preference, but something I just realized now: When writing a review, the reviewer might talk about the previously made comment. If the comments get sorted by file, they may appear in a different order as they have been written and then it may not make sense if they say something like "same as above". See this example:

How the comments got written | Shown in a random order (by file)
--- | ---
as-authored | by-file

Therefore I think as a default, showing review comments in the same order how they got written makes the most sense.

robertrossmann:
I'd like to suggest how I would personally prefer to see the reviews in Atom.

For me, there are two types of review comments (not sure how they are named here):

1. The PR review comment - it's the comment you leave when you hit the green "Submit Review" button
  > I would prefer these comments to be situated in a completely separate group somewhere in the UI, probably on top
1. The review comment - it's the comment you leave on a line of code which is directly tied to a specific position/code in a file
  > I would prefer to see these comments grouped by file, sorted by the time they were authored by the reviewer, because it is easier for me to address all comments for a specific file while I have that file open, rather than switch files all the time.

To address the problem with some review comments being shown in weird order, I think the ideal solution would be to sort the files in the same way they are sorted on Github.com. Then the comments can be grouped by file and sorted by time of creation and still they would appear in the same way as they appear on Github.com . This, however, requires that the overall review comments be shown separately, maybe on top of the review comments, in their own section maybe. 🤔

vanessayuenn:
>I think the ideal solution would be to sort the files in the same way they are sorted on Github.com. 
I like this suggestion a lot! ✨

simurai:
> I think the ideal solution would be to sort the files in the same way they are sorted on Github.com.

That’s a great suggestion. :+1:

I just realized that the `vy/prototype-sorted-by-file` prototype already is ordered the same as github.com. So I guess the question is: Does it not bother you that there might be review comments from  different reviews/reviewers mixed together? On github.com you have both options:

1. Go to [Files changed](https://github.com/atom/github/pull/1919/files) to read comments by file.
2. Go to [Conversation](https://github.com/atom/github/pull/1919#pullrequestreview-196307045) to read comments by review.

In the ["this is a really long name"](https://github.com/atom/github/pull/1919#pullrequestreview-196307045) example, it's nice that the other related comments are following right after so you can address them at the same time.

Too much of an edge case? If so, we could implement "sorted by files" first and _maybe_ add an option to sort by review later.

smashwilson:
Could we implement per-review filtering in `vy/prototype-sorted-by-file`? Maybe have a checkbox or a toggling eye icon next to each review in the "Reviews" section, then only render review comment threads that include review comments from selected reviews. If you wanted to see the comments from a single review, you could filter it to only that one.

Also of these two options I'm 👍 on `vy/prototype-sorted-by-file`, btw. I like that it makes review comment threads with comments from different reviews a non-issue, and I like that it groups comment threads in a way that you're likely to naturally resolve them.

robertrossmann:
> So I guess the question is: Does it not bother you that there might be review comments from different reviews/reviewers mixed together?

I believe you almost always want to see **all** review comments at once for any given file, for several reasons:

1. You usually want to address all the review comments at once while you are working on that file because they might guide you in a different direction when considered together than when addressing them one by one
1. The comments trigger a discussion among reviewers & PR author and you want to see the whole context and the eventual resolution of that discussion (which might be different than the original review comment might have suggested ⚠️)

Therefore, I think that that seeing review comments from multiple reviewers for a given file is almost always preferred to seeing only specific people's comments. If the concern is "seeing too many comments at once" then I feel the ability to collapse review comments of other files/seeing only review comments for a single file at once would provide enough visual clarity & focus.

> ⚠️ I have never worked on a project where I would have to deal with more than 2 PR reviews (3 people collaborating), so my opinions might differ from those of who have worked on large scale projects like the Node.js repo etc.

simurai:
> Could we implement per-review filtering in vy/prototype-sorted-by-file? Maybe have a checkbox or a toggling eye icon next to each review in the "Reviews" section, then only render review comment threads that include review comments from selected reviews. If you wanted to see the comments from a single review, you could filter it to only that one.

Here review summaries with checkboxes that lets you hide review comments from other reviews:

image

Hmmm.. 🤔 That would work, but feels a bit complex? Showing avatar and time for collapsed headers could already be enough to see which review comments "belong together"?

image

The last review comment ☝️ is not part of any review because it's a "single comment", but I think that doesn't really matter. I often make a review and then after submitting the review, make more single comments that should be "part" of the review. 🤷‍♂️

smashwilson:
Merging so the implementation PR #1995 has somewhere up-to-date to point :tada:

PR #1901, opened 2019-01-10
Test coverage for GitTempDir

smashwilson:

### Requirements

### Description of the Change

Adds test coverage for `GitTempDir` to prevent coverage flapping in unrelated pull requests.

### Alternate Designs

I could have only covered the lines that weren't hit by other codepaths, but it's a small class so I just covered the whole thing.

### Benefits

GitTempDir is one of the culprits in test coverage flapping that we see on unrelated PRs. This is one step toward minimizing those changes and keeping CodeCov output relevent.

### Possible Drawbacks

_N/A_

### Applicable Issues

_N/A_

### Metrics

_N/A_

### Tests

This should raise GitTempDir's coverage to 100%.

### Documentation

_N/A_

### Release Notes

_N/A_

_N/A_

### DIFF
--- a/test/git-temp-dir.test.js
+++ b/test/git-temp-dir.test.js
@@ -1,0 +1,71 @@
+import fs from 'fs-extra';
+import path from 'path';
+
+import GitTempDir, {BIN_SCRIPTS} from '../lib/git-temp-dir';
+
+describe('GitTempDir', function() {
+  it('ensures that a temporary directory is populated', async function() {
+    const tempDir = new GitTempDir();
+    await tempDir.ensure();
+
+    const root = tempDir.getRootPath();
+    for (const scriptName in BIN_SCRIPTS) {
+      const script = BIN_SCRIPTS[scriptName];
+      const stat = await fs.stat(path.join(root, script));
+      assert.isTrue(stat.isFile());
+      if (script.endsWith('.sh') && process.platform !== 'win32') {
+        // eslint-disable-next-line no-bitwise
+        assert.isTrue((stat.mode & fs.constants.S_IXUSR) === fs.constants.S_IXUSR);
+      }
+    }
+
+    await tempDir.ensure();
+    assert.strictEqual(root, tempDir.getRootPath());
+  });
+
+  it('generates getters for script paths', async function() {
+    const tempDir = new GitTempDir();
+    await tempDir.ensure();
+
+    const scriptPath = tempDir.getScriptPath('git-credential-atom.js');
+    assert.isTrue(scriptPath.startsWith(tempDir.getRootPath()));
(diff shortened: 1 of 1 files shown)


smashwilson:
Looks like [we're good](https://codecov.io/gh/atom/github/pull/1901/src/lib/git-temp-dir.js) 👌

test/git-temp-dir.test.js
```diff
@@ -0,0 +1,71 @@
...
+      const stat = await fs.stat(path.join(root, script));
+      assert.isTrue(stat.isFile());
+      if (script.endsWith('.sh') && process.platform !== 'win32') {
+        // eslint-disable-next-line no-bitwise
+        assert.isTrue((stat.mode & fs.constants.S_IXUSR) === fs.constants.S_IXUSR);
+      }
+    }
+
+    await tempDir.ensure();
+    assert.strictEqual(root, tempDir.getRootPath());
```
annthurium: just for my own education...

you assign `root = tempDir.getRootPath` on line 11, and then assert that that these are still equal.  With this assertion, are you testing that `ensure()` does not change the root path?

test/git-temp-dir.test.js
```diff
@@ -0,0 +1,71 @@
...
+    const tempDir = new GitTempDir();
+    assert.throws(() => tempDir.getAskPassJs(), /uninitialized GitTempDir/);
+  });
+
+  if (process.platform === 'win32') {
+    it('generates a valid named pipe path for its socket', async function() {
+      const tempDir = new GitTempDir();
+      await tempDir.ensure();
+
+      assert.match(tempDir.getSocketPath(), /^\\\\\?\\pipe\\/);
```
annthurium: fun regex.  oh, windows.

annthurium (approved):
thanks for making our test coverage more stable!  I have a question or two for my own education but nothing blocking. 

Why were these files causing flapping in the first place?  Were there timing issues or something, where sometimes we met certain conditions and sometimes not?

test/git-temp-dir.test.js
smashwilson: > With this assertion, are you testing that `ensure()` does not change the root path?

Yup, exactly. I wanted to assert that multiple calls to `ensure()` wouldn't create new temp directories, basically.

test/git-temp-dir.test.js
smashwilson: Named pipes are fun 🙃

smashwilson:
> Why were these files causing flapping in the first place? Were there timing issues or something, where sometimes we met certain conditions and sometimes not?

`GitTempDir` was only tested incidentally through `GitShellOutStrategy` and `Repository` tests, so we had a few lines that weren't being explicitly exercised. Getting us to full coverage with an explicit test is the best way to make sure these don't fall through the cracks.

Issue #1817, opened 2018-12-01
Message template does not ignore commented lines like Git does


lee-dohm:
### Prerequisites

* Have you tried to reproduce the problem in [Safe Mode][safe-mode]?
* Have you followed all applicable steps in the [debugging guide][debugging-guide]?
* Have you checked the FAQs on the [message board][message-board] for common solutions?
* Have you checked that your issue [isn't already filed][already-filed]?

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

### Description

Comment lines are included in the commit message upon commit when your `~/.gitmessage` file contains commented lines.

### Steps to Reproduce

1. Create a `~/.gitmessage` file that contains two blank lines followed by two lines starting with a `#` character
1. Execute the command `touch test` from the command-line inside a Git repository
2. Launch Atom in Safe Mode
3. Open the Git panel
1. Double-click the `test` file in the Unstaged Changes panel
1. Type `foo` in the first line of the Git commit message box
1. Click "Commit to master"
1. Go back to the command line and execute `git log -1`

**Expected behavior:**

The commit comment text to consist only of `foo`

**Actual behavior:**

The commit comment contains `foo`, the blank line, and the two following commented lines

**Reproduces how often:**

100%

### Versions

```
Atom    : 1.35.0-nightly2
Electron: 2.0.14
Chrome  : 61.0.3163.100
Node    : 8.9.3
```

### Additional Information

* I haven't tested other configurations of commented and uncommented lines (such as starting with comments and then having blank lines)
* My `~/.gitmessage` file is symlinked to my dotfiles directory (though I hope that won't make a difference)

annthurium:
just merged https://github.com/atom/github/pull/1902 which should resolve this

PR #1902, opened 2019-01-10
make commit message template ignore commented lines

annthurium:
### Description of the Change

This change fixes a bug  where commit message templates that include commented lines were included in the body of the commit. 

If a commit template exists, istrip commented lines from the commit message. 

### Alternate Designs

- My initial approach had us ignoring the `verbatim` flag if a commit template was present.  However, we want to preserve the verbatim flag because otherwise commits made from the mini editor might be formatted weirdly in some cases.  See https://github.com/atom/github/pull/1500 for more info. 

- We discussed ripping out the `verbatim` flag entirely, to aid complexity, but wanted to maintain the hard wrapping so as not to break formatting for users who are using the mini editor but not using a template.

### Benefits

- Users can make commits with message templates, without including commented lines in the commit output, because probably they do not want that.
- Consistent with command line git behavior

### Possible Drawbacks

- Adds more complexity to some code that's already rife with special cases.
- There might be some users who have commit templates and want to keep lines that start with `#`

### Applicable Issues

https://github.com/atom/github/issues/1817

### Metrics

N/A

### Tests

- Manually tested making a commit with a commit template set, where the template included `#` characters.  Verified that those commented lines did not show up when using `git log` to inspect the commit.
- unit test in `git-shell-out-strategy` to verify that commented lines are stripped when using commit message templates.  (We already have a unit test for validating that commented lines are not stripped, if a commit template does not exist.)

### Documentation

Added some code comments.

### Release Notes

Fixed an issue where commits made with commit message templates were not stripping comment characters.

### DIFF
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -506,8 +506,12 @@ export default class GitShellOutStrategy {
       msg = rawMessage;
     }
 
+    // if commit template is used, that will not play nicely with `verbatim`
+    // because we want to strip commented lines from templates.
+    const template = await this.fetchCommitMessageTemplate();
+
     // Determine the cleanup mode.
-    if (verbatim) {
+    if (verbatim && !template) {
       args.push('--cleanup=verbatim');
     } else {
       const configured = await this.getConfig('commit.cleanup');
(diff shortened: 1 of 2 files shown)


kuychaco (approved):
Looks good to me! 👍 :shipit:

Issue #1875, opened 2019-01-03
Show committer name or GitHub username in CommitDetailItem


smashwilson:
> I usually recognize my co-workers/fellow open-sourcerers by their names or by their Github handles. I would prefer to see at least the committer's name in the commit details pane.
>
> > Stretch goal: show a card on hover with additional information fetched from Github ([inspiration](https://atom.io/packages/github-user-datatip))

![screenshot 2019-01-03 at 12 01 06](https://user-images.githubusercontent.com/3058150/50634615-9483a800-0f4f-11e9-9796-374a7e486487.png)

cc @roberrossmann from https://github.com/atom/github/issues/1870#issuecomment-451115442

annthurium:
just merged https://github.com/atom/github/pull/1906 so this should be in the next release!  Thanks @robertrossman for the suggestion.

robertrossmann:
Awesome, thanks for incorporating that idea! You guys & gals are great! ❤️

PR #1906, opened 2019-01-12
add committer name to commit detail pane

annthurium:
### Description of the Change

This change adds the committer's full name to the commit detail pane.  Instead of email address, we show the full name.

screen shot 2019-01-15 at 1 38 23 pm

### Alternate Designs

- I considered showing the name and email address.  However, it looked kinda cluttered, and I couldn't find a good way of showing co author info in that scenario. 

image

### Benefits

- Makes it easier to understand at a glance who made the commit
- Feature has been requested by folks in the community

### Possible Drawbacks

- If folks need the email address of a committer for some reason, now it would take more effort to find it. 

### Applicable Issues

https://github.com/atom/github/issues/1875

### Metrics

N/A

### Tests

- Manually tested that committer name shows up as expected for single commits
- Commiter name shows up as expected for co authored commits
- name falls back to something reasonable (such as email address or GitHub username) if committer name is not set in git config
- all new lines covered by unit tests

### Documentation

N/A

### Release Notes

TBD

N/A

### DIFF
--- a/lib/git-shell-out-strategy.js
+++ b/lib/git-shell-out-strategy.js
@@ -700,12 +700,13 @@ export default class GitShellOutStrategy {
     // %x00 - null byte
     // %H - commit SHA
     // %ae - author email
+    // %an = author full name
     // %at - timestamp, UNIX timestamp
     // %s - subject
     // %b - body
     const args = [
       'log',
-      '--pretty=format:%H%x00%ae%x00%at%x00%s%x00%b%x00',
+      '--pretty=format:%H%x00%ae%x00%an%x00%at%x00%s%x00%b%x00',
       '--no-abbrev-commit',
       '--no-prefix',
       '--no-ext-diff',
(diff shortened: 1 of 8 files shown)


annthurium:
hey @simurai -- this isn't urgent, but if you feel like making this look better, that'd be cool.

Right now it's like: 
screen shot 2019-01-11 at 4 46 20 pm

annthurium:
Note for future Tilde or whomever ends up picking this up...

I tried testing the case where the commiter name does not exist by unsetting my email, username, and github handle in my git config.  Git won't let you commit without one of those things set.

If I unset my github user alias, somehow `name` is falling back to `annthurium`.  So weird!  I don't think it's being pulled from my email address either, because I tried modifying that. Needs further investigation.

smashwilson:
From now on, all of my commit messages will be "hunks hunks hunks"

simurai:
How about replacing the current email with the author name and not have both?

image

And only use the email as a fallback if name or username doesn't exist?

Then the case of multiple authors should also be covered. E.g. `[name] and 3 others committed...`.

annthurium:
@simurai I like that!  I'll move this in that direction.

simurai (approved):

PR #1919, opened 2019-01-24
Refactoring builders [do not merge]

annthurium:
This pull request extracts a PatchBuffer out of the common code in (three) places we build up TextBuffers with Markers organized into MarkerLayers (hunks, patches, additions, deletions, unchanged lines...). It also adds operations to the PatchBuffer model that will let us build them and manipulate them as you expand and collapse individual patches, preserving marker locations.

Also, we changed `test/builders/patch.js` to build "raw" what-the-diff output and use the "real" builders to build model objects and getting our tests green where those are used.

### DIFF
--- a/lib/controllers/commit-detail-controller.js
+++ b/lib/controllers/commit-detail-controller.js
@@ -16,7 +16,7 @@ export default class CommitDetailController extends React.Component {
     this.state = {
       messageCollapsible: this.props.commit.isBodyLong(),
       messageOpen: !this.props.commit.isBodyLong(),
-    };
+    }
   }
 
   render() {
--- a/lib/models/patch/builder.js
+++ b/lib/models/patch/builder.js
@@ -189,10 +189,27 @@ function buildHunks(diff, patchBuffer, insertRow) {
   const inserter = patchBuffer.createInserterAt([insertRow, 0]);
 
   let patchMarker = null;
+  let firstHunk = true;
   const hunks = [];
+
+  // Separate multiple patches on the same buffer with an unmarked newline
+  if (patchBuffer.getBuffer().getLength() > 0) {
+    console.log('inserting newline to separate Patches');
+    inserter.insert('\n');
+  }
+
   inserter.markWhile(Patch.layerName, () => {
     for (const rawHunk of diff.hunks) {
+      let firstRegion = true;
       const regions = [];
+
+      // Separate hunks with an unmarked newline
+      if (firstHunk) {
+        firstHunk = false;
+      } else {
+        console.log('inserting newline to separate Hunks');
+        inserter.insert('\n');
+      }
 
       inserter.markWhile(Hunk.layerName, () => {
         let currentRegionText = '';
@@ -201,6 +218,14 @@ function buildHunks(diff, patchBuffer, insertRow) {
         function finishRegion() {
           if (currentRegionText === '' || CurrentRegionKind === null) {
             return;
+          }
+
+          // Separate regions with an unmarked newline
+          if (firstRegion) {
+            firstRegion = false;
+          } else {
+            console.log('inserting newline to separate Regions');
+            inserter.insert('\n');
           }
 
           inserter.insertMarked(currentRegionText, CurrentRegionKind.layerName, {
@@ -215,9 +240,12 @@ function buildHunks(diff, patchBuffer, insertRow) {
           if (NextRegionKind === undefined) {
             throw new Error(`Unknown diff status character: "${rawLine[0]}"`);
           }
-          const nextLine = rawLine.slice(1) + '\n';
+          const nextLine = rawLine.slice(1);
 
           if (NextRegionKind === CurrentRegionKind) {
+            if (currentRegionText !== '') {
+              currentRegionText += '\n';
+            }
             currentRegionText += nextLine;
             continue;
           } else {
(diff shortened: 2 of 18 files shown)


lib/controllers/pr-reviews-controller.js
```diff
@@ -19,6 +19,7 @@ export default class PullRequestReviewsController extends React.Component {
       ),
     }),
     getBufferRowForDiffPosition: PropTypes.func.isRequired,
+    isPatchTooLargeOrCollapsed: PropTypes.func.isRequired,
```
annthurium: this is a really long name -- can you come up with something more concise?

lib/models/patch/multi-file-patch.js
```diff
@@ -339,6 +301,70 @@ export default class MultiFilePatch {
...
+
+    filePatch.triggerExpandIn(this.patchBuffer, {before, after});
+
+    this.filePatchesByMarker.set(filePatch.getMarker(), filePatch);
+    for (const hunk of filePatch.getHunks()) {
+      this.hunksByMarker.set(hunk.getMarker(), hunk);
+    }
+  }
+
+  isPatchTooLargeOrCollapsed = filePatchPath => {
```
annthurium: same as above - this name is too long and the line length linters are gonna be unhappy.

test/views/pr-comments-view.test.js
```diff
@@ -6,6 +6,23 @@ import {pullRequestBuilder} from '../builder/pr';
...
 describe('PullRequestCommentsView', function() {
+  function buildApp(multiFilePatch, pullRequest, overrideProps) {
+    const relay = {
+      hasMore: () => {},
+      loadMore: () => {},
+      isLoading: () => {},
+    };
+    return shallow(
+      <PullRequestCommentsView
+        isPatchTooLargeOrCollapsed={sinon.stub().returns(false)}
```
annthurium: again, this variable name is too long

lib/controllers/commit-detail-controller.js
```diff
@@ -16,7 +16,7 @@ export default class CommitDetailController extends React.Component {
     this.state = {
       messageCollapsible: this.props.commit.isBodyLong(),
       messageOpen: !this.props.commit.isBodyLong(),
-    };
+    }
```
annthurium: mind adding the missing semicolon here?

lib/models/patch/builder.js
```diff
@@ -1,31 +1,43 @@
...
 import {Unchanged, Addition, Deletion, NoNewline} from './region';
 import FilePatch from './file-patch';
 import MultiFilePatch from './multi-file-patch';
 
-export function buildFilePatch(diffs) {
-  const layeredBuffer = initializeBuffer();
+export const DEFAULT_OPTIONS = {
+  // Number of lines after which we consider the diff "large"
+  // TODO: Set this based on performance measurements
+  largeDiffThreshold: 100,
```
annthurium: @simurai: how many lines do you think constitutes a large diff?  Not just from a performance perspective, but from a user experience perspective.  Like how many lines is disruptive to a user when they're trying to read, because often large diffs are the result of auto generated code.

simurai: Hmmm.. will large diffs be collapsed by default or there is a "load" button? Maybe if the diff is so large that it fills the whole scroll height. Then I can uncollapse only if I'm really interested in that file. 100 seems fine. 👍

annthurium: @vanessayuenn @kuychaco @smashwilson care to weigh in?

lib/models/patch/file-patch.js
```diff
@@ -200,6 +274,34 @@ export default class FilePatch {
     }
   }
 
+  /*
+   * Construct a String containing diagnostic information about the internal state of this FilePatch.
+   */
+  inspect(opts = {}) {
```
annthurium: nice!  This is going to be so useful when we are trying to debug the marker layer on a text buffer.

lib/models/patch/builder.js
vanessayuenn: mm we were using 1000 lines as the threshold before, but that's for performance reasons. 100 does seem a bit small though considering it's counting both deleted and added lines. 🤔 300 seems a bit more reasonable, but still an arbitrary number.

Aerijo: > but still an arbitrary number.

Would this be good as a package option then? Pick a value that should work for most users, but let people who need to configure it. Also, would that be something that can be refined by `metrics`?

lib/containers/changed-file-container.js
```diff
@@ -27,13 +29,26 @@ export default class ChangedFileContainer extends React.Component {
   constructor(props) {
     super(props);
     autobind(this, 'fetchData', 'renderWithData');
+
+    this.lastMultiFilePatch = null;
+    this.sub = new CompositeDisposable();
```
vanessayuenn: one line comment -- not part of a review.

lib/containers/changed-file-container.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import yubikiri from 'yubikiri';
+import {CompositeDisposable} from 'event-kit';
```
vanessayuenn: lalalala

PR #1938, opened 2019-02-05
Minor tweaks to release process

kuychaco:
I'll update this with any other tweaks I come across as I walk through the rest of the release process

**Notable changes:**

Removed Atom CI verification step. 
Rationale: This step was originally introduced to catch snapshotting problems, but we now have other measures for catching this. Specifically we have a CI job that runs on our atom/github PRs, and we do a local build of Atom with a pre-release package version as part of our release QA testing. 

Cut release branch at the start.
Rationale: So that work can continue on master without interfering with the release process.

Added steps for handling when publishing fails (make sure to delete tag and get release commit out of commit history)

-----
[View rendered docs/how-we-work.md](https://github.com/atom/github/blob/kuychaco-patch-1/docs/how-we-work.md)


docs/how-we-work.md
```diff
@@ -131,19 +131,28 @@ The github package ships as a bundled part of Atom, which affects the way that o
...
+    1. Cherry changes from master to release branch.
+    1. Begin again by cutting a new pre-release and proceeding through the above steps once again.
   * :white_check_mark: _Otherwise,_ comment in and close the issue, then continue.
 8. _In your atom/github repository:_ run `apm publish minor` to publish the next minor version.
+  * :boom: _If publishing fails,_ before trying to publish again
+    1. Check if a release commit was created (`git log`). If one exists, remove it from the commit history (`git reset --hard <sha-prior-to-commit>`).
+    1. Check if a release tag was created (`git tag`). If one exists, delete it (`git tag -d 0.${MINOR}.0`).
+    1. Address the problem that interfered with publishing.
+    1. Try to publish again with `apm publish minor`.
 9. _In your atom/github repository:_ create a release branch for this minor version with `git checkout -b 0.${MINOR}-releases`. Push it to atom/github.
```
smashwilson: This step should be deleted now, yeah?

docs/how-we-work.md
```diff
@@ -131,19 +131,28 @@ The github package ships as a bundled part of Atom, which affects the way that o
...
-9. _In your atom/atom repository:_ update the version of the `"github"` entry beneath `packageDependencies` in `package.json`  to match the published minor version. Run `script/build` to update `package-lock.json` files. Commit and push these changes.
+9. _In your atom/atom repository:_ checkout a new branch (`git checkout -b bump-github-${VERSION}`), update the version of the `"github"` entry beneath `packageDependencies` in `package.json`  to match the published minor version. Run `script/build` to update `package-lock.json` files. Commit and push these changes.
```
smashwilson: Sidenote: you can get away with just running `script/bootstrap` instead of `script/build`. But `script/build` is safer because it actually runs through the final snapshotting and everything, so let's leave it like this.

PR #1949, opened 2019-02-07
Diff color tweaks

### LINKED ISSUE #1933: Opening staging view autohighlights first hunk
### Description

With most recent updates whenever I open the staging view for a file it seems to auto highlight the first hunk entirely in blue making it now more difficult to view that hunk without first clicking off of it.

### Steps to Reproduce

1. Have a project that is using git
2. Edit a tracked file so that it is recognized by git as having changed
3. Open the git side panel
4. Click on the changed file to open the staging view

**Expected behavior:**

The file to open in staging view with nothing selected

**Actual behavior:**

The file opens and the first hunk is always auto highlighted with the blue colors making it harder to see due to contrast.

**Reproduces how often:**

100% of the time

### Versions

Atom    : 1.34.0
Electron: 2.0.16
Chrome  : 61.0.3163.100
Node    : 8.9.3
MacOs Mojave 10.14.1

### Additional Information

image

simurai:
### Description of the Change

This makes the blue-ish background color weaker when selecting the whole hunk. But keeps it stronger when selecting individual lines/words/characters.

Some additional changes:

- Tweaked the diff colors
- Replaced the gutter icons to be text based. The icons felt a bit too bold.

### Screenshot/Gif

Before | After
--- | ---
screen shot 2019-02-07 at 5 40 10 pm | screen shot 2019-02-07 at 5 39 34 pm

diff

### Alternate Designs

1. When clicking the "See All Staged Changes" button with the mouse, we could keep the focus on the button and not move it to the first hunk. That would also match the behaviour of single file diffs.
2. We could differentiate between making a selections of single characters/words and selecting the whole line. Maybe have two classes `github-FilePatchView-line--selected` and `github-FilePatchView-line--partiallySelected`.
3. Separate "selecting text/code" and "select to staging". Same behaviour as in Desktop. Maybe necessary for editable diffs.
4. Also add diff coloring (red/green) to the gutters. Then when selecting characters/words, it's still easy to see if a line is deleted or added.

/cc @kuychaco ☝️ 

### Benefits

Easier to read the diff when the first hunk gets automatically selected in commit previews.

### Possible Drawbacks

The difference between a selected hunk and selected lines might be confusing?

### Applicable Issues

Closes #1933

### Metrics

N/A

### Tests

Tested with Atom's bundled themes and a few community themes. Certain themes work better than others.

### Documentation

N/A

### Release Notes

N/A

N/A

### DIFF
--- a/styles/file-patch-view.less
+++ b/styles/file-patch-view.less
@@ -298,6 +298,16 @@
   }
 }
 
+// Selected + focused hunkmode
+.github-FilePatchView.github-FilePatchView--hunkMode:focus-within {
+
+  atom-text-editor {
+    .selection .region {
+      background-color: mix(@button-background-color-selected, @syntax-background-color, 12%);
+    }
+  }
+}
+
 .gitub-FilePatchHeaderView-basename {
   font-weight: bold;
 }
--- a/styles/file-patch-view.less
+++ b/styles/file-patch-view.less
@@ -184,20 +184,32 @@
 
   &-line {
     // mixin
-    .hunk-line-mixin(@bg;) {
-      background-color: fade(@bg, 16%);
+    .hunk-line-mixin(@color) {
+      background-color: fade(@color, 15%);
 
       &.line.cursor-line {
-        background-color: fade(@bg, 22%);
+        background-color: fade(@color, 25%);
       }
     }
 
-    &--deleted {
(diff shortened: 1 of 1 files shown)


simurai:
Ok, let's merge this, regardless of potentially implementing one of the alternatives above ☝️.

PR #1954, opened 2019-02-14
CodeCov package upgrade

smashwilson:
Go back to the [official CodeCov package](https://www.npmjs.com/package/codecov) now that it has a release that includes [Azure DevOps support](https://github.com/codecov/codecov-node/pull/114).


kuychaco (approved):
👍

PR #1956, opened 2019-02-15
React upgrades

smashwilson:
Hooooks


smashwilson:
Gave this a shot in dev mode and it looks like we're rendering all clean 👍

vanessayuenn (approved):

PR #1967, opened 2019-02-19
keytar@4.4.0

smashwilson:
Split off from #1942.


smashwilson:
Merging without review because it's a dependency bump 🤞

PR #1980, opened 2019-02-25
react@16.8.3

smashwilson:


smashwilson:
Tested out locally and it works okay 👍

PR #1983, opened 2019-02-26
Refactor some of the item components

### LINKED ISSUE #1923: Item components share a lot of code
Our "item" React components - `ChangedFileItem`, `CommitDetailItem`, `CommitPreviewItem`, `GitTabItem`, `GithubTabItem`, `IssueishDetailItem` - share code for most of their implementation. It'd be nice to refactor their common behavior out somehow to reduce boilerplate and cut down on the amount of copy-and-paste in maintenance.

vanessayuenn:
Resolves #1923. I am sure there is stuff that can be further refactored, but I didn't want to overdo it. Thought this would be a good start.

### DIFF
--- a/lib/items/issueish-detail-item.js
+++ b/lib/items/issueish-detail-item.js
@@ -1,6 +1,6 @@
-import React, {Component} from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
-import {Emitter} from 'event-kit';
+import ItemComponent from './item-component';
 
 import {autobind} from '../helpers';
 import {GithubLoginModelPropType, WorkdirContextPoolPropType} from '../prop-types';
@@ -8,9 +8,8 @@ import {addEvent} from '../reporter-proxy';
 import Repository from '../models/repository';
 import {getEndpoint} from '../models/endpoint';
 import IssueishDetailContainer from '../containers/issueish-detail-container';
-import RefHolder from '../models/ref-holder';
 
-export default class IssueishDetailItem extends Component {
+export default class IssueishDetailItem extends ItemComponent {
   static propTypes = {
     // Issueish selection criteria
     // Parsed from item URI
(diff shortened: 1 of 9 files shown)


lib/items/issueish-detail-item.js
```diff
@@ -150,42 +146,15 @@ export default class IssueishDetailItem extends Component {
...
       uri: this.getURI(),
       deserializer: 'IssueishDetailItem',
     };
   }
 
-  getTitle() {
-    return this.title;
-  }
-
   observeEmbeddedTextEditor(cb) {
```
smashwilson: `observeEmbeddedTextEditor` is another "magic hidden Atom item protocol method." It might be worth pulling into `ItemComponent` even if it isn't widely used yet, just to keep it close to the similar methods.

smashwilson (changes requested):
The autobind thing is the only actual "we should change this before :ship:" issue I found - other than that it's all comments and discussion. Thanks for tackling this, the duplication has been bugging me 😆

lib/items/item-component.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import RefHolder from '../models/ref-holder';
+import {Emitter} from 'event-kit';
+import {autobind} from '../helpers';
+
+export default class ItemComponent extends React.Component {
```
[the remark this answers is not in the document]
vanessayuenn: I think you're right in that inheritance is inherently (heh) not very react-y. Upon some investigation, I think wrapper component but implemented with render prop pattern might be the best way to go, especially since we have quite some existing render prop patterns in the codebase already.

lib/items/issueish-detail-item.js
vanessayuenn: sounds good!

vanessayuenn: yea totally!

vanessayuenn: totally!

vanessayuenn:
Closing this out for now until we figure out a better approach for the refractor :)

PR #1989, opened 2019-02-27
Suggested changes styling

### LINKED ISSUE #1970: suggested changes styling looks janky.
Open a pull request that has suggested changes in Atom.  This is how it looks:

screen shot 2019-02-20 at 4 17 03 pm

The same suggested changes on dotcom look a lot more polished, including line numbers and such:

screen shot 2019-02-20 at 4 17 07 pm

While it would be nice in the long term to support applying suggested changes within Atom, in the short term we can at least make them look a bit nicer.

simurai:
### Description of the Change

This styles the suggested changes in review comments.

### Screenshot/Gif

Before | After
--- | ---
before | after

### Alternate Designs

N/A

### Benefits

Looks less broken. Easier to see what the suggestion is.

### Possible Drawbacks

The diffs use the same colors for dark and light themes again. This could make certain themes look less optimized.

### Applicable Issues

Closes #1970

### Metrics

N/A

### Tests

1. Open a review comment that has a "suggested change".
2. Verify the suggested change looks styled.

### Documentation

N/A

### Release Notes

N/A

N/A

### DIFF
--- a/styles/pr-comment.less
+++ b/styles/pr-comment.less
@@ -5,7 +5,7 @@
 
 
 .github-PrCommentThread {
-  padding: @component-padding 0;
+  padding: @component-padding @component-padding @component-padding 0;
 }
 
 
--- a/styles/pr-comment.less
+++ b/styles/pr-comment.less
@@ -58,4 +58,86 @@
     vertical-align: middle;
   }
 
+
+  // Suggested changes
+  .js-suggested-changes-blob {
+    font-size: .9em;
+    border: 1px solid @base-border-color;
+    border-radius: @component-border-radius;
+
+    .d-flex { display: flex; }
+    .d-inline-block { display: inline-block; }
+    .flex-auto { flex: 1 1 auto; }
+    .p-1 { padding: @component-padding/2; }
+    .px-1 { padding-left: @component-padding/2; padding-right: @component-padding/2; }
+    .p-2 { padding: @component-padding; }
+    .text-gray { color: @text-color-subtle; }
+    .lh-condensed { line-height: 1.25; }
+    .rounded-1 { border-radius: @component-border-radius; }
+    .border { border: 1px solid; }
+    .border-bottom { border-bottom: 1px solid @base-border-color; }
+    .border-green { border-color: @text-color-success; }
+    .octicon { vertical-align: -4px; fill: currentColor; }
+
+    .blob-wrapper > table {
+      width: 100%;
+      margin: 0;
+      font-family: var(--editor-font-family);
+      font-size: var(--editor-font-size);
+      line-height: var(--editor-line-height);
+
+      td {
+        border: none;
+        padding: 0 .5em;
+      }
+    }
+
+    .blob-num {
+      padding: 0 .5em;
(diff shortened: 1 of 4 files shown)


styles/pr-comment.less
```diff
@@ -58,4 +58,86 @@
     vertical-align: middle;
   }
 
+
+  // Suggested changes
+  .js-suggested-changes-blob {
```
simurai: I assume we don't have much control over the markup that is coming from the GitHub API?

screen shot 2019-02-27 at 3 28 47 pm

`.js-suggested-changes-blob` shouldn't be used for styling, but it seems like the best identifier to scope "suggested changes". Also, the class names from .com could change anytime and might break the styling in Atom.

simurai:
I tried to add `user-select: text;` and the `native-key-bindings` class so that the suggestion can be selected and copied, but for some reason, it copies the line above the diff (notice the jump in scroll position) and not the selection? :thinking: 

copy

Maybe not super necessary once it would be possible to apply the suggested change, but currently it's probably annoying having to retype the line.

styles/variables.less
```diff
@@ -7,3 +7,13 @@
 @gh-background-color-red: #dc3545;
 @gh-background-color-purple: #6f42c1;
 @gh-background-color-green: #28a745;
+
+
+// diff colors -----------------
+// Needs to be semi transparent to make it work with selections and different themes
+
+@github-diff-deleted:           fade(hsl(353, 100%, 66%), 15%); // similar to .com's hsl(353, 100%, 97%)
```
annthurium: so cool.  I really want to learn more about css colors.

styles/pr-comment.less
annthurium: yeah, I thought about how styling based on this classname is fragile 😦 

On the plus side, no functionality would be broken...it would just look janky again.

The "right" way to do this is to identify "suggested changes" comments based on data attributes that come from the graphQL api (which is probably possible, but I haven't looked at the schema.) . Then we apply our own classname.

annthurium (approved):

styles/variables.less
simurai: If possible, it's best to only use the [UI](https://github.com/atom/atom/blob/master/static/variables/ui-variables.less) and [syntax](https://github.com/atom/atom/blob/master/static/variables/syntax-variables.less) theme variables. Then it _should_ look ok in all themes.

But sometimes making an exception is fine too. :grimacing: In this case, it felt better to use the .com colors for the diff and not "success" and "error". But it needed to be semi transparent to make the selections visible underneath. Less has a bunch of functions to tweak colors, like [fade()](http://lesscss.org/functions/#color-operations-fade) to make an existing color more transparent.

I also have been using [mix()](http://lesscss.org/functions/#color-operations-mix) sometimes. For example 

```less
.class {
  background-color: mix(@text-color, @base-background-color, 10%);
}
```

mixes `10%` of the text color with the background color -> making the new background color slightly darker with light themes or slightly lighter in dark themes.

simurai:
Ok, let's merge this as is.

PR #2007, opened 2019-03-06
Fix bug expanding large-diff-gated patches with comments

annthurium:
### Description of the Change

Once upon a time, there was a bug where clicking "expand" on a file patch that exceeded the large diff gate size, that had comments, would throw a stacktrace.  Sadness!

@vanessayuenn and I determined that the root cause of this bug was that because large patches are initially collapsed, we were never populating the red and black tree used to keep track of the diff offset indices for comment placement.

Solution was to extract a function to populate the diff offset indices, and then call that function when expanding file patches.

### Screenshot/Gif

N/A

### Alternate Designs

None considered. 

### Benefits

Users can now expand large-diff-gated file patches that contain comments without errors and sadness.  (Well, I can't guarantee there will be no sadness.)

### Possible Drawbacks

Can't think of any drawbacks but if you can, I'm open to discussion.

### Metrics

N/A

### Tests

- Wrote unit test to ensure that:
  - we are only populating the `diffOffsetIndices` if it's empty
  - after expanding an initially-collapsed too large pass, calling `getBufferRowForDiffPosition` in fact gives us the correct buffer row.

- Manually testing expanding the file patch on https://github.com/kuychaco/test-repo/pull/7

### Documentation

Added one code comment like a champ. 

### Release Notes

This bug is for a feature that hasn't been released yet so I'm ok leaving it out of the release notes.

N/A

### DIFF
--- a/lib/models/patch/multi-file-patch.js
+++ b/lib/models/patch/multi-file-patch.js
@@ -196,6 +196,10 @@ export default class MultiFilePatch {
     return Range.fromObject([[selectionRow, 0], [selectionRow, Infinity]]);
   }
 
+  isDiffRowOffsetIndexEmpty(filePatchPath) {
+    const diffRowOffsetIndex = this.diffRowOffsetIndices.get(filePatchPath);
+    return diffRowOffsetIndex.index.size === 0;
+  }
   populateDiffRowOffsetIndices(filePatch) {
     let diffRow = 1;
     const index = new RBTree((a, b) => a.diffRow - b.diffRow);
@@ -337,10 +341,9 @@ export default class MultiFilePatch {
       this.hunksByMarker.set(hunk.getMarker(), hunk);
     }
 
-    const diffRowOffsetIndex = this.diffRowOffsetIndices.get(filePatch.getPath());
     // if the patch was initially collapsed, we need to calculate
     // the diffRowOffsetIndices to calculate comment position.
-    if (diffRowOffsetIndex.index.size === 0) {
+    if (this.isDiffRowOffsetIndexEmpty(filePatch.getPath())) {
       this.populateDiffRowOffsetIndices(filePatch);
     }
   }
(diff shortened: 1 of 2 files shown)


lib/models/patch/multi-file-patch.js
```diff
@@ -332,6 +336,13 @@ export default class MultiFilePatch {
     for (const hunk of filePatch.getHunks()) {
       this.hunksByMarker.set(hunk.getMarker(), hunk);
     }
+
+    const diffRowOffsetIndex = this.diffRowOffsetIndices.get(filePatch.getPath());
+    // if the patch was initially collapsed, we need to calculate
+    // the diffRowOffsetIndices to calculate comment position.
+    if (diffRowOffsetIndex.index.size === 0) {
```
kuychaco: 🎨minor readability suggestion -- I noticed I had to look at this a couple times, as well as `assert.strictEqual(multiFilePatch.diffRowOffsetIndices.get('1.txt').index.size, 0)` in the tests.

Perhaps we could add a method `diffRowOffsetIndexUnpopulated` that does this check and reads more declaratively. And we can also use the method to improve readability in our test assertions.

kuychaco: Hmm is there an edge case here for symlink and file mode change diffs that would have empty indexes? I guess we'd just unnecessarily call `populateDiffRowOffsetIndices` and create a new empty index, which isn't a big deal. It's probably not worth the extra effort to distinguish between empty indexes from large diffs versus symlink or file mode changes, huh?

kuychaco (approved):
Thanks for putting in this fix so quickly! It looks good 👍  

Just left a couple of comments -- one suggestion and one question about a potential edge case. But I think neither of them are necessarily ship-blocking.

lib/models/patch/multi-file-patch.js
annthurium: yeah, I'd take readability and simplicity over dealing with symlink and file mode changes here.  Creating a new empty index shouldn't impact performance very much.

I'll work on the other changes though, thanks for the suggestion.

Issue #2056, opened 2019-04-10
review comments feature improvements from community & safety audit


annthurium:
I had a meeting with @gallexi and @rebelagentm from the community and safety team to review the atom review comments feature.  We haven't found anything release blocking (yet!) although there are still some things that need to be tested.  Here are some safety improvements that Michael and Lexi suggested, that we could implement in the future. 

## Future improvements

- [x] "Edited" badge on comments
a user could make a nasty comment, then edit it to not be nasty anymore.  Atom doesn't show a badge, or any sort of indicator that a comment was edited.  This information about whether a comment is edited should be available via the graphQL api, so we should display it to users.  This is the highest priority issue we found.  Dotcom shows the full edit history of a comment, but building that would be a big effort, so showing the "edited" badge is a nice middle ground.

- [x] add commenter associations
Dotcom shows badges like "first time contributor" to encourage people to be kind to new folks, etc.  The commenter association data is available via graphQL, so we could display it in our comments dock.

- [x] link to comment on dotcom
Linking to dotcom allows users to see more context on the comment, including the edit history.  We could put this on a kebab menu, and also hide the "report abuse" link in there.

- [x] passing data when clicking the "report abuse" icon
When reporting a comment, fill in some info for the user via query string params.
example: https://github.com/contact/report-content?report=USER&content_url=URL_TO_CONTENT

PR #2057, opened 2019-04-10
Indicate that comment has been edited

kuychaco:
Now that we've shipped review comments we'll want to let folks know when a comment was edited. This was brought to our attention by the community and safety team, with respect to the fact that nasty comments can be edited.

While Dotcom shows the full edit history, we decided to simply indicate that the comment was edited and link the user to Dotcom where they can view more context if they would like to.

See #2056 for reference.

edited comment (modeled after dotcom):
Screen Shot 2019-04-10 at 4 01 15 PM

edited review:
Screen Shot 2019-04-10 at 4 35 36 PM

## Alternate designs

I explored a visual treatment where we use a badge instead of bullet-separated text.  It felt maybe a bit too visually overwhelming imo, in an already busy design.

Screen Shot 2019-04-10 at 3 59 14 PM


lib/views/reviews-view.js
```diff
@@ -270,6 +270,13 @@ export default class ReviewsView extends React.Component {
           />
           <a className="github-ReviewSummary-username" href={`https://github.com/${reviewAuthor}`}>{reviewAuthor}</a>
           <span className="github-ReviewSummary-type">{copy}</span>
+          {review.lastEditedAt !== null && (
+            <span className="github-ReviewSummary-edited">
+              &nbsp;•&nbsp;
+              <a className="github-ReviewSummary-edited" href={review.url}>edited</a>
```
kuychaco: I duplicated the `github-ReviewSummary-edited` class to get the subtle color styling to apply to the link. There's probably a better way... @annthurium perhaps you can 🎨?

Same for `github-Review-edited` below...

annthurium: adding the same classname to the `span` and the `anchor` is totally fine. 

I did extract a function for rendering the link, which allowed me to only change the css in one place.

annthurium (approved):
looks great, @kuychaco !  Thanks for getting this out so quickly.

PR #2072, opened 2019-04-15
whitelist a bunch of dependencies (or: Let's Trust Bots To Not Mess Things Up)

vanessayuenn:
These are the "safest" dependencies to update automatically -- they are only used in tests, so we can trust that as long as tests are passing in CI, they should be good to merge.


smashwilson (approved):
Yessss. Trust the robots 🤖

PR #1995, opened 2019-03-01
Reviews dock item

smashwilson:
### Description of the Change

View pull request reviews and review comments in a dock item.

* [📈  Feature project](https://github.com/atom/github/projects/20)
* [:memo: Feature request](https://github.com/atom/github/blob/master/docs/feature-requests/003-pull-request-review.md) + #1897. Specifically:
  * Review progress bar in the pull request list
  * ReviewsItem within the dock
  * Footer showing review progress in PullRequestDetailItem
* Prototype branch: `vy/prototype-for-uxr`

### Screenshot/Gif

_TODO_

### Benefits
Pull request authors will now enjoy many features for responding to code review comments from the comfort of their editor.
-  see review comments alongside the code
-  jump to code from reviews dock
- jump to diff from reviews dock
- reply to comments
- resolve comments
- see a progress bar of resolved comments, to keep track of progress
- react with emoji to comments

### Metrics
- opening the reviews tab
`='open-reviews-tab', {package: 'github', from: IssueishDetailController});`
- opening the review thread from the gutter decorations
`open-review-thread', {package: 'github'}`
- jumping to a file from a review comment thread
`'reviews-dock-open-file', {package: 'github'});`
- jumping to a diff from a reviews comment thread
`'reviews-dock-open-diff', {package: 'github'});`
- opening a pull request detail view from the reviews dock
   `'reviews-dock-open-pr', {package: 'github', component: this.constructor.name});`
- running the command to show more context lines in a review comment thread
`'reviews-dock-show-more-context', {package: 'github'});`
- running the command to show less context lines in a review comment thread
`'reviews-dock-show-less-context', {package: 'github'});`
- resolving a comment thread
`'resolve-comment-thread', {package: 'github'});`
- unresolving a comment thread
`'unresolve-comment-thread', {package: 'github'});`
- adding a single comment
`'add-single-comment', {package: 'github'});`
- starting a pull request review review
`'start-pr-review', {package: 'github', component: (ReviewsFooterView or ReviewsView})`
- adding an emoji reaction
`'add-emoji-reaction', {package: 'github'});`
-removing an emoji reaction
`'remove-emoji-reaction', {package: 'github'});`

##### PR detail view
- [x] open reviews dock from footer in pull request detail view
- [x] start a review button from footer in pull request detail view takes you to dotcom
- [x] open diff from dock
  - [x] activate changed files tab
  - [x] scrolls to where the comment is
[...]

PR #2074, opened 2019-04-15
Prefill some metadata for report abuse link in review comments

vanessayuenn:
A piece of the feedback (#2056) we got from community & safety audit for the recently shipped review comments feature (#1995) suggests that the "report abuse" link in the review comment should include some metadata, so that important info can be prefilled.

### DIFF
--- a/lib/views/reviews-view.js
+++ b/lib/views/reviews-view.js
@@ -537,7 +537,7 @@ export default class ReviewsView extends React.Component {
               <span className="github-Review-pendingBadge badge badge-warning">pending</span>
             )}
           </div>
-          <a className="github-Review-reportAbuseLink" title="report abuse" href="https://github.com/contact/report-content">
+          <a className="github-Review-reportAbuseLink" title="report abuse" href={`https://github.com/contact/report-content?report=${comment.author.login}&content_url=${encodeURIComponent(comment.url)}`}>
             <Octicon icon="alert" />
           </a>
         </header>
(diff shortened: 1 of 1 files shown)


lib/views/reviews-view.js
```diff
@@ -537,7 +537,7 @@ export default class ReviewsView extends React.Component {
               <span className="github-Review-pendingBadge badge badge-warning">pending</span>
             )}
           </div>
-          <a className="github-Review-reportAbuseLink" title="report abuse" href="https://github.com/contact/report-content">
+          <a className="github-Review-reportAbuseLink" title="report abuse" href={`https://github.com/contact/report-content?report=${comment.author.login}&content_url=${encodeURIComponent(comment.url)}`}>
```
annthurium: thanks for tacking this, @vanessayuenn !

do we need to handle cases where `comment.author` doesn't exist?  I see some `comment : comment.author` kinds of ternaries sprinkled in other places.

vanessayuenn: Good point! Yea I notice we have checks for `author` kinda inconsistently throughout `reviews-view`. I think this would be a good opportunity to clean that up.

lib/views/reviews-view.js
```diff
@@ -260,15 +260,15 @@ export default class ReviewsView extends React.Component {
       return null;
     }
 
-    const reviewAuthor = review.author ? review.author.login : '';
+    const reviewAuthor = {avatarUrl: '', login: '', ...review.author};
```
annthurium: elegant!

annthurium (approved):
lgtm, thanks for addressing this!

It might be good to have a unit testing asserting that the query string params are part of the link, but ehh, I don't feel strongly enough about it to block merging.

PR #2095, opened 2019-04-24
Feature Request: Review author flow

vanessayuenn:
The first draft of the review author workflow proposal is ready! This is nowhere near its final form yet, but hopefully, it will at least get the conversation started. 

-----
[View rendered docs/feature-requests/006-pull-request-reviewer-flow.md](https://github.com/atom/github/blob/vy/meta-rfc-pr-reviewer-flow/docs/feature-requests/006-pull-request-reviewer-flow.md)

### DIFF
--- a/docs/feature-requests/006-pull-request-reviewer-flow.md
+++ b/docs/feature-requests/006-pull-request-reviewer-flow.md
@@ -18,7 +18,11 @@ We already have an innovative review-comments-in-dock (RCID) workflow built out 
 This tab shows all review summaries and review comments, including the ones that are part of a _pending review_ that has not been submitted yet.
 
 #### Header
-- When there is no pending review, button reads "Start a pending review", clicking on which will take you to the Pending Review tab in its empty state.
+| current header | proposed header | proposed header with pending review |
+|---|---| --- |
+|<img width="380" alt="old header" src="https://user-images.githubusercontent.com/6842965/56446925-7f027600-62d3-11e9-8635-e17b946f3b1b.png">|<img width="380" alt="new header" src="https://user-images.githubusercontent.com/6842965/56446924-7f027600-62d3-11e9-98f5-0ea1e2826343.png">|<img width="380" alt="new header" src="https://user-images.githubusercontent.com/6842965/56447078-4911c180-62d4-11e9-8543-4935ba587b38.png">|
+
+- When there is no pending review, button reads "Start a review", clicking on which will take you to the Pending Review tab in its empty state.
 - When there is already a pending review, the button reads "Resume review (2)". The number indicates the number of comments currently in the pending review. When adding more pending comments _within the All Reviews tab_, there should be some emphasis on the number changing -- akin to the button on dotcom. Clicking on this button takes user to the Pending Review tab.
 
 ##### On scaling down the "Checkout" button to just an icon
--- a/docs/feature-requests/006-pull-request-reviewer-flow.md
+++ b/docs/feature-requests/006-pull-request-reviewer-flow.md
@@ -1,7 +1,3 @@
-<!---
-For community contributors -- Please fill out Part 1 of the following template. This will help our team collaborate with you and give us an opportunity to provide valuable feedback that could inform your development process. Sections in Part 2 are not mandatory to get the conversation started, but will help our team understand your vision better and allow us to give better feedback.
---->
-
 **_Part 1 - Required information_**
 
 # Pull Request Review -- Reviewer Flow
@@ -60,6 +56,36 @@ The header looks very similar to the one of All Reviews tab, with the exception 
 
 
 
+### Reviewer Workflow
+
+**Note**: In the name of clarity, in this RFC we will differentiate between "new comment" and "pending comment":
+- a new comment: _after_ a user decides to add a comment from the gutter, and _before_ the user actually adds it to a pending review. (i.e. a pending-pending comment, if you prefer confusion. :laughing:)
+- a pending comment: already added to a pending review
+
+#### 1. Start a review
+
+##### From the gutter
+
+within Files tab in `PullRequestDetailView`
+within an editor
+
+If user has checked out a PR branch, an "add comment" icon should show up *on hover* over the gutter of an editor of any file. Clicking on the icon will either activate new comment
+
+Nothing if not on a PR branch.
+
+##### By responding to a thread
+
+#### 2. Continue a review
+
+
+
+#### 3. Submit a review
+- The only way to submit a review within Atom is by using the "Submit review" button in the Pending Review tab
+- the button will be disabled if a review type has not been chosen from the dropdown menu
+- if there is any new comment that has not been added to the pending review, a warning modal should pop up
+- after publishing, the Pending Review tab will be destroyed. User will be led back to the All Reviews tab, which will immediately reflect the just published review.
+
+--------------------
 
 **_Part 2 - Additional information_**
 
--- a/docs/feature-requests/006-pull-request-reviewer-flow.md
+++ b/docs/feature-requests/006-pull-request-reviewer-flow.md
@@ -14,18 +14,51 @@ Provide code review to an existing pull request within Atom.
 
 ## :checkered_flag: Motivation
 
-We already have a workflow built out for the receiving end of pull request reviews; it only makes sense to also build out a workflow for users to author pull request reviews.
+We already have an innovative review-comments-in-dock (RCID) workflow built out for the receiving end of pull request reviews. In order to complete the full experience of code review within Atom, we should also build out a workflow for users to author pull request reviews.
 
 ## 🤯 Explanation
 
+### "All Reviews" tab
+This tab shows all review summaries and review comments, including the ones that are part of a _pending review_ that has not been submitted yet.
+
+#### Header
+- When there is no pending review, button reads "Start a pending review", clicking on which will take you to the Pending Review tab in its empty state.
+- When there is already a pending review, the button reads "Resume review (2)". The number indicates the number of comments currently in the pending review. When adding more pending comments _within the All Reviews tab_, there should be some emphasis on the number changing -- akin to the button on dotcom. Clicking on this button takes user to the Pending Review tab.
 
 
-Explain the proposal as if it was already implemented in the GitHub package and you were describing it to an Atom user. That generally means:
+- rationale for scaling down the "Checkout" button
+- caveats of doing so
+- mitigation
 
-- Introducing new named concepts.
-- Explaining the feature largely in terms of examples.
-- Explaining any changes to existing workflows.
-- Design mock-ups or diagrams depicting any new UI that will be introduced.
+#### Responding to a comment thread
+start a new review button
+
+#### Pending comments
+
+Pending comments within the All Reviews tab are styled differently from the already published comments. Pending comments contain a way to direct user to the Pending Review tab.
+
+
+
+
+### "Pending Review" tab
+This tab shows a *subset* of all reviews -- only the summary and comments of a pending review. Since a user is only allowed to have one pending review at a time, there should also only be one active Pending Review tab.
+
+#### Header
+The header looks very similar to the one of All Reviews tab, with the exception that the primary button now reads "See all reviews", and will send users back to the All Reviews tab.
+
+
+#### Summary section
+
+- sticky
+- drop down to select review type
+- button to submit review
+
+#### Comments section
+
+- empty state
+
+
+
 
 
 **_Part 2 - Additional information_**
--- a/docs/feature-requests/006-pull-request-reviewer-flow.md
+++ b/docs/feature-requests/006-pull-request-reviewer-flow.md
@@ -10,7 +10,7 @@ Provide code review to an existing pull request within Atom.
 
 ## :checkered_flag: Motivation
 
-We already have an innovative review-comments-in-dock (RCID) workflow built out for the receiving end of pull request reviews. In order to complete the full experience of code review within Atom, we should also build out a workflow for users to author pull request reviews.
+We already have an innovative review-comments-in-dock (RCID) workflow built out for the receiving end of pull request reviews. In order to complete the full code review experience within Atom, we should also build out a workflow for users to author pull request reviews.
 
 ## 🤯 Explanation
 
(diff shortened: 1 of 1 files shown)


docs/feature-requests/006-pull-request-reviewer-flow.md
```diff
@@ -0,0 +1,156 @@
...
+3. [Click on a "add comment" icon on the gutter](#add-comment-gutter-icon)
+
+#### 2. Continue a review
+
+Once a pending review has been started, user can add more comments to it by:
+1. [Responding to an existing review thread](#responding-to-a-comment-thread)
+2. [Clicking "add comment" icon on the gutter](#add-comment-gutter-icon)
+
+#### 3. Submit a review
+The only way to submit a review within Atom is by using the ["Submit review" button in the Pending Review tab](#summary-section). After publishing the review, the Pending Review tab will be destroyed. User will be led back to the All Reviews tab, which will immediately reflect the just published review.
```
smashwilson: Hooray for Relay optimistic rendering support :tada:

docs/feature-requests/006-pull-request-reviewer-flow.md
```diff
@@ -0,0 +1,156 @@
...
+
+### "All Reviews" tab
+This tab shows all review summaries and review comments, including the ones that are part of a _pending review_ that has not been submitted yet.
+
+#### Header
+| current header | proposed header | proposed header with pending review |
+|---|---| --- |
+|<img width="380" alt="old header" src="https://user-images.githubusercontent.com/6842965/56446925-7f027600-62d3-11e9-8635-e17b946f3b1b.png">|<img width="380" alt="new header" src="https://user-images.githubusercontent.com/6842965/56446924-7f027600-62d3-11e9-98f5-0ea1e2826343.png">|<img width="380" alt="new header" src="https://user-images.githubusercontent.com/6842965/56447078-4911c180-62d4-11e9-8543-4935ba587b38.png">|
+
+- When there is no pending review, button reads "Start a review", clicking on which will take you to the Pending Review tab in its empty state.
```
smashwilson: Can you start a review on a pull request that isn't checked out?

I wonder if we could keep the "Checkout" button when the pull request isn't checked out, then contextually turn it into either the "Start a review" or "Resume review" buttons when the pull request is checked out.

docs/feature-requests/006-pull-request-reviewer-flow.md
```diff
@@ -0,0 +1,156 @@
...
+- When there is already a pending review, the button reads "Resume review (2)". The number is a counter of comments currently in the pending review. When adding more pending comments _within the All Reviews tab_ (more on that flow below), there should be some emphasis on the counter changing -- akin to the button on dotcom. Clicking on this button takes user to the Pending Review tab.
+
+##### Alternative: footer
+An alternative would be to add a footer bar at the bottom and place the add review button there. The caveat is that it might not be as noticeable as the header, and "start a new review" is an arguably more important action than "checkout".
+
+#### Responding to a comment thread
+![responding to a comment thread](https://user-images.githubusercontent.com/6842965/56689748-cdd05700-66a9-11e9-90e8-266c69cbc589.png)
+
+User can respond to a comment thread by adding a single line comment (current implementation) or starting a new review. The two buttons should only show up when the comment textbox is in focus, or is _not_ empty.
```
smashwilson: :+1:

We can fiddle with the "in focus or not empty" behavior once we have an implementation to play with. I'm remembering that we had to scrap the "comment editor that expands when focused" because it made the surrounding UI elements jolt around.

It also feels a little awkward to have two rows of buttons right next to one another. Not sure what else we could do there, though 🤔

docs/feature-requests/006-pull-request-reviewer-flow.md
```diff
@@ -0,0 +1,156 @@
...
+The summary section of the Pending Review tab is sticky (although still collapsible), so it stays within view regardless of how long the list of comments below it is. The icon on the left indicates the type of review, which can be selected in the dropdown underneath the text box. The button to submit review will be disabled a review type has not been chosen from the dropdown menu.
+
+
+#### Comments section
+
+![pending review comments](https://user-images.githubusercontent.com/6842965/56699828-31b44900-66c5-11e9-948c-a5c03215e5d8.png)
+
+The comments section of the Pending Review tab looks very similar to that of the All Reviews tab, except that the progress bar is replaced by a small comment counter on top of the whole section.
+
+**Empty State** of this section should contain a picture tutorial of how to add a comment via gutter icon.
```
smashwilson: 💡 Random idea: maybe we could change the empty state of this tab to a list of the files changed by the pull request, so you can quickly navigate to the bits of the codebase that are covered by the diff and eligible to be commented on?

docs/feature-requests/006-pull-request-reviewer-flow.md
```diff
@@ -0,0 +1,156 @@
...
+
+A new comment block can appear in either All Reviews tab or Pending Review tab, depending on the scenarios covered in [the section below](#add-comment-gutter-icon). When in focus, a new comment block always has a glowing border to emphasize itself. If there is already a pending review, there should only be one `btn-primary` button that reads "Comment".
+
+
+### "Add comment" gutter icon
+
+A user can start a review or add a comment to an existing pending review by clicking on the "add comment" icon which shows up on hover over the gutter of:
+
+1. MultiFilePatch view within Files tab in `PullRequestDetailView`
+2. an editor *if on a checked out PR branch*
```
smashwilson: The big question left for me: we can only create comments in regions _modified by the pull request_. How do we want to signal which areas are able to be commented upon? Can we give users a quick way to navigate from one to another, so they don't get "lost" trying to find another place to comment on?

docs/feature-requests/006-pull-request-reviewer-flow.md
```diff
@@ -0,0 +1,156 @@
...
+
+The flow of starting a review or adding a comment from the gutter varies a bit depending on the state of reviews:
+
+* If there is no reviews at all
```
smashwilson:
```suggestion
* If there are no reviews at all
```

(Tiny grammar nit)

docs/feature-requests/006-pull-request-reviewer-flow.md
```diff
@@ -0,0 +1,156 @@
...
+  1. User clicks on "add comment" icon in gutter
```
smashwilson: Under what circumstances do we show the gutter that will contain this icon?

* Always? Could be distracting, especially if there are other gutters added by linters and such.
* When the current branch corresponds to a checked-out pull request?
* When a pending review is available for the current pull request?

smashwilson (commented):
Left some initial thoughts based on the writeup 😄 Looking pretty promising 👍 

I'm wondering a little about the complexity of adding a distinct additional tab for pending reviews. I like keeping them all in one place, separate from the existing reviews, but I do worry about increasing the number of navigation steps needed to reach it; GitHub tab, reviews tab, pending review tab. Can we add a "pending" section to a single reviews tab, instead? It would likely feel pretty crowded, so I'm not sure.

docs/feature-requests/006-pull-request-reviewer-flow.md
```diff
@@ -0,0 +1,156 @@
...
+2. [Clicking "add comment" icon on the gutter](#add-comment-gutter-icon)
+
+#### 3. Submit a review
+The only way to submit a review within Atom is by using the ["Submit review" button in the Pending Review tab](#summary-section). After publishing the review, the Pending Review tab will be destroyed. User will be led back to the All Reviews tab, which will immediately reflect the just published review.
+
+## 🤯 Components Explanation
+
+
+### "All Reviews" tab
+This tab shows all review summaries and review comments, including the ones that are part of a _pending review_ that has not been submitted yet.
```
kuychaco: Hmm... since the pending review will be in the "All Reviews" tab, is it necessary to have a separate "Pending Review" tab?

Quite possibly. I haven't thought too deeply about it. Just wanted to throw out the question.

docs/feature-requests/006-pull-request-reviewer-flow.md
```diff
@@ -0,0 +1,156 @@
...
+#### Responding to a comment thread
+![responding to a comment thread](https://user-images.githubusercontent.com/6842965/56689748-cdd05700-66a9-11e9-90e8-266c69cbc589.png)
+
+User can respond to a comment thread by adding a single line comment (current implementation) or starting a new review. The two buttons should only show up when the comment textbox is in focus, or is _not_ empty.
+
+When there is already an existing pending review, there should only be **one** `btn-primary` button that reads "Comment".
+
+#### Pending comments
+
+![pending comment](https://user-images.githubusercontent.com/6842965/56692893-2ce59a00-66b1-11e9-81cc-bc7956bc8bec.png)
```
kuychaco: Oooh interesting. I kinda like this placement, and I like the arrow that indicates it'll take you to somewhere else if you click it. 

Minor UI question...

I think our current implementation has the badge on the top of the comment next to the author name. Should we consider keeping it up there? Or do we want to make it more pronounced like it is here?

docs/feature-requests/006-pull-request-reviewer-flow.md
```diff
@@ -0,0 +1,156 @@
...
+
+#### Header (or the alternate footer)
+The header looks very similar to the one of All Reviews tab, with the exception that the primary button now reads "See all reviews", and will send users back to the All Reviews tab.
+
+
+#### Summary section
+
+![pending review summary](https://user-images.githubusercontent.com/6842965/56699584-23196200-66c4-11e9-94a4-193c9d662bb3.png)
+
+The summary section of the Pending Review tab is sticky (although still collapsible), so it stays within view regardless of how long the list of comments below it is. The icon on the left indicates the type of review, which can be selected in the dropdown underneath the text box. The button to submit review will be disabled a review type has not been chosen from the dropdown menu.
```
kuychaco: Does this imply we won't have a default? Looks like dotcom has "Comment" as default (at least for me)

docs/feature-requests/006-pull-request-reviewer-flow.md
kuychaco: Yeah, I was wondering the same thing. 

> we can only create comments in regions modified by the pull request

... plus a few adjacent context lines.

While this would be pretty cool, my gut is to first ship with comment creation on the `PullRequestDetailView` and leave in-editor comment creation out of scope since it has these extra complexities. 

If you think about the reviewer flow, the diff is usually what's being looked at when there is a need to create a comment. Jumping to the file could be helpful for extra context, but I think it's unlikely that a user will be in an editor first and realize they want to make a comment. And we can always try adding this in after the initial ship.

kuychaco (commented):
Thanks for tackling this @vanessayuenn! It's super clear and thorough ✨ 

Left a few questions to ponder, all pretty minor

docs/feature-requests/006-pull-request-reviewer-flow.md
vanessayuenn: Yup, definitely didn't realize you can only comment on the modified regions! That changed things considerably (in a good way) -- think it makes sense to scope it down to only allowing comments to be made on `PullRequestDetailView` like @kuychacho suggested, at least for the first iteration.

docs/feature-requests/006-pull-request-reviewer-flow.md
vanessayuenn: Oh I like this :bulb:! Although if we are only letting user comment on files changed in PR, maybe we can just add a way to navigate to that tab easily? Also I think this idea and the graphical tutorial can co-exist in the empty state, and in fact compliment each other.

docs/feature-requests/006-pull-request-reviewer-flow.md
vanessayuenn: I would say always, but only within files changed tab.

vanessayuenn:
@smashwilson:
>I'm wondering a little about the complexity of adding a distinct additional tab for pending reviews. 

Another thing I just thought of is that while a user can only have one pending review per PR, it is entirely possible for a user to have a pending review for _multiple PRs_, which would then result in 2 review tabs per PR -- that can get confusing really fast. 

I started exploring the option of having "sub-tabs" within a review tab 👇. That way we can keep all reviews (pending and published) for a PR in one place, with a simpler navigation model.

image

docs/feature-requests/006-pull-request-reviewer-flow.md
vanessayuenn: :thinking: It probably makes sense to allow comments on PR without checking it out, just like you would on dotcom.

docs/feature-requests/006-pull-request-reviewer-flow.md
vanessayuenn: Hrmm.. I can see the value in both having "Comment" as a preseleted default, and making users explicitly choose the type of review they're making. I think I am indifferent? Do you have a preference for one?

docs/feature-requests/006-pull-request-reviewer-flow.md
vanessayuenn: My major concerns with keeping both published comments and pending comments in the same space are that 1) it would get too crowded, and 2) there would be no obvious hierarchy in the information presented. 

I would argue that when a user is in the middle of reviewing a PR (i.e. have a review pending), the most relevant information to them would be their pending comments, as well as the summary textbox that they need to fill out before publishing a review. I couldn't figure out a way to fit all this along with other published comments, while preserving the hierarchy of importance.

docs/feature-requests/006-pull-request-reviewer-flow.md
vanessayuenn: Hrmm yea honestly I am 50/50 on this one. I am open to fiddle with it once we are at implementation stage, and then decide which treatment has better UX.

docs/feature-requests/006-pull-request-reviewer-flow.md
```diff
@@ -0,0 +1,137 @@
...
+## 🤯 Components Explanation
+
+
+### "All Reviews" tab
+This tab shows all review summaries and review comments, including the ones that are part of a _pending review_ that has not been submitted yet.
+
+#### Header tabs
+| header with no pending review | header with pending review |
+|---|---|
+|<img width="380" alt="new header" src="https://user-images.githubusercontent.com/6842965/57045364-57989b00-6c3b-11e9-8e95-309a17441d98.png">|<img width="380" alt="new header" src="https://user-images.githubusercontent.com/6842965/57045356-54051400-6c3b-11e9-8918-1ba043163c4d.png">|
```
smashwilson: :100: I like how this looks a lot :smile:

docs/feature-requests/006-pull-request-reviewer-flow.md
smashwilson: Yeah - the existing "pending" badge is at the top. But it also has no click action. I think being actionable would justify moving it below though. 🤔

docs/feature-requests/006-pull-request-reviewer-flow.md
smashwilson: Preselecting it would reduce the interactions required to submit a comment review, but increase the number of users who submit one by accident or without thinking. I'm kind of indifferent too, but I think slightly on the side of not having a default and making the user make a conscious choice... ?

smashwilson (approved):
👍 I think we've got enough to get started :smile:

PR #2104, opened 2019-04-29
Get author info tooltips working again

smashwilson:
**Please be sure to read the [contributor's guide to the GitHub package](https://github.com/atom/github/blob/master/CONTRIBUTING.md) before submitting any pull requests.**

### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* Suggestion: You can use checklists to keep track of progress for the sections on metrics, tests, documentation, and user research.

### Description of the Change

Fixes #2094 by updating the GraphQL query and component props to reflect the renaming of `Organization#membersWithRole` to `Organization#members`.

### Screenshot

_N/A_

### Alternate Designs

UserMentionTooltipItem is one of the old-style tooltips. I could modernize it, but I opted for a surgical fix here.

### Benefits

User mention tooltips will work again.

### Possible Drawbacks

_N/A_

### Applicable Issues

Fixes #2094

### Metrics

_N/A_

### Tests

I tested it out with the PR in the issue screenshot :smile:

### Documentation

_N/A_

### Release Notes

* User mention tooltips correctly fetch data again.

### User Experience Research (Optional)

_N/A_

Issue #2094, opened 2019-04-24
Author info tooltip card is broken


kuychaco:
![author-card-broken](https://user-images.githubusercontent.com/7910250/56697900-ec801f00-66a4-11e9-9ec1-a8381beb8f86.gif)

womp womp 😞 

Expected: A card should show up on hover that shows information about the user, much like the card that shows information for the PR# being hovered over

On master at 4ef8aacec

PR #2107, opened 2019-04-29
Exclude deprecated fields from the fetched schema

smashwilson:

### Requirements

### Description of the Change

Inline the introspectionQuery from [the graphql package](https://github.com/graphql/graphql-js/blob/master/src/utilities/introspectionQuery.js#L20) and change the `includeDeprecation:` arguments to `false`. This will exclude deprecated fields from the schema we give to `relay-compiler`, which will in turn give us early notice of us depending on fields which are due to be removed.

### Screenshot/Gif

_N/A_

### Alternate Designs

It might be worth making a pull request to the GraphQL package itself to give `getIntrospectionQuery()` an option to include deprecated fields or not. We'd need to wait for it to be included in a release and for Relay to catch up before we could take advantage of it, though.

### Benefits

Early notice of GraphQL schema changes that will affect us. This would help prevent regressions like #2094 from occurring.

### Possible Drawbacks

Well... we can't actually merge the schema changes until #2104 is merged 🙃 

### Applicable Issues

#2094, #2104.

### Metrics

_N/A_

### Tests

_N/A_

### Documentation

_N/A_

### Release Notes

_N/A_

_N/A_

### DIFF
--- a/script/fetch-schema
+++ b/script/fetch-schema
@@ -3,12 +3,99 @@ const fs = require('fs');
 const path = require('path');
 
 const fetch = require('node-fetch');
-const {
-  buildClientSchema,
-  introspectionQuery,
-  printSchema,
-} = require('graphql/utilities');
+const {buildClientSchema, printSchema} = require('graphql/utilities');
 const schemaPath = path.resolve(__dirname, '..', 'graphql', 'schema');
+
+const introspectionQuery = `
+  query IntrospectionQuery {
+    __schema {
+      queryType { name }
+      mutationType { name }
+      subscriptionType { name }
+      types {
+        ...FullType
+      }
+      directives {
+        name
+        description
+        locations
+        args {
+          ...InputValue
+        }
+      }
+    }
+  }
+  fragment FullType on __Type {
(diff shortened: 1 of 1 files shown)


smashwilson:
> I guess it's technically somewhat redundant to #2108, but still seems like a good change!

Yeah, kind of? There's likely a way to share code between them once they're both merged, even.

> @smashwilson you indicated N/A for testing in the PR description. It's probably worth doing a quick manual check on the command line just for good measure...

Oh, right, I did check it out manually - I didn't commit the schema changes it made because they broke relay-compiler on account of #2104 not being merged yet. :smile:

> I'm also wondering if there is a way to programmatically modify the includeDeprecated arguments in graphql/utilities's introspectionQuery.

Oh, interesting idea... I'd thought about sending a pull request to the GraphQL package to add a `deprecated` parameter to `getIntrospectionQuery()`, but that stalled a bit when I found out I'd have to do the Facebook CLA thing to contribute. I think we have access to an intermediate representation of the query that we can likely walk and modify... ?

PR #2108, opened 2019-04-30
GitHub Actions to automate our GraphQL schema management

smashwilson:

### Requirements

### Description of the Change

Create a [GitHub Actions](https://developer.github.com/actions/) to manage our local copy of the GitHub GraphQL schema in the least intrusive way I can think of.

It'll fetch the latest schema from GitHub and write it into the working directory. It'll then run the relay-compiler with the updated schema. If no generated Relay files have changed, the new schema will be committed and pushed directly onto `master`. If there _have_ been changes, the action will commit the schema and the Relay changes to a new branch, push it, open a pull request, and apply the "schema changed" label to it. (No pull request will be created if there's an existing open pull request with that label.)

### Screenshot

_N/A_

### Alternate Designs

We could have done this with our existing CI. Too late, this is more fun.

I had a second action I was working on that would watch as status checks and commit statuses were set on the head ref of each pull request, with the intent that it would automatically merge it if they were all green. The problem is that it would create a new check suite on every PR for each individual status that was reported, as it came in... which spammed them quite a bit.

### Benefits

Ideally, this will give us an early warning for upcoming GraphQL deprecations that will cause issues for our package; the "schema-up" action will fail on its relay-compiler step if the schema has changed in an incompatible way. The action will also exclude deprecated fields, so we should see the notice before the schema change is made in production.

### Possible Drawbacks

It's hard to actually validate these until they run on real webhook triggers. They might be totally broken right now. I'll just have to fix any issues as they arise?

### Applicable Issues

Closes #2116, which contains all of the test runs.

### Metrics

_N/A_

### Tests

image

### Documentation

I've got READMEs in the source of the `./action` subdirectory.

### Release Notes

_N/A_

_N/A_


smashwilson:
I don't think there's any way for me to run this manually before we merge it into master. I'd feel better if someone other than me had a quick look at my source to find obvious bugs before I start running it, though 🙇

kuychaco:
Awesome, thanks so much for tackling this!

In the PR description it says:
> I've got READMEs in the source of both ./action directories.

I only see one directory `schema-up`. I'm assuming this comment is stale?

smashwilson:
Hah! Yes. I originally had another action that would automatically merge the schema-change pull requests if all check suites and commit statuses were green, but it was _super_ noisy - it added a check suite for every suite and status as it arrived. Edit'd 🙇

actions/schema-up/index.js
```diff
@@ -0,0 +1,111 @@
...
+  if (hasRelayChanges === 0 && !relayFailed) {
+    tools.log.info('Generated relay files are unchanged.');
+    await tools.runInWorkspace('git', ['push']);
+    tools.exit.success('Schema is up to date on master.');
+  }
+
+  tools.log.info('Checking for unmerged schema update pull requests.');
+  const {data: openSchemaUpdatePRs} = await tools.github.graphql(`
+    query OpenPullRequests($owner: String!, $repo: String!) {
+      repository(owner: $owner, name: repo) {
```
kuychaco:
```suggestion
      repository(owner: $owner, name: $repo) {
```

actions/schema-up/index.js
```diff
@@ -0,0 +1,111 @@
...
+        pullRequests(first: 1, states: [OPEN], labels: ["schema update"]) {
```
kuychaco: Readability suggestion: Encapsulate the label information in an object. That way the relationship between the label name and id is explicit, and if we ever use a different label it'll be a more clear and straightforward to update the code.

```
const schemaUpdateLabel = {
  name: 'schema update',
  id: 'MDU6TGFiZWwxMzQyMzM1MjQ2'
}
```

```suggestion
        pullRequests(first: 1, states: [OPEN], labels: [schemaUpdateLabel.name]) {
```

along with corresponding change below for the label id

actions/schema-up/index.js
```diff
@@ -0,0 +1,111 @@
...
+      addLabelsToLabelable(input: {
+        labelableId: $id,
+        labelIds: []
+      }) {
+        labelable {
+          id
+        }
+      }
+    }
+  `, {id: createdPullRequest.id, labelIDs: ['MDU6TGFiZWwxMzQyMzM1MjQ2']});
```
kuychaco: See comment above for `schemaUpdateLabel`...

```suggestion
  `, {id: createdPullRequest.id, labelIDs: [schemaUpdateLabel.id]});
```

actions/schema-up/index.js
```diff
@@ -0,0 +1,111 @@
...
+    body += ' The modified files have been committed to this branch and pushed. ';
+    body += 'If all of the tests pass in CI, merge with confidence :zap:';
+  } else {
+    body += ' `relay-compiler` failed with the following output:\n\n```\n';
+    body += relayOutput;
+    body += '\n```\n\nCheck out this branch to fix things so we don\'t break.';
+  }
+
+  const {data: createPR} = await tools.github.graphql(`
+    mutation CreatePullRequest($owner: String!, $repo: String!, $body: String!, $headRefName: String!) {
```
kuychaco: Looks like the required input for `createPullRequest` is:

GraphQL_API_Explorer___GitHub_Developer_Guide

So we also need `repositoryId` and `baseRefName`.

```suggestion
    mutation CreatePullRequest($repositoryId: ID!, $baseRefName: String!, $body: String!, $headRefName: String!) {
```

actions/schema-up/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+  const createdPullRequest = createPR.createPullRequest.pullRequest;
+  tools.log.info(
+    `Pull request #${createdPullRequest.number} has been opened with the changes from this schema upgrade.`);
+
+  await tools.github.graphql(`
+    mutation LabelPullRequest($id: ID!, $labelIDs: [ID!]!) {
+      addLabelsToLabelable(input: {
+        labelableId: $id,
+        labelIds: []
```
kuychaco:
```suggestion
        labelIds: $labelIDs
```

actions/schema-up/index.js
```diff
@@ -0,0 +1,111 @@
...
+      }) {
+        labelable {
+          id
+        }
+      }
+    }
+  `, {id: createdPullRequest.id, labelIDs: ['MDU6TGFiZWwxMzQyMzM1MjQ2']});
+}, {
```
kuychaco: Is it worth checking/logging that the label has been applied successfully? Maybe not a big deal...

kuychaco (changes requested):
This looks great! Excited to get our first GitHub Action in place, and be guarded against more regressions due to graphql API deprecations!

Thanks for putting this together @smashwilson ✨ 

Caught a few things that warrant attention :). 

I'd be curious if we could try out testing this prior to merging by temporarily changing the Actions trigger to be on push. We can always push empty commits if we need to. That could be a more efficient way of testing, and it'd be nice if all the changes were contained in a single PR for future reference.

smashwilson:
#2116 contains all of the testing output for this, with a minimal diff on the actual Actions code:

```diff
smashwilson @ hubtop ~/src/atom/github (aw/schema-update-action-test=)
$ git diff aw/schema-update-action -- actions/ .github/
diff --git a/.github/main.workflow b/.github/main.workflow
index 490262a3b..bb567a330 100644
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -1,6 +1,8 @@
 workflow "GraphQL schema update" {
   // Every Monday at 1am.
-  on = "schedule(0 1 * * 1)"
+  // on = "schedule(0 1 * * 1)"
+
+  on = "push"
   resolves = "Update schema"
 }

diff --git a/actions/schema-up/index.js b/actions/schema-up/index.js
index db4c34f2b..6af28f367 100644
--- a/actions/schema-up/index.js
+++ b/actions/schema-up/index.js
@@ -87,6 +87,7 @@ The GraphQL schema has been automatically updated and \`relay-compiler\` has bee
     body += relayOutput;
     body += '\n```\n\nCheck out this branch to fix things so we don\'t break.';
   }
+  tools.log.info('Pull request body:\n%s', body);

   const createPullRequestMutation = await tools.github.graphql(`
     mutation createPullRequestMutation($repositoryId: ID!, $headRefName: String!, $body: String!) {
@@ -127,6 +128,7 @@ The GraphQL schema has been automatically updated and \`relay-compiler\` has bee
   tools.exit.success(
     `Pull request #${createdPullRequest.number} has been opened and labelled for this schema upgrade.`,
   );
+  tools.exit.success('Not creating pull request to reduce testing noise');
 }, {
   secrets: ['GITHUB_TOKEN'],
 });
```

actions/schema-up/index.js
```diff
@@ -0,0 +1,132 @@
...
+    `Pull request #${createdPullRequest.number} has been opened with the changes from this schema upgrade.`,
+  );
+
+  await tools.github.graphql(`
+    mutation labelPullRequestMutation($id: ID!, $labelIDs: [ID!]!) {
+      addLabelsToLabelable(input: {
+        labelableId: $id,
+        labelIds: $labelIDs
+      }) {
+        clientMutationId
```
kuychaco: Hmm it's a bit annoying that you're forced to ask for return data, even if there's no use for it...

kuychaco (approved):
Awesome work! 🚢 :shipit:

smashwilson:
Thanks for the review 🙇 I guess we'll see it work live on Monday?

Issue #2078, opened 2019-04-16
reconsider grouping of review comments 


annthurium:
we got some feedback from a user, via twitter:

https://twitter.com/flaximus_prime/status/1118210262446821376

"When there are many comments in a review, and some are resolved, it’s hard for me to scan that list and pick out those not resolved. They’re interspersed and it just doesn’t work for my eyes/brain. I’d love a way to hide the resolved ones or group them."

![image](https://user-images.githubusercontent.com/3781742/56233797-bc97a280-6038-11e9-8bde-72d1b8313e4d.png)

We tried to validate some hypotheses about comment ordering / grouping during our initial user research, but the results were inconclusive.  It could be worth revisiting comment grouping/ordering, now that we have something shipped.

vanessayuenn:
Closed via #2117

PR #2117, opened 2019-05-02
Group resolved comments

vanessayuenn:
### Screenshot/Gif

Kapture 2019-04-29 at 18 08 38

### Alternate Designs

I experimented with using filter toggle but the user experience feels clunky, and with the current design, there isn't really a natural placement for the toggle.

### Benefits

Resolved comments are moved out of the way, further encouraging the use of review comments as a to-do list behaviour model. 

### Possible Drawbacks

It might confuse users slightly when the resolved comment gets moved to another location. The highlighting should mitigate this confusion, but the problem can still exist for cases where there are many comments. However, the current implementation of collapsing comments immediately after resolving is probably just as confusing, so I don't see this feature making the experience worse. 

### Applicable Issues

#2078 

### Tests

TBD: still working on them!

### Release Notes

Group resolved review comments into its own list.

### DIFF
--- a/lib/controllers/reviews-controller.js
+++ b/lib/controllers/reviews-controller.js
@@ -81,16 +81,14 @@ export class BareReviewsController extends React.Component {
       threadIDsOpen: new Set(
         this.props.initThreadID ? [this.props.initThreadID] : [],
       ),
-      highlightedThreadIDs: new Set(
-        this.props.initThreadID ? [this.props.initThreadID] : [],
-      ),
+      highlightedThreadIDs: new Set(),
     };
   }
 
   componentDidMount() {
     const {scrollToThreadID} = this.state;
     if (scrollToThreadID) {
-      setTimeout(() => this.setState({scrollToThreadID: null}), FLASH_DELAY);
+      this.highlightThread(scrollToThreadID);
     }
   }
 
@@ -99,9 +97,8 @@ export class BareReviewsController extends React.Component {
     if (initThreadID && initThreadID !== prevProps.initThreadID) {
       this.setState(prev => {
         prev.threadIDsOpen.add(initThreadID);
+        this.highlightThread(initThreadID);
         return {commentSectionOpen: true, scrollToThreadID: initThreadID};
-      }, () => {
-        setTimeout(() => this.setState({scrollToThreadID: null}), FLASH_DELAY);
       });
     }
   }
--- a/lib/controllers/reviews-controller.js
+++ b/lib/controllers/reviews-controller.js
@@ -239,6 +239,9 @@ export class BareReviewsController extends React.Component {
     }, () => {
       setTimeout(() => this.setState(state => {
         state.highlightedThreadIDs.delete(threadID);
+        if (state.scrollToThreadID === threadID) {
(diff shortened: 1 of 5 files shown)


lib/controllers/reviews-controller.js
```diff
@@ -96,9 +97,8 @@ export class BareReviewsController extends React.Component {
     if (initThreadID && initThreadID !== prevProps.initThreadID) {
       this.setState(prev => {
         prev.threadIDsOpen.add(initThreadID);
+        this.highlightThread(initThreadID);
         return {commentSectionOpen: true, scrollToThreadID: initThreadID};
-      }, () => {
-        setTimeout(() => this.setState({scrollToThreadID: null}), FLASH_DELAY);
```
smashwilson: I wish we had a way to collapse the `setState` updater functions from here and `highlightThread` into a single, atomic transaction. I'm pretty sure that's how it's working anyway and this is much cleaner but still 😛

lib/views/reviews-view.js
```diff
@@ -110,26 +113,14 @@ export default class ReviewsView extends React.Component {
...
     const {scrollToThreadID} = this.props;
     if (scrollToThreadID) {
-      const threadHolder = this.threadHolders.get(scrollToThreadID);
-      if (threadHolder) {
-        threadHolder.map(element => {
-          element.scrollIntoViewIfNeeded();
-          return null; // shh, eslint
-        });
-      }
+      this.scrollToThread(scrollToThreadID);
```
smashwilson: 👍

lib/views/reviews-view.js
```diff
@@ -308,7 +299,8 @@ export default class ReviewsView extends React.Component {
       return null;
     }
 
-    const resolvedThreads = commentThreads.filter(pair => pair.thread.isResolved).length;
```
smashwilson: Hah, should have called this `resolvedThreadCount` or something before. 👍

lib/views/reviews-view.js
```diff
@@ -644,4 +648,23 @@ export default class ReviewsView extends React.Component {
...
     return {lineNumber, positionText};
   }
+
+  /* istanbul ignore next */
+  scrollToThread(threadID) {
+    const threadHolder = this.threadHolders.get(threadID);
+    if (threadHolder) {
+      threadHolder.map(element => {
+        element.scrollIntoViewIfNeeded();
+        return null; // shh, eslint
```
smashwilson: I really need to make a `RefHolder::forEach` method to make these cases nicer...

lib/views/reviews-view.js
```diff
@@ -644,4 +648,23 @@ export default class ReviewsView extends React.Component {
...
+      });
+    }
+  }
+
+  async resolveUnresolveThread(thread) {
```
smashwilson: We keep hitting situations where we need to name methods that toggle something without having a good word for the toggling action. There isn't really a good alternative for this:

* `resolveOrUnresolveThread()`
* `toggleThreadResolution()`
* ... ?

smashwilson (approved):
👍 LGTM!

PR #1934, opened 2019-02-01
Feature request: Identity and Context Tiles

smashwilson:
Feature request for potential a way to improve awareness of things like "what repository am I actually working in right now" and "what account am I using."

PR #2128, opened 2019-05-07
Improve error handling when offline

### LINKED ISSUE #1844: GitHub tab crashes upon lost of internet connection
### Description

GitHub tab does not handle internet connection dropping well. 

### Steps to Reproduce

1. Open a project in Atom that has a GitHub repo with opened PRs
2. Open GitHub tab
2. Click on any PR from the list
3. Opens Network tab in Chrome devtool and check the "Offline" box (or just disconnect from internet in any way)
4. Click on another PR from GitHub tab
5. Watch the world crash and burn

image

**Expected behavior:**

Gracefully fail to fetch any additional info from github.com, and display a nice message informing users they have lost internet connection.

**Actual behavior:**

The whole package crashes with no way to recover beside a full reload.

**Reproduces how often:**

100%

### Versions

All versions

smashwilson:

### Requirements

### Description of the Change

When the network is unavailable, the GitHub package behaves poorly:

* When starting or reloading Atom, the "log in to GitHub" view is shown from each Relay-backed view, which, of course, doesn't work.
* If the network becomes unavailable after Atom is already open, the next attempted network operation brings down the package hard with an uncaught exception.
* Relay actions that users explicitly initiate (like clicking on a refresh button, or triggering a mutation) uniformly create Atom error notifications, with customized messaging for the offline case.

I'm introducing:

- [x] A special `<QueryErrorView>` case that recognizes network errors and offers to open the dev tools for more investigation. (Unfortunately, the errors thrown by `fetch` are `TypeError` instances with little information and a message of "fetch failed.")
- [x] Create an alternate blank-slate case to show in place of the `<GithubLoginView>` when the network is unavailable.
- [x] Ensure that further changes in network state are detected when offline so we can re-render again.
- [x] Separate messaging for "offline" from "non-200, non-401 statuses"
- [x] Handle the issueish tiles in the GitHub tab
- [x] Preserve existing data and notify instead if a refetch fails, so you don't lose cached data.

### Screenshots

When a user attempts to open a new item that uses Relay data (an issueish detail item, the reviews tab) while the network is offline, including items created during window state deserialization (like a reload):

offline message

When a user attempts to click a refresh button or perform a mutation while offline:

Screen Shot 2019-05-09 at 3 22 52 PM

When the network dies in between the git fetch completing and the Relay refetch triggering somehow:

offline-tile

### Alternate Designs

_N/A_

### Benefits

Users on spotty wi-fi or airplanes can have a gracefully degraded Atom experience and informative error states.

### Possible Drawbacks

The errors thrown by the `fetch` API are (deliberately, I believe) very vague. It's possible that we could report that a user is offline when they have misconfigured DNS or proxy settings, for example. The underlying OS network error is shown in the developer tools, but we have no way to read it.

### Applicable Issues

Fixes #1844.

### Metrics

_N/A_

### Tests

I did manual testing both by toggling network connectivity in the dev tools and by actually disabling and enabling wi-fi.

Interestingly, the `'online'` event that I'm using to auto-retry from an OfflineView doesn't seem to fire the _first_ time you uncheck the "Offline" button in the dev tools, but it does every subsequent time. The event fired when I expected it to when the wi-fi was actually restored so I didn't investigate further.

### Documentation

_N/A_

### Release Notes

* The GitHub package reports clearer errors when no network connection is available.

_N/A_

### DIFF
--- a/lib/views/error-view.js
+++ b/lib/views/error-view.js
@@ -25,13 +25,13 @@ export default class ErrorView extends React.Component {
           <h1 className="github-Message-title">{this.props.title}</h1>
           {this.props.descriptions.map(this.renderDescription)}
           <div className="github-Message-action">
+            {this.props.retry && (
+              <button className="github-Message-button btn btn-primary" onClick={this.props.retry}>Try Again</button>
+            )}
             {this.props.openDevTools && (
               <button className="github-Message-button btn btn-devtools" onClick={this.props.openDevTools}>
               Open dev tools
               </button>
-            )}
-            {this.props.retry && (
-              <button className="github-Message-button btn btn-primary" onClick={this.props.retry}>Try Again</button>
             )}
             {this.props.logout && (
               <button className="github-Message-button btn btn-logout" onClick={this.props.logout}>Logout</button>
--- a/lib/views/error-view.js
+++ b/lib/views/error-view.js
@@ -9,6 +9,7 @@ export default class ErrorView extends React.Component {
 
     retry: PropTypes.func,
     logout: PropTypes.func,
+    openDevTools: PropTypes.func,
   }
 
   static defaultProps = {
@@ -24,6 +25,11 @@ export default class ErrorView extends React.Component {
           <h1 className="github-Message-title">{this.props.title}</h1>
           {this.props.descriptions.map(this.renderDescription)}
           <div className="github-Message-action">
+            {this.props.openDevTools && (
+              <button className="github-Message-button btn btn-devtools" onClick={this.props.openDevTools}>
+              Open dev tools
+              </button>
+            )}
             {this.props.retry && (
               <button className="github-Message-button btn btn-primary" onClick={this.props.retry}>Try Again</button>
             )}
--- a/lib/views/error-view.js
+++ b/lib/views/error-view.js
@@ -1,8 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import {autobind} from '../helpers';
-
 export default class ErrorView extends React.Component {
   static propTypes = {
     title: PropTypes.string,
@@ -17,12 +15,6 @@ export default class ErrorView extends React.Component {
     title: 'Error',
     descriptions: ['An unknown error occurred'],
     preformatted: false,
-  }
-
-  constructor(props) {
-    super(props);
-
-    autobind(this, 'renderDescription');
   }
 
   render() {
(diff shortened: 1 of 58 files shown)


lib/views/error-view.js
```diff
@@ -11,6 +9,7 @@ export default class ErrorView extends React.Component {
 
     retry: PropTypes.func,
     logout: PropTypes.func,
+    openDevTools: PropTypes.func,
```
vanessayuenn: How is `openDevTools` used? I can't seem to find any instance in the codebase that passes this prop.

vanessayuenn (approved):
Thanks for tackling this! The new offline view looks good and the message is very clear 👍 . I have got a couple of questions; they're mainly for my own understanding and are not blockers.

lib/views/error-view.js
smashwilson: Oh! It isn't 😅 

I briefly added an "open devtools" button to one of the offline views because you can't see the _actual_ network error anywhere else (the Error that `fetch` rejects with doesn't include it). So I had to drill `atom.openDevTools` everywhere. I decided against it to keep things more simple, but it looks like this propType missed the purge. I'll 🔥 .

lib/views/error-view.js
```diff
@@ -35,6 +28,11 @@ export default class ErrorView extends React.Component {
             {this.props.retry && (
               <button className="github-Message-button btn btn-primary" onClick={this.props.retry}>Try Again</button>
             )}
+            {this.props.openDevTools && (
+              <button className="github-Message-button btn btn-devtools" onClick={this.props.openDevTools}>
+              Open dev tools
+              </button>
+            )}
```
smashwilson: Note to self: :fire: this too.

lib/views/query-error-tile.js
```diff
@@ -10,6 +10,7 @@ export default class QueryErrorTile extends React.Component {
         status: PropTypes.number.isRequired,
```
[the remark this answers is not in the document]
smashwilson: Short answer: not yet.

We don't actually have any uniform way to refresh the GitHub tab during _normal_ operations 😅 . It actually refreshes when you fetch or push, but that reruns all of the QueryRenderers entirely, which doesn't confuse the Accumulators (I did actually test this out to verify). If we nest this in a refetch container when we change the refresh behavior (maybe with #1934?) then yeah we'll need to pass in a well-behaved `onDidRefetch()` method. Once we do that, it might make sense to integrate some auto-retry on network change behavior here or a retry button - it depends on the UI we settle on.

PR #2195, opened 2019-07-04
Avoid stalls caused by large diffs fetched from GitHub

smashwilson:

### Requirements

### Description of the Change

Our `PullRequestPatchContainer` component fetches and parses diffs from GitHub's REST API, then makes this available to child components as a MultiFilePatch. Unfortunately, it seems that pull requests with many or large files -- such as atom/flight-manual.atom.io#541 -- can cause stalls of a full minute (😱) while these are being parsed.

Interestingly, the stalls seem to be in what-the-diff's PEG parser rather than our own `buildMultiFilePatch()` function; in the past, `buildMultiFilePatch()` was the bottleneck. As a consequence our "large diff detection" logic isn't helping.

To mitigate this:

- [x] I'm using [conditional requests](https://developer.github.com/v3/#conditional-requests) to avoid fetching and re-parsing the _same_ large diff from dotcom more than once.
- [x] We don't need the patch at all in `CommentDecorationsContainer` if the current PR has no review comments. Let's find review comments first and bail if there are none (which is the common case anyway).
- [x] Finally, I'll investigate setting an upper size bound on which diffs we attempt to parse from dotcom.
  - [x] It would be good to offer some notification of degraded functionality when we _do_ have to drop comments.
  - [x] When the patch is large enough to trigger the MultiFilePatch "large diff" gate, but not the raw diff filter, comments are unavailable, but only until the diff is expanded. Let's show a notification with a control that allows users to opt in to parsing the diff when it does have comments.
  - [x] We'll need to make sure that the `PullRequestDetailView` can gracefully handle a missing diff. Or at least not crash horribly.
  - [x] Settling on an appropriate size limit will take some experimentation.

### Screenshot/Gif

Here's what I'm doing if you open an editor on a file that has review comments, but whose patch has been omitted for size:

too-large

### Alternate Designs

_N/A_

### Benefits

Using Atom to edit files on branches corresponding to pull requests with "large" diffs will no longer cause periodic stalls.

### Possible Drawbacks

The review comment and patch fetches can happen in parallel presently because they're independent operations. By making the patch fetch depend on the result of the review comment aggregation, it'll take longer to load all of the data and display the first review comment decoration or populate the reviews tab.

### Applicable Issues

_N/A_

### Metrics

_N/A_

### Tests

* [x] Normal case: small patch, review comments.
  * [x] Ensure comment decorations are rendered appropriately.
  * [x] Ensure that the ETag and 304 response are used to prevent re-parsing the patch.
* [x] On a PR with no review comments, ensure the patch is not fetched at all.
* [x] On a PR with a large file patch and a small patch that contains review comments:
  * [x] Ensure the large file patch is filtered out and does not cause a performance stall.
  * [x] Ensure the review comments on the file with the small patch are rendered appropriately.
* [x] On a PR with a large file patch that contains review comments:
  * [x] Ensure the large file patch is filtered out and does not cause a performance stall.
  * [x] Ensure a block decoration is rendered at the front of the editor of the file with the large patch.
  * [x] Ensure that opening the ReviewsItem on the PR does not crash as a result of the missing patch.

### Documentation

_N/A_

### Release Notes

* Improved the performance of the GitHub package when working on checked-out pull request with a large diff.

_N/A_

### DIFF
--- a/lib/containers/comment-decorations-container.js
+++ b/lib/containers/comment-decorations-container.js
@@ -189,7 +189,8 @@ export default class CommentDecorationsContainer extends React.Component {
         repo={repo}
         number={currentPullRequest.number}
         endpoint={endpoint}
-        token={token}>
+        token={token}
+        largeDiffThreshold={Infinity}>
         {(patchError, patch) => this.renderWithPatch(
           {error: patchError, patch},
           {summaries, commentThreads, currentPullRequest, repoResult, endpoint, owner, repo, repoData, token},
--- a/lib/containers/comment-decorations-container.js
+++ b/lib/containers/comment-decorations-container.js
@@ -126,7 +126,7 @@ export default class CommentDecorationsContainer extends React.Component {
           environment={environment}
           query={query}
           variables={variables}
-          render={queryResult => this.renderWithGraphQLData({
+          render={queryResult => this.renderWithPullRequest({
             endpoint,
             owner: variables.headOwner,
             repo: variables.headName,
@@ -137,7 +137,7 @@ export default class CommentDecorationsContainer extends React.Component {
     );
   }
 
-  renderWithGraphQLData({error, props, endpoint, owner, repo}, {repoData, token}) {
+  renderWithPullRequest({error, props, endpoint, owner, repo}, {repoData, token}) {
     if (error) {
       // eslint-disable-next-line no-console
       console.warn(`error fetching CommentDecorationsContainer data: ${error}`);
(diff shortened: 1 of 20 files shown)


smashwilson:
cc @jasonrudolph and @nathansobo who brought this to our attention :smile:

smashwilson:
Lol okay then:

```
smashwilson @ hubtop ~/src/atom/github (aw/large-dotcom-diffs %=) 
$ curl https://api.github.com/repos/atom/flight-manual.atom.io/pulls/541 -H 'Accept: application/vnd.github.v3.diff' -o big-patch.diff
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  124M  100  124M    0     0  3021k      0  0:00:42  0:00:42 --:--:-- 3890k
smashwilson @ hubtop ~/src/atom/github (aw/large-dotcom-diffs %=) 
$ du -h big-patch.diff 
128M    big-patch.diff
```

smashwilson:
It looks like GitHub itself marks review comments as "outdated" on files with patches that grow beyond 1MB, so I'll match that threshold here.

Should be good to go 👌

jasonrudolph:
Thanks for digging into this, @smashwilson. :zap::bow:

kasun-i8 (commented):
Submit general feedback

PR #2298, opened 2019-10-07
Use GitHub actions for tests

smashwilson:
Looks like Azure DevOps doesn't publish tests as check results any more. Let's just move to GitHub Actions instead because it's _right there_.


UziTech:
@smashwilson I created a GitHub Action to install Atom and APM in the PATH for running tests in GitHub Actions. This might help get Windows tests working.

[Setup Atom](https://github.com/marketplace/actions/setup-atom)

smashwilson:
Oh, neat! Have to have a look the next time I have some time to spend here.

PR #870, opened 2017-05-25
Create or add remote repository [WIP]

jamiebuilds:
***Note: This is just an scaffold of the UI so far, none of the features have been fully implemented yet***

### Description of the Change

When there is no remote repository display this view:

screen shot 2017-05-24 at 11 36 09 pm

### Alternate Designs

Happy to go in a different direction if this design isn't what the team wants.

### Applicable Issues

- #869 Create or add remote repository

Issue #555, opened 2017-02-23
Manage remotes


smashwilson:
Provide some kind of UI to create, retrieve, update, and delete a repository's remotes.

We should especially support a workflow popular in the open-source world that uses a forked repository as a head for contributor pull requests:

1. A contributor forks the repository on GitHub
2. The contributor clones the repository to their laptop from their personal fork, adding it as a remote called `origin`.
3. The contributor adds a remote, often called `upstream`, pointing to the parent repository on GitHub.
4. The contributor pushes branches containing work to be integrated to `origin`, their personal fork.
5. To integrate changes from other contributors, the contributor pulls from the master branch on the `upstream` remote and either merges in the fetched changes or rebases on top of them.

sromocki:
I was prompted to select a remote for a branch at one point while using this integration.  I selected one and it says it's not found.  Is it possible to change the remote again after initially  selecting one?  I also tried to manually enter the PR URL although it's not able to find the PR either so I'm a bit stuck.

rsese:
From https://github.com/atom/github/issues/1679:

> One thing that I think Atom's git integration does very poorly with at the moment is remote management. I'd like to be able to interactively choose which remotes I fetch/pull/rebase from or push to, and I can't do that at the moment.
> 
> Ideally, I'd like to see a right-click menu option for "Pull from..." and "Push to..." to get prompted for the remote, and have the option for that interactive prompt to be the default action (i.e. always prompt me which remote I want for the action).

wadethestealth:
Hey @smashwilson and @rsese I totally agree with #1679. I would love to have some kind of option to switch remotes as is common in my workflow, which the only option currently is to chose a default remote when you have multiple. I would love to help contribute this feature, how is the best way to interface with your team and share my ideas.

My first idea is to just create a new atom command `github:choose-remote` which is accessed through the command palatte, and it automatically shows you your options or allows you to add a new remote through `atom-select-list`. nteract/hydrogen has something very similar to this for selecting kernels, and it works well for us.

PR #2299, opened 2019-10-08
WIP: Remote Management

wadethestealth:
### Description of the Change
Moved to WIP awaiting discussion and #2325 

- [ ] Adds the ability to change remotes for GitHub PR Reviews
- [ ] Adds a GitHub tab heading to show which remote you are on.
   - [ ] Adds a selector to swap remotes

#### Current Design:
TBD

### Alternate Designs
Up to discussion.

### Benefits

- This will add Atom support to workflows that often work with multiple remotes.

### Possible Drawbacks

Possible draw backs are the number of remotes someone could have and how we should handle that. I have made note of [visualizing lists in react](https://reactjs.org/docs/optimizing-performance.html#virtualize-long-lists) from official react docs, just in case we need to investigate this.

### Applicable Issues

#555, #870 (This seems stale, but there are some nice ideas in there that I would like to take over) and #1934 (Prototype for the UI of Remote Management)

### Metrics

N/A

### Tests

N/A

### Documentation
- Added Feature Request Doc

### Release Notes
TBD

TBD

---
cc: @simurai @smashwilson 
-----
[View rendered docs/feature-requests/007-remote-management.md](https://github.com/wadethestealth/github/blob/remote-management/docs/feature-requests/007-remote-management.md)


lib/containers/github-header-container.js
```diff
@@ -0,0 +1,104 @@
...
+    endpoint: EndpointPropType.isRequired,
+
+    avatarUrl: PropTypes.string,
+    remotes: RemoteSetPropType.isRequired,
+    currentRemoteName: PropTypes.string.isRequired,
+
+    handleRemoteSelect: PropTypes.func.isRequired,
+  }
+
+  fetchToken = loginModel => {
```
wadethestealth: you should note that this is duplicate code from `remote-controller.js`, and I believe that this duplicate logic should be separated to a component that handles the login/endpoint logic for GraphQL however since it is just token verification functions and one component I was unsure what to call it some ideas I had were some combination of `github-` + (`login-` or `token-`) + (`controller` or `container`)

wadethestealth:
@smashwilson I know we briefly discussed putting remotes inside the project header of #2308, however, I was just wondering what your opinion was on a potential workflow issue.
- If I mostly work and commit on a personal fork, but then open PR's on a remote upstream, it could become awkward to switch remotes on the header just when I want to view a PR and then switch back when I see something I want to change. 
   - Once a review is open, it doesn't truly matter anymore which context I am in because the review can always stay loaded. However, I wonder if it could be beneficial to have separate remote selection for git and github. Thoughts?

wadethestealth:
Was kinda taken over by the project management pr. This feature is still not available, but this code requires a complete restructure as it has drifted far from master.

PR #2308, opened 2019-10-16
Project Management

wadethestealth:
### Description of the Change
This is a functional realization of the project selection prototype in #1934. For those unfamiliar with the PR, this would allow you to view the base name of your currently active project folder, as well as actively change your active project folder within the Git and GitHub panels.

**Note:** This is a **minimum viable product** of a small portion of #1934, and the rest is meant to implemented in separate PR's, which will be updated here as they arrive.

*Functional Realization of Avatar PR: N/A*
*Functional Realization of PR Refresh Button PR: N/A*

#### Achieved:
- [x] Add `Tab Header View` to Git Panel Tab
- [x] Add `Tab Header View` to GitHub Panel Tab
- [x] Integrate new project context priority queue
- [x] Allow users to switch active projects in the `Tab Header View`
- See tests for test coverage achievements.

### Screenshot/Gif

project-management

### Alternate Designs

N/A

### Benefits

- Support for switching current active projects
- Better support for multi-project workflows

### Possible Drawbacks

- Loss of Vertical UI Space

### Applicable Issues

#1934 

### Tests

- [x] Added Tab Header View Unit Tests (100% coverage)
- [x] Added Tab Header View Shallow Tests To `GithubTabView` and `GitTabView` (100% coverage)

### Documentation

TBD

### Release Notes

TBD

I welcome all to critique.


smashwilson:
> IMO, active pane repository detection could be removed, and we could always just select the first project path, which could simplify the code, because the active pane detection branch now only runs for small edge cases, where benefit is negligible.

👍 Yeah, I'd much rather get rid of this complexity entirely. Explicit choice all the way.

> However, I have also considered allowing a toggle for people to prioritize active pane or have them choose. I would love to here thoughts about this.

I think it'd be better to fully commit to the explicit choice from the context tile. It's a pretty significant simplification of behavior and less surprising all around. And it'd be nice not to be stuck maintaining both.

smashwilson:
@wadethestealth Is this PR intended to replace #2299, or are they independent?

wadethestealth:
@smashwilson they are independent, but I unfortunately have no where to put remote selection again. I can update it with new ui, but I will need a place to put it. I know @simurai suggested status bar, but it was not swung one way.

smashwilson:
Hmm. Maybe we can unify them into the same dropdown, one entry per project root - remote pair? Like this:

prototype

smashwilson:
Oh, also, you can steal markup and CSS from the prototype branch:

https://github.com/atom/github/compare/prototypes/fr1934-context

wadethestealth:
@smashwilson, I don't hate it but that is a lot of tiles already. It's definitely functional. And yes I have been doing so (first commit on this branch is authored by simurai from that branch)

wadethestealth:
@smashwilson I have run into some issues with testing. Inside `github-package.test.js`, it seems that the way it was setup there is state leak between all of the async test. By that I mean they are sharing they same global and when executed fast enough the latter ran function will override the state the other need leading to a fail. I was originally trying to fix any test warnings I saw, but now this pr has a 400 diff. If this is ok, I will stop hacking on this pr and move the rest of fixing the tests to another PR. If this is too large for your team I can rewrite a smaller pr not fixing all the test warnings. In any case, if you personally test the cases with only they will pass, just not as is with the state leak. I would love to hear what you think I should do as I am currently trying to figure out a way to get this done.

smashwilson:
Yeah, unfortunately `github-package.test.js` is one of the oldest and gnarliest test suites in the codebase at the moment (the RootController tests are another, as are some of the StagingView-related ones). I'm not terribly surprised you're having problems converting it over.

> If this is ok, I will stop hacking on this pr and move the rest of fixing the tests to another PR.

👍 That's perfect, and I see you've started it in #2315. ✨ 

> If this is too large for your team I can rewrite a smaller pr not fixing all the test warnings.

So to be clear "my team" is currently "me, a few days a month, on my own time." 🙃 That might not be the case forever but I want to make sure our expectations are communicated correctly!

> I would love to hear what you think I should do as I am currently trying to figure out a way to get this done.

I think you're on the right track:

1. Rework the GitHub package tests in a separate PR;
2. Get that merged into master;
3. Merge that into this branch, then change the tests in this PR to exercise the new behavior.

wadethestealth:
> So to be clear "my team" is currently "me, a few days a month, on my own time." 🙃 That might not be the case forever but I want to make sure our expectations are communicated correctly!

Is this because a lack of atom staff?
If so you could always house an intern 😉 to help out. I know where to find one 🙃.

wadethestealth:
Ready for review @smashwilson

lib/views/github-tab-view.js
```diff
@@ -89,4 +93,14 @@ export default class GitHubTabView extends React.Component {
       </div>
     );
   }
+
+  renderHeader() {
+    return (
+      <HeaderView
+        handleProjectSelect={e => this.props.changeProjectWorkingDirectory(e.target.value)}
+        currentProject={this.props.workingDirectory}
+        projectPaths={this.props.project.getPaths()}
```
smashwilson: Project paths don't necessarily correspond to git repository working directories. You can have a project path (or multiple project paths!) that are subdirectories of a git working directory, and you can have project paths that have no git working directories.

Finding the candidate git working directories can be fairly expensive - it takes a lot of filesystem accesses to walk up the directory tree. We have a Workdir

We'll need to implement a function (likely way up in `GithubPackage`) that enumerates the available _working directories_ instead, based on the project paths, and then passes them in here to use in the drop-down instead.

lib/views/header-view.js
```diff
@@ -0,0 +1,32 @@
...
+        <select className="github-Project-path input-select"
+          value={this.props.currentProject ? this.props.currentProject : undefined}
+          onChange={this.props.handleProjectSelect}>
+          {this.renderProjects()}
+        </select>
+      </header>
+    );
+  }
+
+  renderProjects = () => {
```
smashwilson: This doesn't need to be a property-set method because it isn't called in a way that explicitly needs `this` bound:

```suggestion
  renderProjects() {
```

test/atom/decoration.test.js
```diff
@@ -128,15 +129,40 @@ describe('Decoration', function() {
...
-          <div className="decoration-subtree">
-            This is a subtree
-          </div>
-        </Decoration>
-      );
-      assert.throws(() => mount(app), 'You are trying to decorate a gutter but did not supply gutterName prop.');
+    describe('throws an error', function() {
+      let errors;
+
+      // This consumes the error rather than printing it to console.
```
smashwilson: This is pretty cool ✨

test/atom/decoration.test.js
```diff
@@ -128,15 +129,40 @@ describe('Decoration', function() {
...
+          <ErrorBoundary>
+            <Decoration editor={editor} decorable={marker} type="gutter">
+              <div className="decoration-subtree">
+                This is a subtree
+              </div>
+            </Decoration>
+          </ErrorBoundary>
+        );
+        mount(app);
+        assert(errors[0], 'You are trying to decorate a gutter but did not supply gutterName prop.');
```
smashwilson: This `assert()` call is only asserting that `errors[0]` is truthy, and using the expected message as a failure message. Here's what you likely mean:

```suggestion
        assert.strictEqual(errors[0], 'You are trying to decorate a gutter but did not supply gutterName prop.');
```

test/atom/marker.test.js
```diff
@@ -135,8 +136,31 @@ describe('Marker', function() {
...
+      });
+
+      afterEach(function() {
+        errors = [];
+        window.removeEventListener('error', onError);
+      });
+
+      it('if its ID is invalid', function() {
+        mount(<ErrorBoundary><Marker editor={editor} id={67} /></ErrorBoundary>);
+        assert(errors[0], 'Error: Invalid marker ID: 67');
```
smashwilson:
```suggestion
        assert.strictEqual(errors[0], 'Error: Invalid marker ID: 67');
```

Similars

test/containers/comment-decorations-container.test.js
```diff
@@ -184,8 +184,10 @@ describe('CommentDecorationsContainer', function() {
 
     it('renders nothing if query errors', function() {
       const tokenWrapper = localRepoWrapper.find(ObserveModel).renderProp('children')('1234');
+      const stub = sinon.stub(console, 'warn');
       const resultWrapper = tokenWrapper.find(QueryRenderer).renderProp('render')({
         error: 'oh noes', props: null, retry: () => {}});
+      stub.restore();
```
smashwilson: This `.restore()` call isn't strictly necessary - we set up `sinon` as a sandbox that automatically gets restored after each test. 🌠

test/containers/current-pull-request-container.test.js
```diff
@@ -45,6 +45,7 @@ describe('CurrentPullRequestContainer', function() {
         pushInProgress={false}
 
         onOpenIssueish={() => {}}
+        onOpenReviews={() => {}}
```
smashwilson: ✨

test/github-package.test.js
```diff
@@ -647,69 +597,6 @@ describe('GithubPackage', function() {
       });
     });
 
-    describe('with an active pane in a non-repository project', function() {
```
smashwilson: So nice to see all of this red in this test suite 😂

test/views/git-tab-view.test.js
```diff
@@ -282,4 +282,13 @@ describe('GitTabView', function() {
...
     assert.isTrue(setFocus.calledWith(GitTabView.focus.RECENT_COMMIT));
   });
+
+  it('calls changeProjectWorkingDirectory when a project is selected', async function() {
+    const select = sinon.spy();
+    const path = 'test/path';
+    const wrapper = mount(await buildApp({changeProjectWorkingDirectory: select}));
+    wrapper.find('.github-Project-path.input-select').simulate('change', {target: {value: path}});
+    assert.isTrue(select.calledWith(path));
+    wrapper.unmount();
```
smashwilson: This is a bit different than the way the rest of the suites are structured. This is more of an integration test: it uses `mount` and simulates DOM events on selectors that are controlled by child components. Generally, we use `shallow` as much as we can, and we focus tests in each suite on the markup and immediate children created by the component under test in each file.

Ideally, this test would be in a new suite dedicated to the `HeaderView` component. We _could_ have tests up the component hierarchy to ensure the callbacks are passed properly, but it likely wouldn't be worth it 😄

test/views/header-view.test.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import {shallow} from 'enzyme';
+import path from 'path';
+
+import HeaderView from '../../lib/views/header-view';
+
+describe('HeaderView', function() {
```
smashwilson: 👍 Excellent.

lib/atom/marker.js
```diff
@@ -25,7 +25,7 @@ export const DecorableContext = React.createContext();
 class BareMarker extends React.Component {
   static propTypes = {
     ...markerProps,
-    id: PropTypes.string,
+    id: PropTypes.number,
```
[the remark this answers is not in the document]
wadethestealth: Yes

lib/containers/current-pull-request-container.js
```diff
@@ -37,8 +37,6 @@ export default class CurrentPullRequestContainer extends React.Component {
     aheadCount: PropTypes.number,
     pushInProgress: PropTypes.bool.isRequired,
 
-    workspace: PropTypes.object.isRequired,
-
```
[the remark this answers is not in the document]
wadethestealth: The linter caught this as unused

lib/controllers/editor-comment-decorations-controller.js
```diff
@@ -20,7 +20,10 @@ export default class EditorCommentDecorationsController extends React.Component
     workspace: PropTypes.object.isRequired,
     editor: PropTypes.object.isRequired,
     threadsForPath: PropTypes.arrayOf(PropTypes.shape({
-      rootCommentID: PropTypes.string.isRequired,
+      rootCommentID: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.number,
+      ]).isRequired,
```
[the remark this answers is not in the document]
wadethestealth: It is used as a number in test and I went to check on the GitHub graphql documentation and for `ID`.

It says:
>Represents a unique identifier that is Base64 obfuscated. It is often used to refetch an object or as key for a cache. The ID type appears in a JSON response as a String; however, it is not intended to be human-readable. When expected as an input type, any string (such as "VXNlci0xMA==") or integer (such as 4) input value will be accepted as an ID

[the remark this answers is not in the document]
wadethestealth: Maybe I made the wrong change. I should probably have just made the tests use strings.
Lines causing the error:
https://github.com/atom/github/blob/0b58cd46448997d6190823383a1de8e9eb93ff7e/test/controllers/comment-decorations-controller.test.js#L36-L47

test/views/git-tab-view.test.js
wadethestealth: The only reason for this test was branch coverage. So what should I do remove it or move it to header-view.test.js?

lib/github-package.js
```diff
@@ -514,55 +528,40 @@ export default class GithubPackage {
...
+    // It is only possible to reach here if there there was no preferred directory, there are no project paths and the
+    // the active context's repository is undetermined.
     return this.activeContext;
   }
 
-  setActiveContext(nextActiveContext) {
-    if (nextActiveContext !== this.activeContext) {
+  setActiveContext(nextActiveContext, extras) {
+    // Always update rendering when paths change.
+    if (nextActiveContext !== this.activeContext || (extras && extras.projectPaths)) {
```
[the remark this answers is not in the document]
wadethestealth: If I don't rerender when the project paths change the header won't be updated to see new projects. I guess I could probably make the header subscribe to project path changes.

lib/github-package.js
```diff
@@ -601,6 +600,7 @@ export default class GithubPackage {
   }
 }
 
+// eslint-disable-next-line no-unused-vars
```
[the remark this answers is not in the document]
wadethestealth: We can get rid of it. I just thought it could be useful in the future so I left it.

lib/github-package.js
```diff
@@ -514,55 +528,40 @@ export default class GithubPackage {
...
+      // If the context exists chose it, else continue.
+      if (stateContext.isPresent()) {
+        return stateContext;
+      }
     }
 
-    if (this.project.getPaths().length === 1) {
-      // Single project
-      const projectPath = this.project.getPaths()[0];
+    const projectPaths = this.project.getPaths();
```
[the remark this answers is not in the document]
wadethestealth: extras is now removed

lib/views/git-tab-view.js
```diff
@@ -85,62 +87,95 @@ export default class GitTabView extends React.Component {
...
+    } else if (this.props.repository.hasDirectory() &&
+      !isValidWorkdir(this.props.repository.getWorkingDirectoryPath())) {
+      type = 'unsupported-dir';
+      isEmpty = true;
+    } else if (this.props.repository.showGitTabInit()) {
+      type = 'no-repo';
+      isEmpty = true;
+    } else if (this.props.isLoading || this.props.repository.showGitTabLoading()) {
+      isLoading = true;
+    }
```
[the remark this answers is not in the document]
wadethestealth: neat

lib/views/header-view.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import path from 'path';
+
+export default class HeaderView extends React.Component {
```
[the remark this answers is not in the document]
wadethestealth: I was think about the rest of the prototype, and I feel like with an avatar and a refresh button it isn't really a context tile view but more of a header for the git and github tabs

lib/views/github-tab-view.js
wadethestealth: Noted. I will look into it.

lib/controllers/editor-comment-decorations-controller.js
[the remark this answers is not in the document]
smashwilson: Yeah, I think it'd be better to change the tests? You're already taking on a lot (and I've given you a ton of comments 😬) so don't worry about the warning in this PR if you don't want to!

lib/error-boundary.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default class ErrorBoundary extends React.Component {
+  static propTypes = {
+    children: PropTypes.any,
```
smashwilson:
```suggestion
    children: PropTypes.node.isRequired,
```

`PropTypes.node` specifies "anything React can render," just to tighten it up a bit.

lib/error-boundary.js
```diff
@@ -0,0 +1,35 @@
...
+    fallback: PropTypes.any,
```
smashwilson:
```suggestion
    fallback: PropTypes.node,
```

lib/views/github-tab-view.js
smashwilson: (Also, for the record, I'm pretty sure I confused this myself in the feature request writeup. Only realized this while I was doing the review 😬)

lib/views/header-view.js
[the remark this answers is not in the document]
smashwilson: Good point. Hmm, how about something like `TabHeaderView`... ?

test/views/git-tab-view.test.js
smashwilson: Oh, I see, you need to cover this line?

https://github.com/atom/github/blob/5bb76cb7f003ec5ec3ee930e76caf253c5e19406/lib/views/git-tab-view.js#L121

Here's how I would do that in this suite:

```js
it('calls changeWorkingDirectory when a project is selected', async function() {
  const changeWorkingDirectory = sinon.spy();
  const wrapper = shallow(await buildApp({changeWorkingDirectory}));
  wrapper.find('HeaderView').prop('handleProjectSelect')({target: {value: 'some-path'}});
  assert.strictEqual(changeWorkingDirectory.calledWith('some-path'));
});
```

The logic here is that if we keep the `git-tab-view.test.js` tests written in terms of things defined directly in `GitTabView`, we can make changes further down the tree, like changing the markup inside `HeaderView`, without needing to change this suite, too. (And yeah, I know there are a bunch of existing tests that violate this 😬 It's one of those systemic changes I alluded to before. For now we just try to keep new tests written in this style.)

lib/views/header-view.js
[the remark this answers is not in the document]
wadethestealth: Ok

lib/github-package.js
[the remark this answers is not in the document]
smashwilson: Ohhhh. Yeah, you're absolutely right, I didn't think of that.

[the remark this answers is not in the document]
wadethestealth: I have solved this issue by subscribing to the workdir pool and using state to update just the component rather than the whole tree.

lib/views/github-tab-view.js
wadethestealth: Believed to be solved

test/views/git-tab-view.test.js
wadethestealth: 👍

test/containers/comment-decorations-container.test.js
wadethestealth: good to know. I am going to properly stub or fix these in a new PR as right now they consume too much.

wadethestealth:
Ready for review @smashwilson

lib/controllers/editor-comment-decorations-controller.js
```diff
@@ -20,10 +20,7 @@ export default class EditorCommentDecorationsController extends React.Component
     workspace: PropTypes.object.isRequired,
     editor: PropTypes.object.isRequired,
     threadsForPath: PropTypes.arrayOf(PropTypes.shape({
-      rootCommentID: PropTypes.oneOfType([
-        PropTypes.string,
-        PropTypes.number,
-      ]).isRequired,
+      rootCommentID: PropTypes.string.isRequired,
```
smashwilson: 👍 🙇

lib/github-package.js
```diff
@@ -550,7 +551,7 @@ export default class GithubPackage {
     }
 
     // It is only possible to reach here if there there was no preferred directory, there are no project paths and the
-    // the active context's repository is undetermined.
+    // the active context's repository is not undetermined.
```
smashwilson: Bit of a double negative huh 🤔

test/controllers/comment-decorations-controller.test.js
```diff
@@ -34,16 +34,16 @@ describe('CommentDecorationsController', function() {
     const branch = new Branch('featureBranch', upstreamBranch, upstreamBranch, true);
     const {commentThreads} = aggregatedReviewsBuilder()
       .addReviewThread(t => {
-        t.addComment(c => c.id(0).path('file0.txt').position(2).bodyHTML('one'));
+        t.addComment(c => c.id('0').path('file0.txt').position(2).bodyHTML('one'));
```
smashwilson: 🙇

aviatesk:
@wadethestealth @smashwilson 
So after this PR, github package doesn't automatically switch its active repository according to the current active editor, right ?
(In other word, now this package is kinda _sticky_ to the selected repository)

I found the previous behaviour is more useful, especially when editing files outside of the current projects.
Any discussion on this ?

wadethestealth:
@aviatesk There had been talk of this feature a while ago, and during their research I believe they decided it was an unnatural experience and was confusing to users. I merely implemented, what was already decided/designed a while back. There is a bit of discussion on one of my resolved comments, where I thought about leaving in the active pane switching and having a setting, but it was decided, that it was better to simplify code by removing it. It inherently slows down cross project workflows, but allows for clarification of what you are in. Also note that the active pane was not *always* the active context, and it was also a guess as to what a user wanted, whereas this is explicit and user determined.

aviatesk:
@wadethestealth Thanks for the reply.
Still I personally object to the idea, but could you tell me if there is any command to switch the active repository ? With the current behaviour I want a way to switch it without any GUI but with a shortcut, at least.

wadethestealth:
@aviatesk there is none right now, but that would definitely be a good add. I wonder what this would look like though. Would it just give focus to the expanded select and you use arrow keys like normal? I think that's probably the best option.

aviatesk:
That could definitely be one option, and I guess creating a separate selector view (like what can be provided by https://github.com/atom/atom-select-list) could be an alternative.

smashwilson:
@aviatesk: Yeah, you can see the writeup of the motivation and vision for context management on the feature request PR: https://github.com/atom/github/pull/1934 (note the "rendered" link in the PR description).

I'm a definite :+1: on having at least one mouse-free way to change the active git workdir context. Another relatively quick-to-implement idea might be to add a command-palette command that jumps to the context of the current active pane item?

Can you open a new issue to discuss this? Make it a little easier to discover than comments on a merged PR :smile:

aviatesk:
@wadethestealth Thanks for opening the issue :) Just for a reference, you can find the issue here: https://github.com/atom/github/issues/2335

mordash:
Hi,

is there an option to disable this list ?
It really is not cool when working on several projects at the same time.

Best regards.

wadethestealth:
@mordash currently no, but there is a pr not merged that adds a shortcut to switch to the git project of the current file you are in, it's this shortcut is supposed to both facilitate old behavior and make switching easier. We know it is cumbersome for multiproject workflows, but the change was necessary to remove confusion on how project selection worked.

mattlyons0:
I do not believe any of the options listed in #2335 are sufficient to make the new behavior as good as the exiting behavior was.

If you are switching projects quickly finding the currently selected project from a dropdown can be difficult, especially with many projects open. Adding a shortcut to open the currently selected project is a nice step in the right direction but it is still an extra step from before.

I have added a setting to restore the old behavior of selecting the project based on the active panel in #2393 . I like this option because it gets around the confusing aspect since it is an opt-in experience and disables the dropdown but keeps it visible showing the current project name.

That said it is possible in addition to this solution adding a command to select the current project still has value for users who do not want to go all the way of enabling this new setting.

wadethestealth:
@mattlyons0 I'm sorry you feel this way. The main reason for the change was because the old way was extremely user unfriendly, and instead of adding an option to maintain the old version and the new version, we opted out of keeping the old version in to reduce technical debt and code complexity as the old decision algorithm used guessing what the user wanted rather than explicit choice. This is not to deny your or, but rather to explain the context around the situation and some of the relevant to your case avenues of the many avenues discussed. Thank you for the feedback on this change as well.

mattlyons0:
@wadethestealth The old behavior can be conditionally added with 19 lines of code in a single file as I have done so in this commit: https://github.com/atom/github/pull/2393/commits/08aef187deb986edbf98bd4074323de10d902929

I do not see that as a large amount of technical complexity to maintain a feature that is notably useful for users working in multiple repositories. This has only been released for 4 days and already 7 users have found this several month old pull request indicating they would prefer the old behavior.

I don't think the old behavior is for everyone but it certainly has value for power users, of which I think Atom definitely has a lot of (given a major difference between Atom and VSCode is its flexibility). Putting this behind a configuration setting at the cost of merely 2 functions of additional technical complexity added seems like a valuable addition to me.

wadethestealth:
@mattlyons0 it has only officially be released this long, but it has be in the nightly for a while. Also there will always be back lash to change in any degree. There are also a lot more atom users than 7 and of course people frustrated with the change are a lot more likely to seek out this pr and comment. That being said you provide fair points. I am not a maintainer of this repo and this change is a very old todo from a group of the previous maintainers. Ultimately the decision comes down to @smashwilson or an official atom team member.

jarvelov:
Another user in favor of the old autoselect behavior here. I thought this was a bug introduced in 1.43. I don't see why we can't have both the git repo select and the old autoselect functionality. If this functionality can be implemented in the 19 lines of code @mattlyons0 I don't see how that code is complex and the feature is very valuable to me.

I, not being hyperbolic, always work with multiple repos and have to switch between them a lot and now to have to manually change the git repo adds another step for me that I previously didn't have to do.

I'm fine with it not being the default behavior but please add back the old functionality as a configuration setting.

smashwilson:
👋 Hey everyone. Can we centralize discussions about this on #2335 instead of here? Comments on merged PRs are hard to discover for non-involved users and don't have a good way to signal when we reach some kind of consensus.

I'll 🔒 this so this comment with the issue link is the last one and everyone can see it :wink:

PR #2315, opened 2019-10-19
Fix GitHub Package Tests

wadethestealth:
### Description of the Change
- Removes state leak from previous tests.
- Refactors layout/structure of test to reflect BDD best pratices (For good examples, see [Better Specs](http://www.betterspecs.org/#describe)) *Note: we prefer describe in place of context.*
- **All** test functions remain testing the **exact** same properties.

### Benefits

- Remove state leak.
- Fix flakes
- Cleaner test assertion and test description

### Possible Drawbacks

This will slow down test, because we were previously not preventing state leak.

### Applicable Issues

#2308 Ran into trouble implementing good test cases, because state leak was corrupting the results.

### Metrics

- All tests passing without retries.

### Tests

I fixed 'em 😉 

### Release Notes

Updated test case accuracy.


.eslintrc
```diff
@@ -36,8 +36,10 @@
     "describe": true,
     "fdescribe": true,
     "xdescribe": true,
+    "context": true,
```
smashwilson: I'm :-1: on changing our testing language (at least in this PR).

* It's strictly a wording change, not a functional one, so it doesn't support this PR's goal of improving the stability of the GitHub package suite.
* In this context, I feel more strongly about _consistency_ than I do about the "correctness" of either option.
  * We have a giant test suite and it would take a long time to percolate this change through it incrementally. So, introducing this change the more natural way would result in an extended period of inconsistency.
  * Doing it all at once in a batch change has similar drawbacks to linter changes: it introduces spurious conflicts with open PRs, it interferes with blame data, and it's a hard diff to validate visually since it's so repetitive.
* There are a lot of other systematic things I'd like to fix throughout our test suite before we addressed this. For example, we currently have about three distinct ways to create fixture data.
* I have some [`init.js` Atom hacks](https://gist.github.com/smashwilson/79b018bc7c9c1732a698b0d2a188986f) built around intelligently folding `describe` / `it` suites that I've shared around.
* Honestly, I'm more accustomed to `describe` and `it` at this point. I think they read reasonably well.

test/github-package.test.js
```diff
@@ -8,617 +8,836 @@ import {fileExists, getTempDir} from '../lib/helpers';
...
-  });
-
-  afterEach(async function() {
-    await githubPackage.deactivate();
-
-    atomEnv.destroy();
-  });
-
-  async function contextUpdateAfter(chunk) {
+  async function contextUpdateAfter(githubPackage, chunk) {
```
smashwilson: 👍 Yeah, much nicer.

test/github-package.test.js
```diff
@@ -8,617 +8,836 @@ import {fileExists, getTempDir} from '../lib/helpers';
...
 
     async function constructWith(projectPaths, initialPaths) {
       const realProjectPaths = await Promise.all(
         projectPaths.map(projectPath => getTempDir({prefix: projectPath})),
       );
 
+      const {
+        workspace, project, commands, notificationManager, tooltips,
+        deserializers, config, keymaps, styles, grammars,
+      } = atomEnv;
```
smashwilson: Haha yesss. (The original code here far predated JS support for object destructuring. Even in babel, I think.)

test/github-package.test.js
```diff
@@ -8,617 +8,836 @@ import {fileExists, getTempDir} from '../lib/helpers';
...
+
+      const confirm = atomEnv.confirm.bind(atomEnv);
+      const currentWindow = atomEnv.getCurrentWindow();
+      const configDirPath = path.join(__dirname, 'fixtures', 'atomenv-config');
```
smashwilson: Oh, fun fact: `path.join()` implicitly normalizes path separators. So you can write this a tad more compactly:

```suggestion
      const configDirPath = path.join(__dirname, 'fixtures/atomenv-config');
```

And it works on Windows:

```
> const path = require('path')
undefined
> path.join('a/b', 'c/d/e')
'a\\b\\c\\d\\e'
>
```

Not a big deal here, but can sometimes save you a lot of `"", "", ""` stuff.

test/github-package.test.js
```diff
@@ -8,617 +8,836 @@ import {fileExists, getTempDir} from '../lib/helpers';
...
 
-      const context = contextPool.getContext(workdirPath2);
-      assert.isTrue(context.isPresent());
-      assert.strictEqual(context.getRepository(), githubPackage.getActiveRepository());
-      assert.strictEqual(context.getResolutionProgress(), githubPackage.getActiveResolutionProgress());
-      assert.equal(githubPackage.getActiveWorkdir(), workdirPath2);
+      it('uses the project\'s context', function() {
+        const context = contextPool.getContext(workdirPath);
+        assert.isTrue(context.isPresent());
+        assert.equal(githubPackage.getActiveWorkdir(), workdirPath);
```
smashwilson: I see this is just lifted from the previous code right now, but I tend to assert with `.strictEqual` instead of `.equal` unless I have a good reason not to... `.strictEqual` compares its arguments with `===` while `.equal` compares them with `==`, so it has similar properties of avoiding surprising type coercions.

test/github-package.test.js
```diff
@@ -8,617 +8,836 @@ import {fileExists, getTempDir} from '../lib/helpers';
...
-      project.setPaths([workdirPath1, workdirPath2]);
-      await workspace.open(path.join(workdirPath2, 'b.txt'));
+    context('with only projects', function() {
+      let workdirPath1, workdirPath2, nonRepositoryPath;
+      beforeEach(async function() {
+        ([workdirPath1, workdirPath2, nonRepositoryPath] = await Promise.all([
+          cloneRepository('three-files'),
+          cloneRepository('three-files'),
+          getTempDir(),
+        ]));
```
smashwilson: Oh neat, I don't think I've seen this trick to destructure into existing locals like this. Cool!

test/github-package.test.js
```diff
@@ -8,617 +8,836 @@ import {fileExists, getTempDir} from '../lib/helpers';
...
+      });
     });
 
-    // Don't worry about this on Windows as it's not a common op
-    if (process.platform !== 'win32') {
-      it('handles symlinked project paths', async function() {
+    context('with a symlinked repository project', function() {
+      beforeEach(async function() {
+        if (process.platform === 'win32') {
+          this.skip();
```
smashwilson: 👍 Yeah, I like using `.skip()` more than I like conditionally defining test cases.

test/github-package.test.js
```diff
@@ -8,617 +8,836 @@ import {fileExists, getTempDir} from '../lib/helpers';
...
+      getLoadSettings = atomEnv.getLoadSettings.bind(atomEnv);
+      currentWindow = atomEnv.getCurrentWindow();
+      configDirPath = path.join(__dirname, 'fixtures', 'atomenv-config');
+      renderFn = sinon.stub().callsFake((component, element, callback) => {
+        if (callback) {
+          process.nextTick(callback);
+        }
+      });
+
+      useLegacyPanels = !workspace.getLeftDock;
```
smashwilson: ☝️ This check is for pretty ancient versions of Atom. We could probably get rid of it at this point.

test/github-package.test.js
```diff
@@ -683,64 +910,183 @@ describe('GithubPackage', function() {
...
+      styles = atomEnv.styles;
+      grammars = atomEnv.grammars;
+      getLoadSettings = atomEnv.getLoadSettings.bind(atomEnv);
+      currentWindow = atomEnv.getCurrentWindow();
+      configDirPath = path.join(__dirname, 'fixtures', 'atomenv-config');
+      renderFn = sinon.stub().callsFake((component, element, callback) => {
+        if (callback) {
+          process.nextTick(callback);
+        }
+      });
```
smashwilson: 🤔 I'd love to find a way to avoid duplicating all of this. A lot of it might go away if we can move to hooks and use an Atom environment context, though, so I'm unsure if it would be worth the effort to refactor out now.

.eslintrc
wadethestealth: reverted

test/github-package.test.js
wadethestealth: will look for this and replace

test/github-package.test.js
wadethestealth: will remove

test/github-package.test.js
wadethestealth: I simplified it a little more, I think it looks nicer now.

wadethestealth:
> How are we doing on test coverage so far? 

It doesn't seem super awful. 
Things I have noticed:
- Any outside api you are consuming (eg. status bar) is not being covered.
- A couple of branches are only being tested with  one conditional outcome.
- Serialization is not tested.
- Yardstick setup is basically untested.
- A couple of nested function/properties that are functions are not being covered.

Issue #2310, opened 2019-10-17
Commit Diff: Jump to file does nothing


wadethestealth:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Description

When selecting a past commit, a diff view pops up of the commit, but when you right click there is a jump to file context, which when clicked does nothing rather than taking the user to the file shown in the commit diff.

### Steps to Reproduce

1. Open past commit diff
2. Right click on a line
3. Select 'Jump To File'
4. Nothing happens

**Expected behavior:**

You jump to the file.

**Actual behavior:**

Nothing changes.

**Reproduces how often:**

100%

### Versions

Atom    : 1.42.0-nightly8
Electron: 4.2.7
Chrome  : 69.0.3497.128
Node    : 10.11.0
OS: Windows 10

rsese:
Thanks!  Confirmed on macOS 10.14.6 with 1.42.0-nightly8 :+1:

Issue #2098, opened 2019-04-25
"Jump to File" from context menu in Files tab doesn't do anything


vanessayuenn:
### Prerequisites

* Have you tried to reproduce the problem in [Safe Mode][safe-mode]?
* Have you followed all applicable steps in the [debugging guide][debugging-guide]?
* Have you checked the FAQs on the [message board][message-board] for common solutions?
* Have you checked that your issue [isn't already filed][already-filed]?

[safe-mode]: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
[debugging-guide]: http://flight-manual.atom.io/hacking-atom/sections/debugging/
[message-board]: https://discuss.atom.io/c/faq
[already-filed]: https://github.com/atom/github/issues

### Description

Right-clicking in the multifilepatch view in Files tab within PR view shows the context menu with the item "Jump to File", but clicking on it doesn't do anything.

**Expected behavior:**

Ideally, "Jump to File" should take the user to the correct file on disk, provided that the PR has been checked out. 

**Actual behavior:**

Nothing happens.

**Reproduces how often:**

100%

PR #2324, opened 2019-10-25
Removed jump to file from options when viewing past commits

darangi:
### Description of the Change
Removed jump to file from the list of options when viewing a past commit or a diff.

### Possible Drawbacks

None.
### Applicable Issues

Fixes https://github.com/atom/github/issues/2310

### DIFF
--- a/menus/git.cson
+++ b/menus/git.cson
@@ -46,16 +46,6 @@
       'command': 'github:jump-to-file'
     }
   ]
-  '.github-FilePatchView': [
-    {
-      'type': 'separator'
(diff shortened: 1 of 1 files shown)


darangi:
> 👍 Yah, good call.
> 
> I'd love to get this to actually _work_ at some point. But there are a ton of edge cases we'd need to account for.

At some point in the future, we will look into this 👍

wadethestealth:
@smashwilson @darangi @lkashef This should also close #2098. Can one of you close it?

PR #2365, opened 2019-12-10
Re-sync electron-link and electron-mksnapshot with atom/atom

smashwilson:
Looks like the versions of electron-link and electron-mksnapshot that we use to verify snapshottability in our CI builds have drifted from the ones in Atom:

https://github.com/atom/atom/blob/f0e4ca6d5d55f2422c107fcea93edba10ce816ac/script/package.json#L14-L15

Let's re-sync and fix any issues that have arisen in the meantime.


smashwilson:
Turns out any Relay version newer than 5.0.0 segfaults mksnapshot. I have no idea why 🤔

PR #2459, opened 2020-05-06
use action-setup-atom

UziTech:
### Description of the Change

use [`UziTech/action-setup-atom@v1`](https://github.com/marketplace/actions/setup-atom) for installing Atom in GitHub Actions

### Screenshot/Gif

N/A

### Alternate Designs

N/A

### Benefits

Cleaner workflow files and works on any os and channel easily

### Possible Drawbacks

none

### Applicable Issues

https://github.com/atom/github/pull/2298#issuecomment-624331385

### Metrics

N/A

### Tests

Ran tests on every platform

### Documentation

N/A

### Release Notes

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,12 @@ jobs:
     - name: install dependencies
       run: apm ci
 
+    - name: configure git
+      shell: bash
+      run: |
+        git config --global user.name Hubot
+        git config --global user.email hubot@github.com
+
     - name: Run the tests
       if: ${{ !contains(matrix.os, 'windows') }}
       run: atom --test test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       if: contains(matrix.os, 'windows')
       run: |
         choco install visualcpp-build-tools --version=14.0.25420.1 --ignore-dependencies -y --params "'/IncludeRequired'"
-        echo ::set-env name=VCTargetsPath::'C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\v140'
+        echo "VCTargetsPath='C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\v140'" >> $GITHUB_ENV
     - name: install dependencies
       run: apm ci
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,52 +1,46 @@
 name: ci
 on:
+  pull_request:
   push:
     branches:
     - master
-  pull_request:
-    types:
-    - opened
-    - reopened
-    - synchronize
 jobs:
-  linux-tests:
-    name: linux tests
-    runs-on: ubuntu-18.04
+  tests:
+    name: tests
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, macos-10.14, windows-latest]
+        channel: [beta, nightly]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
     env:
-      DISPLAY: ":99"
       ATOM_GITHUB_BABEL_ENV: coverage
-      ATOM_CHANNEL: nightly
-      ATOM: /tmp/atom/usr/bin/atom-nightly
-      APM: /tmp/atom/usr/share/atom-nightly/resources/app/apm/bin/apm
       MOCHA_TIMEOUT: 60000
       UNTIL_TIMEOUT: 30000
     steps:
     - uses: actions/checkout@v1
     - name: install Atom
-      shell: bash
+      uses: UziTech/action-setup-atom@v1
+      with:
+        channel: ${{ matrix.channel }}
+
+    - name: install windows build tools
+      if: contains(matrix.os, 'windows')
       run: |
-        curl -s -L https://atom.io/download/deb?channel=${ATOM_CHANNEL} \
-          -H 'Accept: application/octet-stream' \
-          -o /tmp/atom-amd64.deb
-        /sbin/start-stop-daemon --start --quiet \
-          --pidfile /tmp/custom_xvfb_99.pid \
-          --make-pidfile --background \
-          --exec /usr/bin/Xvfb \
-          -- :99 -ac -screen 0 1280x1024x16
-        sudo apt-get update
-        sudo apt-get install -yyq libgconf-2-4 build-essential git libsecret-1-dev gvfs-bin
-        dpkg-deb -x /tmp/atom-amd64.deb /tmp/atom
+        choco install visualcpp-build-tools --version=14.0.25420.1 --ignore-dependencies -y --params "'/IncludeRequired'"
+        echo ::set-env name=VCTargetsPath::'C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\v140'
     - name: install dependencies
-      shell: bash
-      run: ${APM} ci
-    - name: configure git
-      shell: bash
-      run: |
-        git config --global user.name Hubot
-        git config --global user.email hubot@github.com
-    - name: run tests
-      shell: bash
-      run: ${ATOM} --test test/
+      run: apm ci
+
+    - name: Run the tests
+      if: ${{ !contains(matrix.os, 'windows') }}
+      run: atom --test test
+
+    - name: Run the tests on Windows
+      if: ${{ contains(matrix.os, 'windows') }}
+      continue-on-error: true # due to https://github.com/atom/github/pull/2459#issuecomment-624725972
+      run: atom --test test
+
     - name: report code coverage
       shell: bash
       env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -9,144 +9,75 @@ on:
...
-
-  macos-tests:
-    name: macos tests
-    runs-on: macos-10.14
+  tests:
+    name: tests
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, macos-10.14, windows-2016]
+        channel: [stable, beta, nightly]
```
UziTech, line 17: I wasn't sure if you only wanted to run against `nightly` or any other channels

UziTech:
@smashwilson I'm sure this will need some changes but it is a good start for using the GitHub Action for installing Atom.

When I was testing it out it looks like the windows tests were just failing because paths were getting converted to [8.3 file paths](https://en.wikipedia.org/wiki/8.3_filename)

```
  48) WorkdirCache
       clears the cache when the maximum size is exceeded:

      assert.strictEqual(actualDir, expectedDir)
      + expected - actual

      -C:\Users\runneradmin\AppData\Local\Temp\git-fixture-202046-3340-4cwu10.vgyo3
      +C:\Users\RUNNER~1\AppData\Local\Temp\git-fixture-202046-3340-4cwu10.vgyo3
      
      at Context.<anonymous> (D:\a\github\github/workdir-cache.test.js:143:12)
```

If you want me to close this PR and you want to take this and work on it in your own PR I would be fine with that.

.github/workflows/ci.yml
Duv4lC4ruso: k

aminya (approved):

.github/workflows/ci.yml
```diff
@@ -1,152 +1,84 @@
...
+    strategy:
+      matrix:
+        # os: [ubuntu-18.04, macos-10.14, windows-2016]
+        os: [ubuntu-18.04, macos-10.14]
+        # channel: [stable, beta, nightly]
+        channel: [nightly]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
     env:
-      DISPLAY: ":99"
```
aminya, line 16: Isn't this env variable needed anymore? This is useful for Linux runs.

.github/workflows/ci.yml
```diff
@@ -1,152 +1,84 @@
...
-          -o /tmp/atom-amd64.deb
-        /sbin/start-stop-daemon --start --quiet \
-          --pidfile /tmp/custom_xvfb_99.pid \
-          --make-pidfile --background \
-          --exec /usr/bin/Xvfb \
-          -- :99 -ac -screen 0 1280x1024x16
-        sudo apt-get update
-        sudo apt-get install -yyq libgconf-2-4 build-essential git libsecret-1-dev gvfs-bin
-        dpkg-deb -x /tmp/atom-amd64.deb /tmp/atom
+        choco install vcbuildtools
```
aminya, line 30: Visual Studio is already installed on the GitHub images: https://github.com/actions/virtual-environments

.github/workflows/ci.yml
UziTech: That env variable is set in the action to allow atom to run on Linux. We don't need to set it in two places.

.github/workflows/ci.yml
UziTech: Windows tests weren't working without that install so maybe at the time this PR was created they weren't automatically installed? I'll wait to remove it until this PR actually gets some attention. By that time many other parts might need to be changed.

.github/workflows/ci.yml
aminya: I see. Yes, that works!

.github/workflows/ci.yml
aminya: Windows-latest has the 2019 build tools. Maybe It is because you have to install 2015 or 2017 version. Choco installs 2015 with this command:
https://chocolatey.org/packages/vcbuildtools

UziTech: I agree, if it works with that it would be better

aminya:
```suggestion
        choco install visualstudio2017buildtools -y
```

UziTech: This isn't really worth working on unless someone at @github or @atom is actually considering merging it.

aminya:
@smashwilson Could you take a look at this? Having better CI helps to debug the issues. setup-atom allows installing stable, beta, and nightly.

For example this:
https://github.com/atom/atom/pull/21782

.github/workflows/ci.yml
aminya: This is the fastest way I found to install visual studio 2015:
```suggestion
      - name: Install Visual Studio 2015 on Windows
        if: ${{ contains(matrix.os, 'windows') }}
        run: |
          choco install visualcpp-build-tools --version=14.0.25420.1 --ignore-dependencies -y --params "'/IncludeRequired'"
          echo ::set-env name=VCTargetsPath::'C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\v140'
```

It is used in language-javascript: 
https://github.com/atom/language-javascript/blob/e5a7d95edfca40dcae78fd536fff04393f5fe3c0/.github/workflows/ci.yml#L34

.github/workflows/ci.yml
aminya: @UziTech Please resolve this.

.github/workflows/ci.yml
```diff
@@ -1,152 +1,84 @@
...
 jobs:
-  linux-tests:
-    name: linux tests
-    runs-on: ubuntu-18.04
+  tests:
+    name: tests
+    strategy:
+      matrix:
+        # os: [ubuntu-18.04, macos-10.14, windows-2016]
+        os: [ubuntu-18.04, macos-10.14]
```
aminya, line 13:
```suggestion
        os: [ubuntu-18.04, macos-10.14, windows-2016]
```

UziTech: The tests on windows are failing because of https://github.com/atom/github/pull/2459#issuecomment-624725972

aminya: Still, we should run them to observe if a new test starts to fail. We should continue to on error to not block the CI. This is much better than not running them.

```yaml
      - name: Run the tests
        if: ${{ !contains(matrix.os, 'windows') }}
        run: atom --test spec

      - name: Run the tests on Windows 
        if: ${{ !contains(matrix.os, 'windows') }}
        continue-on-error: true # due to https://github.com/atom/github/pull/2459#issuecomment-624725972
        run: atom --test spec
```

UziTech: I think a better solution would be to figure out how to fix the 8.3 path names in the tests and run them normally on windows. But as I said above, unless @smashwilson actually wants to merge this PR, it is just a POC.

UziTech: It should be pretty easy to create a function that can compare regular paths with FAT 8.3 paths in tests.

aminya:
I made a PR making this PR ready to go: https://github.com/UziTech/github/pull/9

.github/workflows/ci.yml
aminya: Definitely, in the long run, we can fix the test.

smashwilson: > unless @smashwilson actually wants to merge this PR, it is just a POC.

I'm :+1: on merging this if one of you has time to get it green!

And I'm 👍🏻 on including the Windows builds until we can fix that crash.

aminya:
The failing tests are genuine failures. See here:
https://github.com/atom/atom/pull/21782#issuecomment-753330546

UziTech:
@aminya I updated this PR with the code from your PR on my repo.

Looks like the tests are failing because of a timeout error. I'm not sure if that is something to do with this PR or not.

aminya:
I don't think it is because of this PR (as mentioned in [here](https://github.com/atom/atom/pull/21782#issuecomment-753330546))

One of the failing tests:
https://github.com/atom/github/blob/5a8faf5bdfa682bbfeb2f1c8af823bbcaa12d254/test/integration/file-patch.test.js#L241

aminya:
@UziTech Could you update this branch?

.github/workflows/ci.yml
```diff
@@ -1,172 +1,91 @@
...
-      shell: bash
+    - name: install windows build tools
+      if: contains(matrix.os, 'windows')
       run: |
-        curl -s -L https://atom.io/download/mac?channel=${ATOM_CHANNEL} \
-          -H 'Accept: application/octet-stream' \
-          -o /tmp/atom.zip
-        sudo unzip -q /tmp/atom.zip -d /tmp/atom
+        choco install visualcpp-build-tools --version=14.0.25420.1 --ignore-dependencies -y --params "'/IncludeRequired'"
+        echo ::set-env name=VCTargetsPath::'C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\v140'
```
aminya, line 31: The old command is deprecated now. 

```suggestion
        echo "VCTargetsPath='C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\v140'" >> $GITHUB_ENV
```

DeeDeeG:
About the CI failures here:

@smashwilson has indicated multiple times that these tests cause frequent headaches, out of proportion to their use for catching actual problems.

https://github.com/atom/github/pull/2617#issuecomment-769097766

They are notorious for flaking out and he has strongly considered deleting them.

You can also observe that these tests have been patched in multiple ways to make their timeouts more lenient. They rely on actual filesystem access, which is very unreliable, particularly in CI where you are testing in some VM somewhere on a shared datacenter host machine.

Take a look at these two separate workarounds that already exist for these flaky tests:

https://github.com/atom/github/blob/v0.36.6/test/integration/file-patch.test.js#L18-L23

That said, I believe there is a solution.

The following line is clearly meant to _increase_ timeout length to 9 seconds: https://github.com/atom/github/blob/v0.36.6/test/integration/file-patch.test.js#L22

However, upon closer inspection, I believe this line _decreases_ the timeout length. There is already a longer global timeout (30 seconds) set in the CI's env vars: 

https://github.com/atom/github/blob/v0.36.6/.github/workflows/ci.yml#L22

Here is my patch to fix this unintentionally restrictive and short timeout:

https://github.com/DeeDeeG/github/commit/e0bc5dbd198e6da4d9dbf1e3a61e70393c95a775

I believe with my patch (see link just above), this PR would pass CI.

DeeDeeG:
I'm considering the possibility that this is failing simply because of how many jobs there are. 6 parallel matrix jobs, starting in parallel with a lint and snapshot test job (8 jobs in the very beginning).

(Maybe the CI environment becomes super disk-bound -- on a tangent: note also that Atom's CI as of late has become massively parallel, and is at times very disk-heavy in each job).

I'd like to try it with max parallel jobs of 2 and work up from there if it passes.

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymax-parallel

aminya:
> I'm considering the possibility that this is failing simply because of how many jobs there are.

That is unrelated. The failures are genuine failures that should be fixed. 

These matrix jobs run on **different machines**. Nothing runs in parallel in one machine.

DeeDeeG:
This PR is passing if rebased right on top of `v0.36.3` tag. So the regression in the `integration: file-patching` in certain circumstances is real.

DeeDeeG:
I am personally not pleased to see that the main file `dist/index.js` for `action-setup-atom` is 5200 lines long... And rolls a lot of its own code for things that have built-in versions in Node.JS (like a custom filesystem copy function)... This is comparatively hard to read and debug, IMO.

I'm struggling to see what's different between this and the `master` branch's approach to downloading Atom. But I can't very quickly understand these thousands of lines of JS.

https://github.com/UziTech/action-setup-atom/blob/v1.1.12/dist/index.js

I think a shell script could implement `action-setup-atom`'s functionality in under 500 lines, and rely on broadly tested utilities for everything (for example, `cp`). And for this specific repo, the existing bash lines are 4 lines (macOS) or 11 lines (Ubuntu). Very readable and debuggable.

Interesting revelations about the `integration: file-patching` tests aside, why is this better for this repo? Why not just `curl` and `unzip`?

UziTech:
This PR only changes the way Atom and apm are downloaded and installed for the tests. It doesn't change the tests or code in any way. It basically takes the download scripts and puts them in an action. The only real difference is which folder Atom and apm are installed in.

DeeDeeG:
One technically meaningful change that stood out to me was, this makes a symlink `atom` pointing to `atom.sh` and puts those on the `PATH`. And then runs that as `atom [args]`. Whereas before, it was the actual executable binary (`Atom*.app/Contents/MacOS/Atom*`) being called directly on the macOS and Ubuntu CI runs.

So I wonder if changing the CI on `master` to run via `atom.sh` would cause the same timeouts? I intend to try that when I get a free moment.

DeeDeeG:
For the e-mail chain: That last comment was wrong. Sorry about that.

smashwilson:
Ohhh you know what? I wonder if those integration tests now rely on having `user.name` and `user.email` populated in the system's git config, or the git identity pane is shown instead of the staging view. I bet the existing build configuration sets them, but this build does not.

.... I'm going to be upset if that's the _entire build problem._

DeeDeeG:
Upset + bugfix identified is an emotional rollercoaster, which I have been on many times as well, but it ends up fixing the project, so it's the way it's gotta be sometimes.

I definitely think the code is _capable of_ working correctly but isn't for some very superficial cause.

smashwilson:
Oh yeah, this is a feeling I'm well acquainted with 😂 

The better solution would be to rework the integration tests somehow to not rely on global git configuration settings (if we do end up keeping them).  But let's see if this makes the build green, first.

DeeDeeG:
Not my fix to steal the thunder on, so thank you @UziTech and @smashwilson for sticking with this, but:

Linux/macOS are green!!!

aminya:
> Ohhh you know what? I wonder if those integration tests now rely on having `user.name` and `user.email` populated in the system's git config, or the git identity pane is shown instead of the staging view. I bet the existing build configuration sets them, but this build does not.
> 
> .... I'm going to be upset if that's the _entire build problem._

I knew that the issue is an assumption about the environment that does not hold true 😁.

UziTech:
Nice catch @smashwilson . Now to fix windows 🤔

DeeDeeG:
I just meant to emphasize those as Windows has been green all along. There must be a git config loaded into the runner or something on Windows.

aminya:
> I just meant to emphasize those as Windows has been green all along. There must be a git config loaded into the runner or something on Windows.

That's because of the `continues-on-error: true`. Fixing Windows seems too much work now. It should be done in other PRs.

DeeDeeG:
Oh, right. Well that's a whole other conundrum then. But way beyond scope of this PR I think.

aminya:
When I use continue-on-error, I write the number of test failures inside the CI file, to check if the failures are increased. If the changes add new test failures that means there is an issue with that change.

The stats on previous Windows run:
image

This one that is running now should probably make 7 tests passing.

DeeDeeG:
I am seeing that there are some "expected tests" for the PR that are never going to run. The runs for the latest commit are all green.

aminya (approved):

smashwilson:
Excellent. Thanks everyone 🎉

Issue #2509, opened 2020-09-10
Sync electron-* dependencies with Atom.


DeeDeeG:
## Summary

Please sync `electron-mksnapshot` with Atom.

## Motivation

- The old `mksnapshot` binaries were [very large on Linux](https://github.com/electron/electron/releases/tag/v4.2.0), for some reason. (See `mksnapshot-v4.2.0-linux-[arch].zip`, these are ~350-480 MB on Linux, ~30-35 MB on macOS and Windows.) This makes building Atom slower and use more network data and disk space on Linux than it has to.
- Judging by this repo's commit history, staying in sync with Atom has been a goal in the past. Testing on the same version as Atom should help stay on top of any Electron-related bugs when using this in Atom.

## Describe alternatives you've considered

N/A, just bumping depednencies and syncing with Atom.

## Additional context

Atom is on `electron-mksnapshot` `^0.4.3`, and `electron-mksnapshot` `^9.0.2`*

* `electron-mksnapshot` 9.0.2 allows you to install a `mksnapshot` binary from an exact Electron version, down to the patch level. This requires installing `electron-mksnapshot` with the `ELECTRON_CUSTOM_VERSION` env var set to the version of electron you're targeting. Atom is currently targeting Electron 6.1.12, and configuring `electron-mksnapshot` to download the exact correct `mksnapshot` binary using this `ELECTRON_CUSTOM_VERSION` env var.

smashwilson:
👍🏻 

I don't suppose you'd have time to put a pull request together? I could get you a quick merge if you did 😇

DeeDeeG:
Closed by #2513. Thanks for merging.

PR #2513, opened 2020-09-10
Sync electron-* devDependencies with atom/atom

DeeDeeG:

### Requirements

### Description of the Change

Match `electron-`* dependencies to match Atom.

- Bump `electron-link` from 0.4.1 to 0.4.3
- Bump `electron-mksnapshot` to 9.0.2
  - This newer version of the `electron-mksnapshot` module allows specifying an exact Electron version you want to target, down to the patch level. I added a `script/redownload-electron-bins.js` file to set this env var when appropriate, and re-download (the correct version of) the Electron-vendored `mksnapshot` binary.

### Screenshot/Gif

N/A

### Alternate Designs

- We could just use `electron-mksnapshot@6.0.0` and hope that it's similar enough to `6.1.12` to catch all the bugs we need to catch.
  - Pro: Less complexity in the repo. Simple.
  - Cons: Might not catch the same bugs? Hard to know. Also, the `mksnapshot` binaries on Linux are still [huge](https://github.com/electron/electron/releases/tag/v6.0.0) for that release (~420-450 MB).

- When this `github` module is being installed as a dependency in the main Atom repo, the new `script/redownload-electron-bins.js` reads the main Atom repo's `{ package.json }.electronVersion`and downloads the `mksnapshot` binary for that version of Electron.
  - Pro: Atom CI should now fail earlier on the GitHub package tests if an Electron bump breaks the GitHub module.
  - Con: This automatic behavior might be unexpected and confusing.

Let me know if you prefer the version of the Electron-vendored `mksnapshot` binary simply be hard-coded at this repo, and not dynamically set based on the parent Atom repo.

### Benefits

Sync with Atom, catch all the bugs! :tada:

Much smaller (~1/10th as large) `mksnapshot` binary should save Linux users network bandwidth, hard-drive space, memory consumption...

### Possible Drawbacks

`script/redownload-electron-bins.js` adds some complexity to the repo.

### Applicable Issues

#2509

### Metrics

N/A?

### Tests

I am currently relying on the existing CI for validating this.

### Documentation

N/A

### Release Notes

N/A

N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3061,18 +3061,22 @@
       "integrity": "sha512-//oorEZjL6sbPcKUaCdIGlIUeH26mgzimjBB77G6XRgnDl/L5wOnpyBGRe/Mmf5CVW3PwEBE1NjiMZ/ssFh4wA=="
     },
     "abstract-leveldown": {
-      "version": "5.0.0",
-      "resolved": "https://registry.npmjs.org/abstract-leveldown/-/abstract-leveldown-5.0.0.tgz",
-      "integrity": "sha512-5mU5P1gXtsMIXg65/rsYGsi93+MlogXZ9FA8JnwKurHQg64bfXwGYVdVdijNTVNOlAsuIiOwHdvFFD5JqCJQ7A==",
+      "version": "6.3.0",
+      "resolved": "https://registry.npmjs.org/abstract-leveldown/-/abstract-leveldown-6.3.0.tgz",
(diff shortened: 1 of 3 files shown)


DeeDeeG:
cc @smashwilson here's my "updating the electron-* devDependencies" PR.

It adds a script to take advantage of the `ELECTRON_CUSTOM_VERSION` env var. Without that, we'd have to use `electron-mksnapshot@6.0.0`. For some reason the Electron 6 update at atom/atom didn't work with `electron-chromedriver` and `electron-mksnapshot` at `v6.0.0`, it had to be more precisely `6.1.12`. so I thought I'd do the same thing at this repo, and use exact vendored `mksnapshot` binary down to the patch level.

smashwilson:
✨ Excellent, this is really cool. Thank you!

PR #2533, opened 2020-09-28
Flip CI back to nightly

smashwilson:
Let's see if https://github.com/atom/atom/issues/21341 is fixed yet or not 🤞🏻


smashwilson:
It is. Huzzah!

Issue #1357, opened 2018-03-19
[Feature request] Syntax highlighting in diff views


robertrossmann:
The diff pane could really benefit from syntax highlighting. Right now it's just green/red with white text.

I have not seen this in any of the roadmaps I have seen so far so I thought I'd post this as a feature request. Not even sure how difficult it would be to implement, but given the seemingly complex structure of the diff view this might be a non-trivial task. 🤔

Thank you for considering this!

Arcanemagus:
Somewhat related to https://github.com/atom/github/issues/759, since that would make this slightly more complex.

wesinator:
👍 VS Code already has this in its Changes view

Issue #759, opened 2017-05-08
Enhancement Request: Word Highlighting for Diffs


jasonrudolph:
### Description

When viewing the diff for a small change to an existing line, it would be awesome for the package to use word highlighting to help the developer quickly spot the change. For example, compare the two diffs below. In the current diff display in the Atom github package, you have to look closely to identify which part of the line has changed. In the diff display on github.com, the [word highlighting](https://github.com/blog/1885-better-word-highlighting-in-diffs) helps you more quickly see the change.

##### Diff displayed by Atom github package

![screenshot](https://cloud.githubusercontent.com/assets/2988/25807190/ef676a4e-33d3-11e7-93d1-f3be57a36de8.png)

##### Diff displayed on github.com

![github](https://cloud.githubusercontent.com/assets/2988/25807189/ef66b630-33d3-11e7-85e2-524b787cf519.png)

### Steps to Reproduce

1. Open a git repository in Atom and change a part of a line for an existing file. For example:
    1. `git clone https://github.com/atom/github.git && cd github && atom .`
    1. Open the `LICENSE` file
    1. On the [first line](https://github.com/atom/github/blob/0516cb1c3ee0e13cf863374e5abce4d93982ecfa/LICENSE#L1) change "2016" to "2017"
1. Run "GitHub: Toggle Git Tab"
1. Click on the `LICENSE` file in the "Unstaged Changes" tab

**Expected behavior:**

It would be awesome to have the diff highlight the single-character change, much like it does on github.com or via the git command line:

![screen shot 2017-05-08 at 10 06 36 am](https://cloud.githubusercontent.com/assets/2988/25807890/577c477e-33d6-11e7-9af3-a947d4baf425.png)

**Actual behavior:**

The diff does not highlight the single-character change:

![screen shot 2017-05-08 at 10 05 19 am](https://cloud.githubusercontent.com/assets/2988/25807891/577c72a8-33d6-11e7-9c3c-0c2ced2d02b7.png)

**Reproduces how often:**

Always

### Versions

- Atom github package v0.0.2
- Atom 1.18.0-dev-df8dd8b45
- apm  1.18.1
- macOS 10.12.4

AugPro:
Hello, I also think this would be a great enhancement. I see that the issue is still open but not present in any roadmap. Do you guys still plan on making this enhancement? 🙂

thesystemninjaneer:
[...]

PR #2538, opened 2020-10-10
Add support for syntax highlighting

aminya:

### Description of the Change

This adds support for syntax highlighting

### Screenshot/Gif

I selected "source.js" manually here to test and diff now looks so beautiful
image

### Benefits
Colorful editors! 

### Applicable Issues

#1357 
#759

### DIFF
--- a/lib/atom/atom-text-editor.js
+++ b/lib/atom/atom-text-editor.js
@@ -81,6 +81,20 @@ export default class AtomTextEditor extends React.Component {
   componentDidMount() {
     const modelProps = extractProps(this.props, editorCreationProps);
 
+    // syntax highlighting
+    let scopeName = 'text.plain';
+    const buffer = modelProps.buffer;
+    if (buffer) {
+      const bufferPath = buffer.getPath();
+      if (bufferPath) {
+        // find scope name based on the path and text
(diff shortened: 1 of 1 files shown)


smashwilson:
Oh, sweet! 😍 

> Remaining: `buffer.getPath()` usually returns undefined. That means we should provide the file path in the buffer object, so we can choose a better syntax highlighting.

One complication is that we have several views that display the content of multiple files - possibly with different grammars - within the same buffer. See CommitDetailItem,  CommitPreviewItem, or the PullRequestChangedFiles components within the pull request item. I think we'll need a way to apply a grammar to a _range_ within a buffer.

aminya:
> Oh, sweet! 😍
> 

😃 

> > Remaining: `buffer.getPath()` usually returns undefined. That means we should provide the file path in the buffer object, so we can choose a better syntax highlighting.
> 
> One complication is that we have several views that display the content of multiple files - possibly with different grammars - within the same buffer. See CommitDetailItem, CommitPreviewItem, or the PullRequestChangedFiles components within the pull request item. I think we'll need a way to apply a grammar to a _range_ within a buffer.

We can implement the alternate design instead. If I am right, PatchBuffer is responsible for making the multi-file diffs.

Issue #2558, opened 2020-11-11
Git identity prompt always appearing


cvande04:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Description

Hello, I am a long-time Atom user but a first-time poster. I ran into a novel bug earlier today regarding the Git package, specifically the Git identity prompt. Please let me know if the description of the problem is unclear in any way:

The Git Identity prompt appears every time I open the Git panel in Atom. It asks me to enter my Git username and email. However, I have run both `git config --global user.name` and `git config --global user.email` (including prior to the bug manifesting) and double-checking the config file shows that the values are set properly:

```
[user]
	name = (my username)
	email = (my email)
[github]
	user = (my username)
```

For a file in a repository that is initialized on Github, the Git identity panel often (but not always) pre-populates the fields with my username and email as found in the config file. For a file in a repository that has not been initialized, the name and email address fields are blank. Switching between two repositories, regardless of whether they are initialized or not, often (but not always) results in the Git identity prompt reappearing, even if I have just used it to sign in moments before switching, and even if I can initially see the Git panel for a few moments before the identity prompt appears. Sometimes, the identity panel will flicker rapidly back-and-forth between having the fields pre-populated and having them blank over the course of a few seconds. After this, trying to re-enter my username or email in the fields often results in glitches, where the character I typed appears in the field for a fraction of a second before vanishing as if I had hit the backspace key. I have tried everything in the debugging guide, including uninstalling/reinstalling Atom and disabling/re-enabling the Git package, as well as re-entering the `git config --global` values for name and email and repeatedly restarting my computer. 

[...]

PR #2564, opened 2020-11-16
Read and write git configuration without repository

smashwilson:
Reading and writing global or system git configuration entries, like `user.email` and `user.name`, makes sense even if the current repository is absent or uninitialized. Let's let these calls succeed by using a git strategy initialized with a workdir of the drive root.

This addresses #2558 by behaving correctly in a variety of situations where we _should_ be able to read your identity from the global git configuration, but weren't. We prevent git commands from executing in contexts where we don't have a valid working directory to run them with - if a project root is not an initialized git repository, or if we have no project root yet, or if a git repository is still loading - but `git config` doesn't have this constraint; unless you specify `--local`, it's perfectly reasonable to read and write global or system git config values without a working directory. Allowing this to happen closes the windows of time during which the git tab can attempt to read `user.name` and `user.email`, get empty strings back from the stubbed `getConfig` methods on those repository states, and conclude that it needs to display the identity panel to prompt you for them.

I also needed to implement cross-instance cache invalidation among repositories in the `WorkdirContextPool`, so that setting your identity in one context would properly cause any other repositories in other contexts to pick up the changes that you wrote (and not, for example, immediately delete them).

Note that non-present repository states do not cache these config values, so they could be redundantly read frequently. I think that's okay for now because config reads are very quick. I also haven't implemented a filesystem watcher on the global configuration file to pick up external identity changes automatically.


dnorthup-ums:
@smashwilson I'm not seeing the v0.36.6 update available in Atom yet, and I just upgraded it to 1.54.0 to make sure that wasn't the problem. Did you remember to push the release up to the Atom Package Manager server?

smashwilson:
I didn't forget... I hit a mysterious crash trying to do the actual upgrade in https://github.com/atom/atom/pull/21782 that I haven't been able to get to the bottom of yet 😞 . I'm guessing that it'll take me two or three days of dedicated time on the Windows box, which I have not been able to muster.

The good news is that there's a partial fix in Atom 1.54.0 that should improve things somewhat in the meantime.

nhassan201920:
#2643 to

paydenn (commented):
Hello world @ #

Issue #2624, opened 2021-02-10
Update uses of Electron API calls that require Promises in Electron 7+ (callbacks are deprecated)


DeeDeeG:
## Summary

There are some Electron APIs that changed to only return Promises (and not run callbacks passed to them), mostly as of Electron 7+. Updating to be compatible with those would be great for updating Atom's Electron version.

See: https://www.electronjs.org/docs/api/modernization/promisification

## Motivation

There is a Work In Progress pull request to update Atom to Electron 9, and I think updating these API calls may be necessary for the `github` package to continue to work with Electron 9. (https://github.com/atom/atom/pull/21777)

## Describe alternatives you've considered

N/A

## Additional context

Most (I think all) of these API calls generally _support_ Promises starting with Electron 6. (Some of them support Promises as of even earlier, in Electron 5.) So implementing these changes should still be compatible with all the current and in-development versions of Atom.

`github`'s tests are failing over at Atom core's Electron 9 PR: https://github.com/atom/atom/pull/21777#issuecomment-773708566

> Here are the error messages for the `github` package tests:

<details><summary>Error messages for the github package tests (click to expand):</summary>

```
  2441 passing (15m)
  1 pending
  4 failing

  1) GithubDotcomMarkdown
       opening issueish links
         opens item in pane and activates accordingly:
     Error: Trying to stub property 'open' of null
      at throwOnFalsyObject (/Users/runner/work/1/s/node_modules/github/node_modules/sinon/lib/sinon/throw-on-falsy-object.js:7:15)
      at Function.stub (/Users/runner/work/1/s/node_modules/github/node_modules/sinon/lib/sinon/stub.js:70:24)
      at Sandbox.stub (/Users/runner/work/1/s/node_modules/github/node_modules/sinon/lib/sinon/sandbox.js:331:37)
      at Context.<anonymous> (/Users/runner/work/1/s/node_modules/github/github-dotcom-markdown.test.js:143:13)
      at processImmediate (internal/timers.js:439:21)

  2) GithubDotcomMarkdown
       opening issueish links
         records event for opening issueish in pane item:
     Error: Trying to stub property 'open' of null
      at throwOnFalsyObject (/Users/runner/work/1/s/node_modules/github/node_modules/sinon/lib/sinon/throw-on-falsy-object.js:7:15)
      at Function.stub (/Users/runner/work/1/s/node_modules/github/node_modules/sinon/lib/sinon/stub.js:70:24)
[...]

PR #2625, opened 2021-02-11
Update shell.openExternal to promise due to electron update on atom

sadick254:
### Requirements

### Description of the Change
There are some Electron APIs that changed to only return Promises (and not run callbacks passed to them), mostly as of Electron 7+. Updating to be compatible with those would be great for updating Atom's Electron version.

### Applicable Issues
https://github.com/atom/github/issues/2624

### DIFF
--- a/lib/controllers/issueish-list-controller.js
+++ b/lib/controllers/issueish-list-controller.js
@@ -90,15 +90,9 @@ export class BareIssueishListController extends React.Component {
     return null;
   }
 
-  openOnGitHub = url => {
-    return new Promise((resolve, reject) => {
-      shell.openExternal(url, {}, err => {
-        if (err) { reject(err); } else {
-          resolve();
-          addEvent('open-issueish-in-browser', {package: 'github', component: this.constructor.name});
-        }
-      });
-    });
+  openOnGitHub = async url => {
+    await shell.openExternal(url);
+    addEvent('open-issueish-in-browser', {package: 'github', component: this.constructor.name});
   }
 
   showActionsMenu = /* istanbul ignore next */ issueish => {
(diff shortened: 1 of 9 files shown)


lib/controllers/issueish-list-controller.js
```diff
@@ -90,15 +90,14 @@ export class BareIssueishListController extends React.Component {
...
-      });
-    });
+  openOnGitHub = async url => {
+    try {
+      const resolvedPromise = await shell.openExternal(url);
+      addEvent('open-issueish-in-browser', {package: 'github', component: this.constructor.name});
+      return resolvedPromise;
+    } catch (err) {
+      throw new Error(err);
+    }
```
smashwilson, line 100: We can simplify these!

The try/catch statements aren't actually doing anything, since they're just re-throwing any error that's raised. We also don't need to return `resolvedPromise` since the return value isn't significant. So these functions can look like this:

```suggestion
    await shell.openExternal(url);
    addEvent('open-issueish-in-browser', {package: 'github', component: this.constructor.name});
```

smashwilson (commented):
Hmm CodeCov is complaining about these... they must not be covered by existing tests. I think that's probably not worth addressing here, though, because these methods don't do anything but call two other functions that we'd need to stub anyway.

smashwilson:
Looks like we'll need to update the stubbing in some tests, too.

lib/controllers/issueish-list-controller.js
sadick254: Hey @smashwilson. Good point. Done 👍

test/controllers/issueish-list-controller.test.js
```diff
@@ -90,15 +90,15 @@ describe('IssueishListController', function() {
 
     it('calls shell.openExternal with specified url', async function() {
       const wrapper = shallow(buildApp());
-      sinon.stub(shell, 'openExternal').callsArg(2);
+      sinon.stub(shell, 'openExternal').callsFake(() => { });
```
smashwilson, line 93: I think these could also be `.returns(Promise.resolve())`. No big deal though.

smashwilson (approved):
✨

PR #2626, opened 2021-02-13
Update to promise api for some methods in the electron API

asturur:

### Requirements

### Description of the Change

This PR aims at using the promise version of

- showSaveDialog,
- showOpenDialog,
- executeJavascript

electron API

In order to ease the upgrade path to newer versions of electron where the callback is no more supported.

### DIFF
--- a/lib/views/directory-select.js
+++ b/lib/views/directory-select.js
@@ -45,15 +45,13 @@ export default class DirectorySelect extends React.Component {
     );
   }
 
-  chooseDirectory = () => new Promise(resolve => {
+  chooseDirectory = () =>
     this.props.showOpenDialog(this.props.currentWindow, {
       defaultPath: this.props.buffer.getText(),
       properties: ['openDirectory', 'createDirectory', 'promptToCreate'],
-    }, filePaths => {
-      if (filePaths !== undefined) {
+    }).then(({filePaths}) => {
+      if (filePaths.length) {
         this.props.buffer.setText(filePaths[0]);
       }
-      resolve();
     });
-  });
 }
--- a/lib/views/directory-select.js
+++ b/lib/views/directory-select.js
@@ -45,13 +45,13 @@ export default class DirectorySelect extends React.Component {
     );
   }
 
-  chooseDirectory = () =>
-    this.props.showOpenDialog(this.props.currentWindow, {
+  chooseDirectory = async () => {
+    const {filePaths} = await this.props.showOpenDialog(this.props.currentWindow, {
       defaultPath: this.props.buffer.getText(),
       properties: ['openDirectory', 'createDirectory', 'promptToCreate'],
-    }).then(({filePaths}) => {
-      if (filePaths.length) {
-        this.props.buffer.setText(filePaths[0]);
-      }
     });
+    if (filePaths.length) {
+      this.props.buffer.setText(filePaths[0]);
+    }
+  }
 }
--- a/lib/views/git-timings-view.js
+++ b/lib/views/git-timings-view.js
@@ -328,9 +328,9 @@ class WaterfallWidget extends React.Component {
     const buffer = new TextBuffer({text: json});
     dialog.showSaveDialog({
       defaultPath: 'git-timings.json',
-    }, filename => {
-      if (!filename) { return; }
-      buffer.saveAs(filename);
+    }).then(({filePath}) => {
+      if (!filePath) { return; }
+      buffer.saveAs(filePath);
     });
   }
 }
@@ -423,9 +423,9 @@ export default class GitTimingsView extends React.Component {
     e.preventDefault();
     dialog.showOpenDialog({
       properties: ['openFile'],
-    }, async filenames => {
-      if (!filenames) { return; }
-      const filename = filenames[0];
+    }).then(async ({filePaths}) => {
+      if (!filePaths.length) { return; }
+      const filename = filePaths[0];
       try {
         const contents = await fs.readFile(filename, {encoding: 'utf8'});
         const data = JSON.parse(contents);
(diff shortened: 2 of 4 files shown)


lib/views/directory-select.js
```diff
@@ -45,15 +45,13 @@ export default class DirectorySelect extends React.Component {
...
   }
 
-  chooseDirectory = () => new Promise(resolve => {
+  chooseDirectory = () =>
     this.props.showOpenDialog(this.props.currentWindow, {
       defaultPath: this.props.buffer.getText(),
       properties: ['openDirectory', 'createDirectory', 'promptToCreate'],
-    }, filePaths => {
-      if (filePaths !== undefined) {
+    }).then(({filePaths}) => {
```
smashwilson, line 52: I think I'd prefer to use the `async`/`await` style to call the Promise-based APIs here for consistency, if you don't mind 🙇🏻 .

Also, it looks like the return value of `showOpenDialog` has changed, too; it now returns an `Object` with certain properties:

https://github.com/electron/electron/blob/v9.4.3/docs/api/dialog.md#dialogshowopendialogbrowserwindow-options

So I think this call would be something like:

```js
chooseDirectory = async () => {
  const {canceled, filePaths} = await this.props.showOpenDialog(this.props.currentWindow, {
      defaultPath: this.props.buffer.getText(),
      properties: ['openDirectory', 'createDirectory', 'promptToCreate'],
  };
  if (!canceled && filePaths.length > 0) {
    this.props.buffer.setText(filePaths[0]);
  }
}
```

But, we have to make sure that the different return value is compatible with the call signature on the Electron versions back to Atom 1.55 (the current beta), too, so we don't break our ability to backport 🤔

test/views/directory-select.test.js
```diff
@@ -47,7 +47,7 @@ describe('DirectorySelect', function() {
 
   describe('clicking the directory button', function() {
     it('populates the destination path buffer on accept', async function() {
-      const showOpenDialog = sinon.stub().callsArgWith(2, ['/some/directory/path']);
+      const showOpenDialog = sinon.stub().returns(Promise.resolve({filePaths: ['/some/directory/path']}));
```
smashwilson, line 50: :+1: Perfect.

test/worker-manager.test.js
```diff
@@ -167,7 +167,7 @@ describe('WorkerManager', function() {
       });
       `;
 
-      await new Promise(resolve => browserWindow.webContents.executeJavaScript(script, resolve));
+      await browserWindow.webContents.executeJavaScript(script);
```
smashwilson, line 170: 🙇🏻

smashwilson (commented):
Thanks for tackling this! I'll be happy to merge if we can resolve:

* Using `async`/`await` instead of `.then` (which is just a style/consistency thing), and
* Confirming that this will still work against Atom 1.55, or finding a way to bridge the Electron API gap, since the return value of `showOpenDialog` has changed.

lib/views/directory-select.js
smashwilson: Hah! I totally missed at first read that you'd already added the `{filePaths}` destructuring in your `.then` callback argument.

asturur:
i ll get to requested changes soon

asturur:
@smashwilson if atom 1.55 is based on electron 6 those api change will work.
The callback was removed from documentation on electron 6 and still supported, but this promise API has been introduced in 6.0

smashwilson:
Looks like Atom 1.55 uses Electron 6.1.12:

https://github.com/atom/atom/blob/24274440ec3e9c296a20daa3ad72aa57c074024a/package.json#L15

And here are the `showOpenDialog` docs for that version:

https://github.com/electron/electron/blob/v6.1.12/docs/api/dialog.md#dialogshowopendialogbrowserwindow-options

So you're correct, looks like we're good 👍🏻

lib/views/directory-select.js
```diff
@@ -45,15 +45,13 @@ export default class DirectorySelect extends React.Component {
...
-      if (filePaths !== undefined) {
-        this.props.buffer.setText(filePaths[0]);
-      }
-      resolve();
     });
-  });
+    if (filePaths.length) {
+      this.props.buffer.setText(filePaths[0]);
+    }
+  }
```
smashwilson, line 56: :+1: Perfect!

lib/views/git-timings-view.js
```diff
@@ -328,9 +328,9 @@ class WaterfallWidget extends React.Component {
     const buffer = new TextBuffer({text: json});
     dialog.showSaveDialog({
       defaultPath: 'git-timings.json',
-    }, filename => {
-      if (!filename) { return; }
-      buffer.saveAs(filename);
+    }).then(({filePath}) => {
+      if (!filePath) { return; }
+      buffer.saveAs(filePath);
```
smashwilson, line 333: Can we use async/await here too?

lib/views/git-timings-view.js
```diff
@@ -423,9 +423,9 @@ export default class GitTimingsView extends React.Component {
     e.preventDefault();
     dialog.showOpenDialog({
       properties: ['openFile'],
-    }, async filenames => {
-      if (!filenames) { return; }
-      const filename = filenames[0];
+    }).then(async ({filePaths}) => {
```
smashwilson, line 426: And here ☝🏻 

Sorry, should have clarified that I meant at all three callsites on the first review 👀

asturur: oh ok no issue, i can convert them, i m still here.

smashwilson (approved):
Excellent, thank you. Will merge when builds are green 🍏

asturur:
Is it possible to get a new version published? what is your procedure around that?

smashwilson:
Ah, to make it easier to move forward the Electron upgrade work? Sure, I can do that:

```
atom-github ➤ apm publish patch
Preparing and tagging a new version ✓
Pushing v0.36.8 tag ✓
Publishing github@v0.36.8 ✓
```

asturur:
thanks @smashwilson appreciated

PR #2681, opened 2021-05-14
fix warning at #2680

icecream17:
**Please be sure to read the [contributor's guide to the GitHub package](https://github.com/atom/github/blob/master/CONTRIBUTING.md) before submitting any pull requests.**

### Description of the Change

https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops says:
_"If you need to __perform a side effect__ (for example, data fetching or an animation) in response to a change in props, use `componentDidUpdate` lifecycle instead."_

`componentDidUpdate` is called after updating, rather than `componentWillRecieveProps`. If this (or some other subtle change) happens to break atom, feel free to restart this pr and simply rename the method to `UNSAFE_componentWillReceiveProps` instead.

Warning text:

___
Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.

* Move data fetching code or side effects to componentDidUpdate.
* If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
* Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.
___

In this case, it's used for side effects (1st bullet point)

This implies there will still be a warning when using `UNSAFE_componentWillReceiveProps` in strict-mode, so trying `componentDidUpdate` first.

### Screenshot or Gif

N/A

### Applicable Issues

Fixes #2680

PR #2683, opened 2021-05-14
update os matrix in ci

icecream17:
which is currently `ubuntu-18.05`

See https://github.com/atom/github/pull/2681#issuecomment-841298220

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     name: tests
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-10.14, windows-latest]
+        os: [ubuntu-18.04, macos-latest, windows-latest]
         channel: [beta, nightly]
       fail-fast: false
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     name: tests
     strategy:
       matrix:
-        os: [ubuntu-18.04, macos-10.14, windows-latest]
+        os: [ubuntu-latest, macos-10.14, windows-latest]
         channel: [beta, nightly]
       fail-fast: false
     runs-on: ${{ matrix.os }}


icecream17:
oh wait, I accidentally did ubuntu instead of macOS

smashwilson:
> oh wait, I accidentally did ubuntu instead of macOS

Would it be worthwhile to do both?

smashwilson:
No idea what's going on with Actions here. It looks like all of the jobs are being instantly cancelled... ?

icecream17:
1. Maybe the explicit version was kept for backwards compatibility??? If not, then I guess so
2. https://github.community/t/github-actions-failed-due-to-internal-failure-how-to-rerun/18104 => https://www.githubstatus.com/, it seems like github actions is down or something

smashwilson:
Yeah, I just saw the Actions outage right after I posted that, heh. I'll re-run once it's resolved to see how this does 🤞🏻

smashwilson:
Alright, progress 👍🏻 

The failure message on Windows makes me wonder if the Windows images have been changed to already include the build tools:

```
ERROR: Running ["C:\Users\runneradmin\AppData\Local\Temp\chocolatey\visualcpp-build-tools\14.0.25420.1\visualcppbuildtools_full.exe" -NoRestart -Quiet -Log "C:\Users\runneradmin\AppData\Local\Temp\chocolatey\visualcpp-build-tools.log" ] was not successful. Exit code was '1603'. Exit code indicates the following: Generic MSI Error. This is a local environment error, not an issue with a package or the MSI itself - it could mean a pending reboot is necessary prior to install or something else (like the same version is already installed). Please see MSI log if available. If not, try again adding '--install-arguments="'/l*v c:\visualcpp-build-tools_msi_install.log'"'. Then search the MSI Log for "Return Value 3" and look above that for the error..
```

The Ubuntu errors are more odd:

```
  1) Git commands for CompositeGitStrategy made of [GitShellOutStrategy]
       exec
         when the WorkerManager is not ready or disabled
           kills the git process when cancel is triggered by the prompt server:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/github/github/test/git-strategies.test.js)
  

  2) Git commands for CompositeGitStrategy made of [GitShellOutStrategy]
       ssh authentication
         falls back to Atom credential prompts if SSH_ASKPASS is present but goes boom:
     Error: Timeout of 60000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/github/github/test/git-strategies.test.js)
```

These tests do rely on subprocess communication and have been flaky in the past, but never _this_ flaky. I would worry a bit more that these are finding an actual bug (with newer Ubuntu versions maybe) but it isn't clear to me what that could be.

smashwilson:
Let's merge this as-is to get _some_ kind of green in our CI 😅 .

